PR #68, opened 2025-04-16
[core] Introduce ContentClient abstraction

illiakovalenko:
- [ ] Changelog updated (Will be updated separately once we get all the information about naming)

## Description / Motivation
* Introduced a new `ContentClient` abstraction for interacting with the Content API.
* Added a static `createClient` method for client initialization, allowing developers to conveniently configure the client using environment variables.
* Implemented a `get` method to perform GraphQL requests and retrieve data from the API.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other - created a .ts script where I executed a single _manyBlog_ query

## Types of changes

### DIFF
--- a/packages/core/src/content/content-client.ts
+++ b/packages/core/src/content/content-client.ts
@@ -65,7 +65,7 @@ export class ContentClient {
       url: url || process.env.SITECORE_CS_URL,
       tenant: tenant || process.env.SITECORE_CS_TENANT || '',
       environment: environment || process.env.SITECORE_CS_ENVIRONMENT || 'main',
-      preview: preview || process?.env?.SITECORE_CS_PREVIEW === 'true',
+      preview: preview || process.env.SITECORE_CS_PREVIEW === 'true',
       token: token || process.env.SITECORE_CS_TOKEN || '',
     };
 
--- a/packages/core/src/content/content-client.ts
+++ b/packages/core/src/content/content-client.ts
@@ -1,0 +1,103 @@
+import { DocumentNode } from 'graphql';
+import { GraphQLRequestClient } from '../graphql-request-client';
+import { getContentUrl } from './utils';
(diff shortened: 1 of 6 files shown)


packages/core/src/content/content-client.ts
```diff
@@ -0,0 +1,103 @@
...
+  static createClient({
+    url,
+    tenant,
+    environment,
+    preview,
+    token,
+  }: Partial<ContentClientOptions> = {}): ContentClient {
+    const options = {
+      url: url || process.env.SITECORE_CS_URL,
+      tenant: tenant || process.env.SITECORE_CS_TENANT || '',
```
yavorsk, line 66: Most of the ContentClientOptions properties are required, but here we still add fallback logic in case they are missing?

yavorsk (approved):
See just one question above otherwise looks good!

packages/core/src/content/content-client.ts
yavorsk: oh, is the idea here to be validation logic?

illiakovalenko: It was marked as "Partial" in order to fallback to environment variables and fallback values, it's not required to be passed to factory

yavorsk: 👍 yep cool

packages/core/src/content/content-client.ts
```diff
@@ -0,0 +1,103 @@
...
+      environment: environment || process.env.SITECORE_CS_ENVIRONMENT || 'main',
+      preview: preview || process?.env?.SITECORE_CS_PREVIEW === 'true',
```
yavorsk:
```suggestion
      preview: preview || process.env.SITECORE_CS_PREVIEW === 'true',
```

PR #71, opened 2025-04-23
[cli] Add code extraction CLI command

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
v2-exclusive rework of https://github.com/Sitecore/jss/pull/2075

Notable changes:
- focus on componentMap instead of componentBuilder
- no longer reading env values for m2m authentication - config used instead
- cli command no longer reads tsconfig directly - compileOptions passed in cli config

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/core/src/tools/codegen/extract-components.ts
```diff
@@ -0,0 +1,75 @@
...
+import {
+  ExtractedFileType,
+  resolveComponentImportFiles,
+  sendCode,
+  validateConsent,
+  validateDeployContext,
+} from './utils';
+import { SitecoreConfig } from '../../config';
+
+export type ExtactComponentsArgs = {
```
yavorsk: typo here, `ExtractComponentsArgs`

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -9,6 +16,11 @@ export default defineCliConfig({
       generateSites({
         scConfig: config,
       }),
+      () =>
```
yavorsk: I think it would look better if extractComponents() returns the function to be executed by the cli, instead of being the actual function. Similarly to generateSites() and generateMetadata() above

packages/core/src/tools/codegen/extract-components.ts
yavorsk: actually another comment here - wouldn't it be better to follow the convention of the other ootb build commands that we have - generateMetadata, generateSites - and rename this type to `ExtractComponentsConfig`? 'args' sounds too cli-y

packages/core/src/tools/codegen/utils.ts
```diff
@@ -0,0 +1,243 @@
...
+  Json = 'json',
+  Package = 'package.json',
+}
+
+/**
+ * Validates consent for code extraction procedures
+ * @returns {boolean} - true if consent is given, false otherwise
+ */
+export const validateConsent = () => {
+  if (!process.env.EXTRACT_CONSENT) {
```
yavorsk: should we add `EXTRACT_CONSENT` to the template's .env? or not yet?

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -9,6 +16,11 @@ export default defineCliConfig({
...
+        extractComponents({
```
yavorsk: we control the execution of extractComponents only with `EXTRACT_CONSENT` variable, correct? do we need to provide corresponding var in the sitecore.config?

yavorsk (approved):
again, great work :) 
please have a look at few suggestions and questions.
approved in advance.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
art-alexeyenko: The consent logic is not finalized - current approach is a placeholder and we'll adjust this later.
Perhaps consent will indeed go into config, to be decided.

packages/core/src/tools/codegen/utils.ts
art-alexeyenko: The consent logic is not finalized, we'll decide this later.

packages/core/src/tools/codegen/extract-components.ts
art-alexeyenko: I can do it if you inisist. 
We kind of work with CLI here, since function is part of build CLI command. That's one point. Another - it would be better to not spam `...Config` types too much, IMO.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
art-alexeyenko: I actually wanted to ask you about this :)
What's the upside of using functions that return functions in our new CLI? On the surface it seems like unneccessary complication.

yavorsk: I had a variant as yours here initially, decided to changed to the current state after discussions with Illia. 
The goal is cleaner/simpler looking sitecore.cli.config

packages/core/src/tools/codegen/extract-components.ts
yavorsk: yea but extract components is not directly related to cli. it does not know/care that it is used by a cli command, correct? that's why i thought 'args' is not suitable here

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts
```diff
@@ -16,6 +16,12 @@ export default defineConfig({
       apiKey: process.env.NEXT_PUBLIC_SITECORE_API_KEY || '',
       apiHost: process.env.NEXT_PUBLIC_SITECORE_API_HOST || '',
     },
+    m2m: {
```
illiakovalenko: Do we expect that M2M will be used by the content-management CLI and potentially other features?
Currently, if I remove all the options under this property, I get an error stating that an ID and Secret must be provided.
However, at the moment, the code extraction feature is optional.

packages/core/src/tools/codegen/extract-components.ts
```diff
@@ -0,0 +1,77 @@
...
+  compilerOptions: {
+    [key: string]: unknown;
+  };
+  appFolder?: string;
+  componentMapPath?: string;
+};
+
+/**
+ * Extracts components from the app folder and sends them to XMCloud.
+ * @param {ExtractComponentsConfig} args - Arguments for extracting components
```
illiakovalenko: It would be good to align the naming with our other build-time commands and methods for consistency.
```suggestion
 * @param {ExtractComponentsConfig} config - Config for components extraction
```

packages/core/src/tools/codegen/extract-components.ts
```diff
@@ -0,0 +1,77 @@
...
+  validateDeployContext,
+} from './utils';
+import { SitecoreConfig } from '../../config';
+
+export type ExtractComponentsConfig = {
+  scConfig: SitecoreConfig;
+  compilerOptions: {
+    [key: string]: unknown;
+  };
+  appFolder?: string;
```
illiakovalenko: This parameter is not needed, right?
`process.cwd` can be mocked in unit tests instead

packages/core/src/tools/codegen/utils.ts
```diff
@@ -0,0 +1,245 @@
...
+  }
+  return false;
+};
+
+/**
+ * Parses the componentBuilder.ts file and returns a map of component names
+ * and their respective import strings
+ * @param {string} appPath path to the JSS app root
+ * @param {object} compilerOptions compiler options from tsconfig.json
+ * @param {string} componentMapPath path to the app's component map file. Default: 'src/lib/componentMap.ts'
```
illiakovalenko:
```suggestion
 * @param {string} [componentMapPath] path to the app's component map file. Default: 'src/lib/componentMap.ts'
```

packages/core/src/tools/auth/fetch-bearer-token.ts
```diff
@@ -0,0 +1,38 @@
+import chalk from 'chalk';
+import { DEFAULT_M2M_AUDIENCE, DEFAULT_M2M_ENDPOINT } from '../../constants';
+
+export const fetchBearerToken = async ({
```
illiakovalenko: let's align the changes from V1 to V2 and add missing jsdoc across the changes

packages/core/src/config/models.ts
```diff
@@ -202,6 +223,12 @@ export type SitecoreCliConfigInput = {
      */
     templates?: ScaffoldTemplate[];
   };
+  extractCode?: {
```
illiakovalenko: Should be removed

packages/core/src/tools/codegen/utils.ts
```diff
@@ -0,0 +1,245 @@
...
+ * @returns map of component names and their respective import strings
+ */
+export const resolveComponentImportFiles = (
```
illiakovalenko: To me, it looks like this functionality specific to files resolution, file extraction is quite framework-specific, particularly tied to React and Next.js.
Have you considered that?
Would it even be possible to reuse this functionality in other frameworks like Nuxt or Astro?
Probably not, especially since it requires providing a TypeScript configuration. Probably `core` is not the best place to keep such things

illiakovalenko (changes requested):

packages/core/src/tools/codegen/utils.ts
stasmaxymov: I want to add my point of view here. Main responsibility of `core` `next` and `react` packages is provide functionality to application and they are intended to run as part of application lifecycle. They should not be polluted with infra or tools related stuff. 
Additionally here we created an implicit `peerDependency` to typescript, it is not mentioned in the package.json, but it still is required to be resolved when this function is called, which complicates responsibilities of this package.

I suggest to move everything specific to dev/ci/cd context to `cli` package.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -1,6 +1,13 @@
...
 import config from './sitecore.config';
 import { defineCliConfig } from '@sitecore-content-sdk/nextjs/config';
-import { generateSites, generateMetadata } from '@sitecore-content-sdk/nextjs/tools';
+import {
+  generateSites,
+  generateMetadata,
+  extractComponents,
+} from '@sitecore-content-sdk/nextjs/tools';
+
+const tsConfig = ts.readConfigFile('./tsconfig.json', ts.sys.readFile);
```
stasmaxymov: I think we can resolve tsconfig within `extractComponents`, check project folder if there is one present and also allow optional parameter on function to account for not common locations or names. 
It also makes `cli.config` less typescript specific, enabling us to provide JS only solution,

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts
art-alexeyenko: yes, we do expect it to be the case

packages/core/src/tools/codegen/utils.ts
art-alexeyenko: Outside of code extraction, should nextjs/tools and react/tools be moved in future? @stasmaxymov

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts
stasmaxymov: I don't think  these belong to application config, it will be used only from CLI.
I also think env names aren't ideal, M2M is not name of auth technology, and might be somewhat misleading. 
Lets maybe change it to Sitecore_Auth_ClientId and so on

packages/core/src/tools/auth/fetch-bearer-token.ts
```diff
@@ -0,0 +1,38 @@
...
+        client_id: clientId,
+        client_secret: clientSecret,
+        audience: audience,
+        grant_type: 'client_credentials',
+      }),
+    });
+    const jsonResponse = await authenticateResponse.json();
+    return jsonResponse.access_token;
+  } catch (error) {
+    console.error(chalk.red('Error authenticating with M2M token endpoint:', error));
```
stasmaxymov: To be technically correct it is "client credentials grant" instead of "m2m"

packages/core/src/tools/codegen/utils.ts
stasmaxymov: If you talk about `generateSites, generateMetadata`, then yes. I overlooked this part on initial `cli.config` review, sorry for that.

stasmaxymov: I also want to add that it is absolutely OK for `cli` package to reference `core` for types, utils,etc. , because it is agnostic to both frameworks and runtime. The problem is when we bake runtime specific stuff into it

stasmaxymov (changes requested):

packages/core/src/tools/codegen/utils.ts
illiakovalenko: > If you talk about `generateSites, generateMetadata`, then yes. I overlooked this part on initial `cli.config` review, sorry for that.

@stasmaxymov, @art-alexeyenko 
I wouldn't say it's strictly necessary to keep all development-related functionality within the CLI, especially when it comes to framework-specific development tools.
If you look at the Next.js implementation and other libraries, you'll see they often include such tools as part of their core packages as well.
Since these tools are executed only at build time and are structured as submodules, they do not impact the production build. It works totally fine if you don't add `peerDependencies`
Using the proposed approach we will need to introduce some framework specific areas under "cli", so why we can't keep these tools under framework specific package? Especially if you need to add framework specific `peerDependencies`. This was the idea I proposed: to keep `generateSites` and `generateMetadata` under the next.js package, within a `tools` submodule.
I agree that generic "tools" can be moved to "cli" package
@stasmaxymov Let's discuss it more to be aligned moving forward

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts
illiakovalenko: Should it be removed?

packages/core/src/tools/codegen/utils.ts
stasmaxymov: hmm, next.js indeed keeps build specific stuff under. I suppose they do it to minimize packages footprint, so you need just next package installed to get everything. 
Such reasoning is also partially applicable to us, you right that there are drawback in putting frameworks specific stuff to into CLI package, especially if we want it to work globally. So we would need to return back `dev-tools` package and have it to store all "project related" tools and helpers both framework non-framework related, which is also OK solution.

On the other hand, having implicit `peerDependency` is also a smell, which probably can be mitigated by setting it as optional through `peerDependenciesMeta`.

illiakovalenko: @art-alexeyenko I had a discussion with Stas regarding this.
The current approach you took - moving the code extraction (file processing with TypeScript) into the cli package - doesn’t fully address the original problem, since it still requires defining a peer dependency.
However, we definitely don't want to define a _peerDependency_ for the _cli_ package.
Instead, you can move this functionality under the _next.js/tools_ submodule and define _typescript_ as a _peerDependency_ there.
Additionally, use _peerDependenciesMeta_ to mark it as optional.
In this setup, if someone tries to import and execute the code extraction methods without having typescript installed, they should receive a clear error indicating that typescript must be installed.

illiakovalenko (approved):
👍

PR #73, opened 2025-04-29
[cli] Introduce "project" subcommands

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Introduces a simple implementation of subcommands using yargs, enabling the CLI to support commands like
`sitecore-tools project build`
`sitecore-tools project component scaffold`.

## Testing Details

- [ ] Unit Test Added - no unit tests, since subcommand wrapper doesn't produce any logic

## Types of changes

### DIFF
--- a/packages/cli/src/scripts/project/component/scaffold.test.ts
+++ b/packages/cli/src/scripts/project/component/scaffold.test.ts
@@ -1,9 +1,9 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
 import { ComponentTemplateType } from '@sitecore-content-sdk/core/config';
-import * as loadConfigModule from '../../../utils/load-config';
-import proxyquire from 'proxyquire';
 import { SitecoreCliConfig } from '@sitecore-content-sdk/core/src/config';
+import proxyquire from 'proxyquire';
+import * as loadConfigModule from '../../../utils/load-config';
 
 describe('scaffold command', () => {
   let loadCliConfigStub: sinon.SinonStub;
--- a/packages/cli/src/scripts/project/component/scaffold.test.ts
+++ b/packages/cli/src/scripts/project/component/scaffold.test.ts
@@ -1,8 +1,7 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
-import path from 'path';
 import { ComponentTemplateType } from '@sitecore-content-sdk/core/config';
-import * as loadConfigModule from '../utils/load-config';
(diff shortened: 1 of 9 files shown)


packages/cli/src/scripts/project/index.ts
```diff
@@ -0,0 +1,25 @@
...
+/**
+ * @param {Argv} yargs
+ */
+export function builder(yargs: Argv) {
+  return yargs.command({
+    command: 'project',
+    describe: 'Performs project level operations',
+    builder: (_yargs: Argv) => {
+      _yargs = _yargs
+        .command([build, component] as any)
```
MenKNas, line 15: There is no way to type those?

packages/cli/src/scripts/project/index.ts
```diff
@@ -0,0 +1,25 @@
...
+        .strict()
+        .demandCommand(1, 'You need to specify a command to run');
+
+      _yargs = component.builder(_yargs as any);
+
+      return _yargs;
+    },
+    handler: () => {},
```
MenKNas, line 23: Don't we need the handler to do something here?

packages/cli/src/scripts/project/component/scaffold.test.ts
```diff
@@ -1,8 +1,7 @@
 import { expect } from 'chai';
 import sinon from 'sinon';
-import path from 'path';
 import { ComponentTemplateType } from '@sitecore-content-sdk/core/config';
-import * as loadConfigModule from '../utils/load-config';
+import * as loadConfigModule from '../../../utils/load-config';
 import proxyquire from 'proxyquire';
 import { SitecoreCliConfig } from '@sitecore-content-sdk/core/src/config';
```
MenKNas: That's a more generic comment and not strictly related to the PR, maybe we could introduce an import structure and import relative imports in the end after external/internal packages.

packages/cli/src/scripts/project/index.ts
illiakovalenko: It's expected, it's a mandatory option. We don't plan to do anything inside of our wrappers

packages/cli/src/scripts/project/index.ts
illiakovalenko: There are conflicts sometimes when you use typescript and yargs, especially if you want to add types for arguments. We just prefer to workaround that and it's not really critical here

PR #77, opened 2025-05-05
[nextjs] Add helper for TipTap rich text format

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Adds a package with TipTap-specific logic, that contains a helper to convert TipTap Rich Text JSON into HTML.
Separate package is neccessary to not mix TipTap library dependencies into Content SDK core or other packages - and it can be enhanced later with components and other utils.

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/richtext/src/utils/parse-tiptap-json.ts
```diff
@@ -0,0 +1,22 @@
+import { generateHTML } from '@tiptap/html';
+import { JSONContent, Extensions } from '@tiptap/core';
+import { debug } from '@sitecore-content-sdk/core';
+import StarterKit from '@tiptap/starter-kit';
+
+export const parseTiptapJSON = (content: JSONContent[], extensions: Extensions = []) => {
+  // prevent warns when StarterKit is passed in extensions
+  const nonDuplicateExtensions = Array.from(new Set([StarterKit, ...extensions]));
```
art-alexeyenko: It would be generally better to extend StarterKit and add new extensions. Unfortunately, extra extensions are not being applied up when doing so, and documentation on that part is lacking.
This current approach works though.

stasmaxymov: Have you test current approach by providing customized extension, like `Heading.configure(...)`

packages/richtext/src/utils/parse-tiptap-json.ts
```diff
@@ -0,0 +1,22 @@
...
+  let result = '';
+  try {
+    result = generateHTML(
+      {
```
stasmaxymov: AFAIK `type: doc` is part of field content, so you don't need to provide it, e.g. here is the example of rich text field in CS:
image

packages/richtext/src/utils/parse-tiptap-json.ts
```diff
@@ -0,0 +1,22 @@
...
+        type: 'doc',
+        content,
+      },
+      nonDuplicateExtensions
+    );
+  } catch (e) {
```
stasmaxymov, line 17: we should rethrow error, instead of swallowing it.

packages/richtext/src/utils/parse-tiptap-json.ts
stasmaxymov: IF yes, what have precedence?

packages/richtext/src/utils/parse-tiptap-json.ts
```diff
@@ -0,0 +1,22 @@
+import { generateHTML } from '@tiptap/html';
+import { JSONContent, Extensions } from '@tiptap/core';
+import { debug } from '@sitecore-content-sdk/core';
+import StarterKit from '@tiptap/starter-kit';
+
+export const parseTiptapJSON = (content: JSONContent[], extensions: Extensions = []) => {
```
stasmaxymov: Add jsdoc for the comment, also it would be good that we use starterkit by default

packages/richtext/package.json
```diff
@@ -0,0 +1,87 @@
...
+        "typescript": "~5.7.3"
+    },
+    "peerDependencies": {
+        "@tiptap/core": "^2.11.7",
+        "@tiptap/html": "^2.11.7",
+        "@tiptap/starter-kit": "^2.11.7",
+        "typescript": "~5.7.3"
+    },
+    "peerDependenciesMeta": {
+        "@tiptap/core": {
```
stasmaxymov: what will happen it user installs "richtext" package, but skip installing tiptap, and call `parseTiptapJSON`

packages/richtext/src/utils/parse-tiptap-json.ts
stasmaxymov: I also think we should rename a method to reflect that it returns HTML, cause it not self explanatory right now I think original `generateHtml` is good enough or `generateRichTextHtml`, and we probably don't need to specify framework underneath to have more or less closed loop, e.g. it will be something like this:
```
const client = new ContentClient();
const blog =  client.getBlog(...);
const html = generateHtml(blog.content);
```

stasmaxymov (changes requested):
Please see comments

packages/richtext/package.json
art-alexeyenko: I'll check, probably need to remove Meta here so that peers install in any case

packages/richtext/src/utils/parse-tiptap-json.ts
art-alexeyenko: From the looks of it, `configure` is used to enable-disable extensions from `StarterKitOptions`. If docs are to be believed, extra extensions are added via `.extend()` call, that's what I tested:
https://tiptap.dev/docs/editor/extensions/custom-extensions/extend-existing

But then the custom extension wasn't picked up.

packages/richtext/src/utils/parse-tiptap-json.ts
art-alexeyenko: We do use StarterKit by default. Or do you mean that `extensions` should override the default StarterKit?

packages/richtext/package.json
art-alexeyenko: Works fine with Meta and without - so I'll remove it. This is a monorepo setup of course, the standalone app will need tiptap peer dependencies installed to ensure package and the app are on the same page.

stasmaxymov (approved):
Looks good

art-alexeyenko:
Moved to https://github.com/Sitecore/content-sdk/pull/85

PR #87, opened 2025-05-07
[core][nextjs] Enhance customizeability for middleware and client

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Adds some extra exports from nextjs package to make customization of personalization and Sitecore Client a bit easier.
- Add overrides for GraphQL services in client and middleware
- Allow expanding experience params in personalize middleware

## Testing Details
- [x] Unit Test Added - TBA

## Types of changes
- [x] Chore?


packages/nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -12,34 +12,12 @@ import { CloudSDK } from '@sitecore-cloudsdk/core/server';
...
-   * Timeout (ms) for CDP request. Default is 400.
-   */
-  timeout?: number;
-};
-
 export type PersonalizeMiddlewareConfig = MiddlewareBaseConfig &
   SitecoreConfig['api']['edge'] &
   SitecoreConfig['personalize'] & {
     clientFactory: GraphQLRequestClientFactory;
+    personalizeService?: GraphQLPersonalizeService;
```
art-alexeyenko, line 18: not 100% on board with this. On one hand - override is useful, on the other - clientFactory is redundant when providing this override.
But then the alterntive: making the service public, is also a no go, as it would make the class brittle and hard to troubleshoot.

packages/core/src/client/sitecore-client.ts
```diff
@@ -169,16 +169,17 @@ export class SitecoreClient implements BaseSitecoreClient {
...
-    this.siteResolver = this.getSiteResolver();
+    this.siteResolver = initOptions.siteResolver ?? this.getSiteResolver();
 
     const baseServiceOptions = this.getBaseServiceOptions();
 
-    this.layoutService = this.getLayoutService(baseServiceOptions);
-    this.dictionaryService = this.getDictionaryService(baseServiceOptions);
-    this.editingService = this.getEditingService();
-    this.errorPagesService = this.getErrorPagesService();
+    this.layoutService = initOptions.layoutService ?? this.getLayoutService(baseServiceOptions);
```
stasmaxymov: Lets make all methods like `getLayoutService` private so we have a single way to replace dependency

packages/core/src/client/models.ts
```diff
@@ -1,6 +1,20 @@
...
+import { GraphQLDictionaryService } from '../i18n/graphql-dictionary-service';
+import { GraphQLLayoutService } from '../layout/graphql-layout-service';
+import { GraphQLErrorPagesService, GraphQLSitePathService, SiteInfo, SiteResolver } from '../site';
 
+/**
+ * Init options for Sitecore Client that allows you to override services too
+ */
 export type SitecoreClientInit = Omit<SitecoreConfig, 'multisite' | 'redirects' | 'personalize'> & {
   sites: SiteInfo[];
+  layoutService?: GraphQLLayoutService;
```
stasmaxymov: should we combine all the services under single property, like `services`, or `custom` or `overrides`, to not mix it with 'real' settings?
I don't have a strong opinion about that, what do you think about it?

packages/nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -266,28 +272,6 @@ export class PersonalizeMiddleware extends MiddlewareBase {
     };
   }
 
-  protected getExperienceParams(req: NextRequest): ExperienceParams {
```
stasmaxymov, line 269: Hold on, this method was already a protected on, not private.  User could override it.

stasmaxymov: And making it public is not correct either, PersonalizeMiddleware like any other middlware is in fact a command pattern and should have just a single public method -> to execute the middleware

packages/nextjs/src/middleware/personalize-middleware.ts
stasmaxymov, line 18: The same as for client should we put these services overrides under some subproperty?

packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -46,7 +48,8 @@ export class RedirectsMiddleware extends MiddlewareBase {
     super(config);
     // NOTE: we provide native fetch for compatibility on Next.js Edge Runtime
     // (underlying default 'cross-fetch' is not currently compatible: https://github.com/lquixada/cross-fetch/issues/78)
-    this.redirectsService = new GraphQLRedirectsService({ ...config, fetch: fetch });
+    this.redirectsService =
```
stasmaxymov, line 64: The same as for client should we put these services overrides under some subproperty?

packages/nextjs/src/middleware/personalize-middleware.ts
stasmaxymov: we can remove `clientFactory` and keep `personalizeService` as parameter for customization

packages/nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -12,34 +12,12 @@ import { CloudSDK } from '@sitecore-cloudsdk/core/server';
...
-   */
-  timeout?: number;
-};
-
 export type PersonalizeMiddlewareConfig = MiddlewareBaseConfig &
   SitecoreConfig['api']['edge'] &
   SitecoreConfig['personalize'] & {
     clientFactory: GraphQLRequestClientFactory;
+    personalizeService?: GraphQLPersonalizeService;
+    getOverrideExperienceParams?: (req: NextRequest) => ExperienceParams;
```
stasmaxymov: Not sure why we need to introduce this option as user can override getExperienceParams method.

stasmaxymov (changes requested):

stasmaxymov (approved):

PR #88, opened 2025-05-08
[chore] Upgrade 3rd party dependencies

addy-pathania:
- [x] Changelog updated

## Description / Motivation
- Updated all third-party dependencies to their latest versions, excluding `chai`, `chalk`, `sass-alias`, and `inquirer` due to their ESM support requirements.
- The warnings appearing during app initialization (related to `glob`) and runtime (related to `punycode`) stem from older version of `eslint` and its sub-dependencies. We will address these along with the `eslint` upgrade in a separate backlog item.

## Testing Details

- [ ] Unit Test Added

## Types of changes

- [x] Chore


illiakovalenko (changes requested):
core
  * graphql-request
    * current: ^6.1.0 - can we update to latest?
    * latest: 7.1.2
  * tsx
    * Update reference to 4.19.4
  * eslint-plugin-jsdoc
    * current: 48.7.0 - can we update to latest?
    * latest: 50.6.11

templates/nextjs

* sass
  * current: ^1.83.3 - can we update reference to latest?
  * latest: 1.87.0

project wide

* eslint - we discussed, you will revisit and update to latest version without migration to "flat config"
* typescript
    * current: ~5.7.3 - can we update to latest?
    * latest: 5.8.3
* "next" released 15.3.2 yesterday, let's update reference

addy-pathania:
> core
> 
> * graphql-request
>   
>   * current: ^6.1.0 - can we update to latest?
>   * latest: 7.1.2
> * tsx
>   
>   * Update reference to 4.19.4
> * eslint-plugin-jsdoc
>   
>   * current: 48.7.0 - can we update to latest?
>   * latest: 50.6.11
> 
> templates/nextjs
> 
> * sass
>   
>   * current: ^1.83.3 - can we update reference to latest?
>   * latest: 1.87.0
> 
> project wide
> 
> * eslint - we discussed, you will revisit and update to latest version without migration to "flat config"
> * typescript
>   
>   * current: ~5.7.3 - can we update to latest?
>   * latest: 5.8.3
> * "next" released 15.3.2 yesterday, let's update reference

@illiakovalenko 
- We can't upgrade `graphql-request` because its latest version uses ESM. Also regarding the punycode warning, Node.js has its own built-in punycode module, which is now deprecated starting from Node.js 22. The recommended alternative is to use the punycode package from npm, which is not maintained by Node.js.
However, in our case, multiple dependencies including `graphql-request` rely on punycode through nested sub-dependencies. This means we can’t fully get rid of the deprecation warnings until all those dependencies update their usage of punycode. That said, I’ve managed to suppress the warnings, so the app now runs without any warnings.
- I also tried to upgrade `ESLint` and its related dependencies, but ran into a few issues.
First, the `ESLINT_USE_FLAT_CONFIG=false` environment variable no longer seems to work it’s possible that support for it has been dropped.
Next, I tried migrating to the new flat config format, but encountered an error originating from `@rushstack/eslint-patch`, which is used internally by Next.js’s ESLint setup. The problem is that `@rushstack/eslint-patch` doesn’t yet support ESLint 9. So i am not sure if we can upgrade until this issue persists. Here is a open issue related to this https://github.com/microsoft/rushstack/issues/4301

PR #89, opened 2025-05-08
[next.js] Rework and simplify .env

illiakovalenko:
- [x] Changelog updated

## Description / Motivation

The main goal of this PR is to **simplify environment variable management**, align with best practices, and reduce the number of unnecessary or redundant environment variables.

---

- **Introduced `.env.container.example`**  
  Intended for local development against a Sitecore container instance.

- **Introduced `.env.remote.example`**  
  Intended for working with a remote Sitecore instance (XM Cloud).

- **Removed `GRAPH_QL_ENDPOINT` environment variable**  
  No longer required or used.

- **Removed `GRAPH_QL_SERVICE_RETRIES` environment variable**  
  This is not environment-specific and should be defined in the configuration instead.

- **Removed `DISABLE_SSG_FETCH` environment variable**  
  - This setting is rarely used by customers, as most prefer to generate static paths for performance reasons.  
  - In XM Cloud, this is set to `true` by default as the application runs as an editing host.  
  - It can still be configured via the `DISABLE_SSG_FETCH` environment variable or the `disableStaticPaths` config property if needed.

- This PR **does not include any environment variable renaming**.  
  That work will be addressed in the next release. I will create a follow-up task for it.

## Testing Details

- [x] Unit Test Added

## Types of changes


yavorsk (approved):
Looks Good!

PR #90, opened 2025-05-08
[core] [next] [create-sitecore-jss] Passing configuration object to `defineConfig` in sitecore.config is now optional. 

yavorsk:
- [x] Changelog updated

## Description / Motivation

Main motivation of the PR is to simplify the starter app's sitecore.config:
- passing SitecoreConfigInput object to the `defineConfig` function is now optional.
- If nothing is provided to the `defineConfig` function, it will populate the required configuration properties from the environment variables defined in .env. Users can decide whether they want to override those default falback values via passing the SitecoreConfigInput object
- sitecore.config.example is added to the starter app to serve as an example
 
## Testing Details

- [x] Unit Test Added

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts
```diff
@@ -1,33 +1,6 @@
 import { defineConfig } from '@sitecore-content-sdk/nextjs/config';
 /**
  * @type {import('@sitecore-content-sdk/nextjs/config').SitecoreConfig}
+ * See documentation about defineConfig here https://doc.sitecore.com/xmc/en/developers/content-sdk/the-sitecore-configuration-file.html
```
illiakovalenko:
```suggestion
 * See the documentation for `defineConfig`:
 * https://doc.sitecore.com/xmc/en/developers/content-sdk/the-sitecore-configuration-file.html
```

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.example
sc-hatimfarhathussain, line 1: would it be better the if naming of the example file should be something like this `sitecore.config.ts.example` , so users can identify it as a ts file. Just suggestion only.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.example
```diff
@@ -0,0 +1,33 @@
+import { defineConfig } from '@sitecore-content-sdk/nextjs/config';
+/**
+ * @type {import('@sitecore-content-sdk/nextjs/config').SitecoreConfig}
```
illiakovalenko:
```suggestion
 * @type {import('@sitecore-content-sdk/nextjs/config').SitecoreConfig}
 * See the documentation for `defineConfig`:
 * https://doc.sitecore.com/xmc/en/developers/content-sdk/the-sitecore-configuration-file.html
```

packages/core/src/config/models.ts
illiakovalenko, line 1: In some cases we know that specific environment variables can be used across of any framework
Can we mention how to control personalize timeouts and editing secret (anything else?) via env variables? tsdoc should mention that

packages/nextjs/src/config/define-config.ts
```diff
@@ -4,6 +4,54 @@ import {
...
+export const getNextFallbackConfig = (config?: SitecoreConfigInput): SitecoreConfigInput => {
+  return {
+    ...config,
+    api: {
+      ...config?.api,
+      edge: {
+        ...config?.api?.edge,
+        contextId:
+          config?.api?.edge?.contextId ||
+          process.env.SITECORE_EDGE_CONTEXT_ID ||
```
illiakovalenko: Check for `SITECORE_EDGE_CONTEXT_ID` is not needed here, right? Since we should have this fallback in `core`

packages/nextjs/src/config/define-config.ts
```diff
@@ -4,6 +4,54 @@ import {
...
+          process.env.NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID ||
+          '',
+        clientContextId:
+          config?.api?.edge?.clientContextId || process.env.NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID,
+        edgeUrl:
+          config?.api?.edge?.edgeUrl ||
+          process.env.SITECORE_EDGE_URL ||
```
illiakovalenko: Check for `SITECORE_EDGE_URL` is not needed here, right? Since we should have this fallback in `core`

illiakovalenko (changes requested):

packages/nextjs/src/config/define-config.ts
```diff
@@ -4,6 +4,54 @@ import {
...
+        ...config?.api?.local,
+        apiKey: config?.api?.local?.apiKey || process.env.NEXT_PUBLIC_SITECORE_API_KEY || '',
+        apiHost: config?.api?.local?.apiHost || process.env.NEXT_PUBLIC_SITECORE_API_HOST || '',
+      },
+    },
+    defaultSite: config?.defaultSite || process.env.NEXT_PUBLIC_SITECORE_SITE_NAME,
+    defaultLanguage: config?.defaultLanguage || process.env.NEXT_PUBLIC_DEFAULT_LANGUAGE,
+    multisite: {
+      ...config?.multisite,
+      enabled: config?.multisite?.enabled ?? true,
```
illiakovalenko: we should have a fallback to `true` in `core` package, right?

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.example
yavorsk: we have the `sitecore.config.ts` in the starter as well. but you are right, I will update :)

packages/nextjs/src/config/define-config.ts
yavorsk: hm yea we have the fallback in `core`. I left it in next as well to preserve the precedence of the two env variables, but I you are correct i guess it is not needed.

packages/nextjs/src/config/define-config.ts
yavorsk: fallback in core is `false`. that's why i included it in nextjs

packages/nextjs/src/config/define-config.test.ts
```diff
@@ -64,4 +64,322 @@ describe('defineConfig', () => {
       });
     });
   });
+
+  describe('getNextFallbackConfig', () => {
```
art-alexeyenko: instead of getNextFallbackConfig it's better to test defineConfig here - since this is what is used by end user.
Need to ensure that nextjs changes and core's `defineConfig` work well together, with everything that needs to be overriden, being overriden.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts.example
```diff
@@ -0,0 +1,35 @@
...
+  },
+  defaultSite: process.env.NEXT_PUBLIC_SITECORE_SITE_NAME,
+  defaultLanguage: process.env.NEXT_PUBLIC_DEFAULT_LANGUAGE || 'en',
+  editingSecret: process.env.JSS_EDITING_SECRET,
+  redirects: {},
+  multisite: {
+    enabled: true,
+    useCookieResolution: () => process.env.VERCEL_ENV === 'preview',
+  },
+  personalize: {
```
art-alexeyenko, line 33: please add `enabled` to this and `redirects` section.
Overall I think we can go wild with the example and show all the different config options available, unless you agreed differently with Christian.

art-alexeyenko (changes requested):

packages/nextjs/src/config/define-config.test.ts
yavorsk: hm but we are unittesting here, we should not rely on what's happening in core. Even if we use `defineConfig` in the tests we would need to mock core's `defineConfig` in that case. Unless I am missunderstanding your comment? 
And since all logic is happening in `getNextFallbackConfig` I decided to directly test that.

packages/core/src/config/models.ts
yavorsk: makes sense. i am wondering how to handle the general edge env vars, seems not accurate to mention them here.

illiakovalenko: yes, you can't do that

CHANGELOG.md
```diff
@@ -16,6 +16,8 @@ Our versioning strategy is as follows:
 
 * `[core]` Introduce get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#75](https://github.com/Sitecore/content-sdk/pull/75))
 * `[cli]` Introduce "project" subcommands ([#73](https://github.com/Sitecore/content-sdk/pull/73))
+* `[core]``[next]``[create-sitecore-jss]` Passing configuration object to defineConfig in sitecore.config is now optional. Introduced sitecore.config.ts.example ([#90](https://github.com/Sitecore/content-sdk/pull/90))
```
illiakovalenko: If you will not add whitespace between backticks it will break preview

```suggestion
* `[core]` `[next]` `[create-sitecore-jss]` Passing configuration object to `defineConfig` in _sitecore.config_ is now optional. Introduced _sitecore.config.ts.example_ ([#90](https://github.com/Sitecore/content-sdk/pull/90))
```

packages/core/src/config/define-config.test.ts
```diff
@@ -148,4 +149,42 @@ describe('define-config', () => {
       524,
     ]);
   });
+
+  describe('getFallbackConfig', () => {
```
illiakovalenko, line 153: There are tests above that are verifying `getFallbackConfig` as well. Can it be consolidated?

packages/create-sitecore-jss/src/templates/nextjs/sitecore.config.ts.example
```diff
@@ -0,0 +1,39 @@
...
+  redirects: {
+    enabled: true,
+    locales: ['en'],
+  },
+  multisite: {
+    enabled: true,
+    useCookieResolution: () => process.env.VERCEL_ENV === 'preview',
+  },
+  personalize: {
+    enabled: true,
```
illiakovalenko: personalization shouldn't be enabled by default in all the environments, since during development it should be disabled, so let's stick to our old approach and don't set it explicitly

packages/nextjs/src/config/define-config.ts
```diff
@@ -4,6 +4,48 @@ import {
   SitecoreConfigInput as SitecoreConfigInputCore,
 } from '@sitecore-content-sdk/core/config';
 
+/**
+ * Provides default NextJs initial values from env variables for SitecoreConfig
+ * @returns default nextjs input config
```
illiakovalenko: tsdoc warning, please add optional `config` param here

packages/nextjs/src/config/define-config.ts
illiakovalenko: nope, this is not true, in core it fallsback to `true`:
https://github.com/Sitecore/content-sdk/blob/feature/JSS-5427/packages/core/src/config/models.ts#L121-L124

packages/nextjs/src/config/define-config.test.ts
illiakovalenko: yes, we don't need to re-test `defineConfig` from `core`
but I understand what exactly Artem would like to apply
you could mock `core` `defineConfig` and instead of executing and testing `getNextFallbackConfig` you could run nextjs `defineConfig`
Can you apply that? shouldn't really affect much

illiakovalenko (changes requested):
Looks really great! See my new comments, it should be a final round

packages/nextjs/src/config/define-config.ts
```diff
@@ -4,6 +4,48 @@ import {
...
+    ...config,
+    api: {
+      ...config?.api,
+      edge: {
+        ...config?.api?.edge,
+        contextId:
+          config?.api?.edge?.contextId || process.env.NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID || '',
+        clientContextId:
+          config?.api?.edge?.clientContextId || process.env.NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID,
+        edgeUrl: config?.api?.edge?.edgeUrl || process.env.NEXT_PUBLIC_SITECORE_EDGE_URL,
```
illiakovalenko, line 23: please, add unit tests to cover `edgeUrl`

packages/nextjs/src/config/define-config.test.ts
```diff
@@ -1,27 +1,374 @@
...
+      defineConfigModule.defineConfig(defaultConfig());
+      const resultConfig = defineConfigCoreStub.getCalls()[0].args[0];
+      expect(resultConfig.multisite?.enabled).to.be.undefined;
+    });
+
+    it('should be able to override default value of enabled', () => {
+      defineConfigModule.defineConfig({ ...defaultConfig(), multisite: { enabled: false } });
+      const resultConfig = defineConfigCoreStub.getCalls()[0].args[0];
+      expect(resultConfig.multisite?.enabled).to.be.false;
+    });
```
illiakovalenko: this is related to `enabled` property, right?
let's group it under `describe('enabled')`

packages/nextjs/src/config/define-config.test.ts
```diff
@@ -1,27 +1,374 @@
...
+      });
+
+      it('should use the value from the config', () => {
+        defineConfigModule.defineConfig({
+          ...defaultConfig(),
+          personalize: { scope: 'custom-scope' },
+        });
+        const resultConfig = defineConfigCoreStub.getCalls()[0].args[0];
+        expect(resultConfig.personalize?.scope).to.equal('custom-scope');
+      });
```
illiakovalenko: these tests are specific to `scope`, let's group it under `describe('scope')`

packages/nextjs/src/config/define-config.test.ts
```diff
@@ -1,27 +1,374 @@
...
+
+      it('should use the env var if config value not present', () => {
+        defineConfigModule.defineConfig(defaultConfig());
+        const resultConfig = defineConfigCoreStub.getCalls()[0].args[0];
+        expect(resultConfig.personalize?.scope).to.equal('custom-env-scope');
```
illiakovalenko: same as above, let's group
since if new properties will appear it will be required to restructure these tests

packages/nextjs/src/config/define-config.test.ts
```diff
@@ -1,27 +1,374 @@
...
-        expect(config.disableStaticPaths).to.equal(false);
+        defineConfigModule.defineConfig(defaultConfig());
+        const resultConfig = defineConfigCoreStub.getCalls()[0].args[0];
+        expect(resultConfig.disableStaticPaths).to.equal(false);
       });
 
       it('should use the value from the config', () => {
-        const config = defineConfig({
+        defineConfigModule.defineConfig({
           disableStaticPaths: false,
```
illiakovalenko: let's set it to `true` so we will test this edge case
because, currently `false` is our default value as well

illiakovalenko (approved):
Looks good to me 👍
I left few additional comments, but I approve in advance

PR #91, opened 2025-05-08
[nextjs] unify component level getServerSide/getStaticProps methods

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Removes getServerSideProps and getStaticProps methods on component level and replaces them with a single getComponentServerProps method.
Oftentimes the logic would need to be duplicated betweent ServerSide and Static component props method, and now this part is simplified.

Next JS contexts are still passed into component level methods. In case different logic is required depending on ServerSidePropsContext or StaticPropsContext, we expose a `isServerSidePropsContext` helper method that can be imported from `@sitecore-content-sdk/nextjs/utils`.

## Testing Details
- [x] Unit Test Added

## Types of changes


illiakovalenko:
@art-alexeyenko it will be a part of our next release, it shouldn't appear in 0.2

CHANGELOG.md
```diff
@@ -35,6 +35,8 @@ Our versioning strategy is as follows:
 * `[core]` SXA Form can't fire CloudSDK events due to initialization error ([#63](https://github.com/Sitecore/content-sdk/pull/63)):
   * Form utilities have been moved from `@sitecore-content-sdk/core/form` to the root of `@sitecore-content-sdk/core`. Update your imports to reflect this change if you are referencing these utilities.
 * `[nextjs]` Update React to version 19 and Next JS to version 15 ([#76](https://github.com/Sitecore/content-sdk/pull/76))
+* `[nextjs]` Component-level `getServerSideProps` and `getStaticProps` methods have been replaced by a single `getComponentServerProps` method for simplicity.
```
illiakovalenko, line 80: I'm not sure that `getComponentServerProps` is the best naming choice, as it could easily be confused with Next.js's `getServerSideProps`, which has a specific and well-defined meaning.

Here are a couple of alternative suggestions:

* getComponentProps - Clean and generic. It's a good fit if we're abstracting away the rendering strategy (SSG/SSR/etc.).

* resolveComponentProps - Also a valid option, though personally I prefer getComponentProps for its simplicity and clarity.

Happy to hear thoughts or other suggestions!

packages/nextjs/src/sharedTypes/component-props.ts
```diff
@@ -21,27 +21,20 @@ export type ComponentPropsFetchFunction<NextContext, FetchedProps = unknown> = {
...
- * Shape of getServerSideProps function on component level
+ * Common server component props used in both SSG and SSR modes
  */
-export type GetServerSideComponentProps = ComponentPropsFetchFunction<GetServerSidePropsContext>;
-
-/**
- * Shape of getStaticProps function on component level
- */
-export type GetStaticComponentProps = ComponentPropsFetchFunction<GetStaticPropsContext>;
+export type GetComponentServerProps = ComponentPropsFetchFunction<
```
illiakovalenko: See my explanation regarding "server" part in naming
```suggestion
export type GetComponentProps = ComponentPropsFetchFunction<
```

packages/nextjs/src/sharedTypes/component-props.ts
```diff
@@ -21,27 +21,20 @@ export type ComponentPropsFetchFunction<NextContext, FetchedProps = unknown> = {
 };
 
 /**
- * Shape of getServerSideProps function on component level
+ * Common server component props used in both SSG and SSR modes
```
illiakovalenko: Suggestion of how to improve and extend the description to make it more clear

```suggestion
 * Defines the shape of a data-fetching function used at the component level.
 *
 * This function can be used in both **Server-Side Rendering (SSR)** and **Static Site Generation (SSG)** contexts.
 * It enables component-specific data loading that integrates with Next.js rendering flows.
 *
 * The returned props are passed directly to the component at render time.
```

packages/nextjs/src/sharedTypes/component-props.ts
```diff
@@ -21,27 +21,20 @@ export type ComponentPropsFetchFunction<NextContext, FetchedProps = unknown> = {
...
  * Represents a nextjs component import
  */
 export type NextjsJssComponent = ReactJssComponent & {
   /**
-   * function for component level data fetching in SSR mode
-   */
-  getServerSideProps?: GetServerSideComponentProps;
-  /**
-   * function for component level data fetching in SSG mode
+   * function for component level data fetching in both SSR and SSG
```
illiakovalenko: this description should be aligned with the description above

packages/nextjs/src/services/component-props-service.ts
```diff
@@ -34,28 +34,15 @@ export class ComponentPropsService {
...
-        placeholders: layoutData.sitecore.route?.placeholders,
-        fetchFunctionFactory,
-        layoutData,
-        context,
-      });
-      return await this.execRequests(requests);
-    }
+    const fetchFunctionFactory = async (componentName: string) =>
+      (await this.getModule(components, componentName))?.getComponentServerProps;
+    const requests = await this.collectRequests({
```
illiakovalenko, line 33: `collectRequests` method accepts generic type `NextContext`, currently you pass `context` but within the `collectRequests` function it is accessible as `any`

packages/nextjs/src/services/component-props-service.ts
```diff
@@ -34,28 +34,15 @@ export class ComponentPropsService {
...
-
-      const requests = await this.collectRequests({
-        placeholders: layoutData.sitecore.route?.placeholders,
-        fetchFunctionFactory,
-        layoutData,
-        context,
-      });
-      return await this.execRequests(requests);
-    }
+    const fetchFunctionFactory = async (componentName: string) =>
```
illiakovalenko: Can we reduce complexity now by removing `fetchFunctionFactory` and instead just call `getModule` within the `collectRequests` method?

illiakovalenko (changes requested):
Great improvements!
I've added some comments below with suggestions to simplify and clarify the solution further

CHANGELOG.md
art-alexeyenko: There's a specific reason I included `server` part in there - these methods will only be called server-side, and I don't want anyone confused, thinking these methods will affect client-side components too. 
`getServerSideProps` is a reserved name, and `getComponentServerProps` sounds different enough IMO.

packages/nextjs/src/sharedTypes/component-props.ts
art-alexeyenko: let's have this convo in the thread above

CHANGELOG.md
illiakovalenko: In Pages router we don't have server-side components, only client-side
@stasmaxymov what is your opinion on this topic?

packages/nextjs/src/services/component-props-service.ts
art-alexeyenko: I thought about it, decided to leave it as is for simplicity:
- `components` closure is used here, passed from `fetchComponentProps`
- This also allows for logical sepration into `collectRequests` and `execRequests` later

CHANGELOG.md
stasmaxymov: I agree with Artem, that it is good to specify server part, exactly because components in Pages router always render on both client and server side,  but our feature to fetch additional data for components - executes only on server. We strip these functions from a client bundle.

illiakovalenko: Thanks for sharing your thoughts, everyone. In this case, feel free to proceed with the current naming

illiakovalenko (approved):
Looks great! Much simpler and cleaner now

CHANGELOG.md
```diff
@@ -77,6 +77,8 @@ Our versioning strategy is as follows:
...
```
illiakovalenko, line 80: @art-alexeyenko  Noticed after the merge, it should be moved under "Unreleased" section

PR #97, opened 2025-05-19
[core] Fallback "core" values are not respected when framework specific value is empty & validate resolved config instead of base

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Pull Request resolves two issues:
1) Framework specific empty values should not override fallback values defined in "core"
2) Config validation should validate resolved (merged) config instead of initially provided

Another problem I've found: currently, it's required to set edge context id in order to don't  get validation error in browser, but this is not right, since we treat client context id as optional, so we need to revisit. I will create a follow up, since it's not related to initial problem

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/core/src/config/define-config.ts
+++ b/packages/core/src/config/define-config.ts
@@ -1,6 +1,6 @@
 import { SITECORE_EDGE_URL_DEFAULT } from '../constants';
 import { DefaultRetryStrategy } from '../retries';
-import { SitecoreConfig, SitecoreConfigInput } from './models';
+import { DeepPartial, SitecoreConfig, SitecoreConfigInput } from './models';
 
 /**
  * Provides default initial values for SitecoreConfig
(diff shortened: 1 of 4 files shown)


packages/core/src/config/define-config.ts
```diff
@@ -56,26 +56,51 @@ export const getFallbackConfig = (): SitecoreConfig => ({
...
+    } else {
+      result[typedKey] = overrideValue as T[typeof typedKey];
+    }
+  }
+
+  return result;
+}
+
+/**
+ * Resolves sitecore config vased on base config and overrides
```
yavorsk:
```suggestion
 * Resolves sitecore config based on base config and overrides
```

yavorsk (approved):
Looks good! see just one typo

PR #99, opened 2025-05-26
[core] Introduce get taxonomy(ies) implementation in the content client

MenKNas:
- [x] Changelog updated

## Description / Motivation
JSS-5343 -> This PR adds the `getTaxonomy` and `getTaxonomies` methods to the Content Client, providing the ability to retrieve taxonomy data from Content Services. It also introduces and exports the related queries and response types 

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -38,7 +38,6 @@ Our versioning strategy is as follows:
   - A new `RobotsMiddleware` class encapsulates HTTP-level logic for generating `robots.txt` responses in Next.js apps.
   - These additions follow the same extensible architecture as existing features enabling custom behavior via service overrides and improving consistency across endpoints.
 * `[sitecore-jss-cli]` Code extraction feature for XMCloud code generation is added ([#71](https://github.com/Sitecore/jss/pull/71))
-
 * `[all]` XM Cloud Content support
   * `[core]` Introduced /content submodule to interact with XM Cloud Content ([#65](https://github.com/Sitecore/content-sdk/pull/65))
   * `[richtext]` Added new `richtext` package to enable Content SDK to work with TipTap Rich Text editors ([#85](https://github.com/Sitecore/content-sdk/pull/85))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,6 +14,7 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
+* `[core]` Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99))
 * `[next.js]` Rework and simplify .env ([#89](https://github.com/Sitecore/content-sdk/pull/89)):
   - Introduced `.env.container.example`  
     - Intended for local development against a Sitecore container instance.
@@ -32,7 +33,7 @@ Our versioning strategy is as follows:
   - A new `RobotsMiddleware` class encapsulates HTTP-level logic for generating `robots.txt` responses in Next.js apps.
   - These additions follow the same extensible architecture as existing features enabling custom behavior via service overrides and improving consistency across endpoints.
 * `[sitecore-jss-cli]` Code extraction feature for XMCloud code generation is added ([#71](https://github.com/Sitecore/jss/pull/71))
-* `[core]` Introduce get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))
+* `[core]` Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))
 * `[all]` XM Cloud Content support
   * `[core]` Introduced /content submodule to interact with XM Cloud Content ([#65](https://github.com/Sitecore/content-sdk/pull/65))
   * `[richtext]` Added new `richtext` package to enable Content SDK to work with TipTap Rich Text editors ([#85](https://github.com/Sitecore/content-sdk/pull/85))
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -16,7 +16,7 @@ Our versioning strategy is as follows:
 
 * `[core]` Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
   - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
-  - Support pagination for the above methods by implementing an internal fetchNext method to handle this in each method
+  - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
   - Introduced respective queries and types
 * `[next.js]` Rework and simplify .env ([#89](https://github.com/Sitecore/content-sdk/pull/89)):
   - Introduced `.env.container.example`  
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,12 +14,11 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* `[core]` 
-  * `[Content - Name TBD] Support`:
-     - Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
-       - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
-       - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
-    - Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))   
+* `[Content - Name TBD] Support`:
+  - Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
+    - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
+    - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
+  - Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))   
 * `[next.js]` Rework and simplify .env ([#89](https://github.com/Sitecore/content-sdk/pull/89)):
   - Introduced `.env.container.example`  
     - Intended for local development against a Sitecore container instance.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,7 +14,10 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* `[core]` Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99))
+* `[core]` Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
+  - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
+  - Support pagination for the above methods by implementing an internal fetchNext method to handle this in each method
+  - Introduced respective queries and types
 * `[next.js]` Rework and simplify .env ([#89](https://github.com/Sitecore/content-sdk/pull/89)):
   - Introduced `.env.container.example`  
     - Intended for local development against a Sitecore container instance.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,10 +14,12 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* `[core]` Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
-  - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
-  - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
-  - Introduced respective queries and types
+* `[core]` 
+  * `[Content - Name TBD] Support`:
+     - Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
+       - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
+       - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
+    - Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))   
 * `[next.js]` Rework and simplify .env ([#89](https://github.com/Sitecore/content-sdk/pull/89)):
   - Introduced `.env.container.example`  
     - Intended for local development against a Sitecore container instance.
@@ -36,7 +38,7 @@ Our versioning strategy is as follows:
   - A new `RobotsMiddleware` class encapsulates HTTP-level logic for generating `robots.txt` responses in Next.js apps.
   - These additions follow the same extensible architecture as existing features enabling custom behavior via service overrides and improving consistency across endpoints.
 * `[sitecore-jss-cli]` Code extraction feature for XMCloud code generation is added ([#71](https://github.com/Sitecore/jss/pull/71))
-* `[core]` Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))
+
 * `[all]` XM Cloud Content support
   * `[core]` Introduced /content submodule to interact with XM Cloud Content ([#65](https://github.com/Sitecore/content-sdk/pull/65))
   * `[richtext]` Added new `richtext` package to enable Content SDK to work with TipTap Rich Text editors ([#85](https://github.com/Sitecore/content-sdk/pull/85))
--- a/packages/core/src/content/content-client.test.ts
+++ b/packages/core/src/content/content-client.test.ts
@@ -244,17 +244,11 @@ describe('content-client', () => {
       requestStub = sinon.stub(client.graphqlClient, 'request');
     });
 
-    it('should retrieve all available taxonomies', async () => {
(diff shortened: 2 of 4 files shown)


CHANGELOG.md
```diff
@@ -14,6 +14,10 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
+* `[core]` Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
```
illiakovalenko: Can we group "content" related features under one entry? 
Let's leave exactly "[Name TBD]", naming is not yet decided
e.g.
* [Name TBD] support:
  * `[core]` Introduced get taxonomy/taxonomies functionality on content client
  * `[core]` Introduced get locales functionality on content client

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
   }
+
+  /**
+   * Retrieves all available taxonomies.
+   *
+   * This method supports optional pagination for both taxonomies and their terms.
+   * - If `taxonomyPageSize` is provided, it limits the number of taxonomies returned per request.
+   * - If `termsPageSize` is provided, it limits the number of terms returned per taxonomy per request.
+   * - If either is omitted, the API defaults are used (returns all results in a single request if no pagination is required).
+   *
```
illiakovalenko: Let's remove this line, it shows warning
```suggestion
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
     return response.manyLocale;
   }
+
+  /**
+   * Retrieves all available taxonomies.
+   *
+   * This method supports optional pagination for both taxonomies and their terms.
+   * - If `taxonomyPageSize` is provided, it limits the number of taxonomies returned per request.
+   * - If `termsPageSize` is provided, it limits the number of terms returned per taxonomy per request.
+   * - If either is omitted, the API defaults are used (returns all results in a single request if no pagination is required).
```
illiakovalenko: All these descriptions should be a part of each appropriate parameter below, it makes sense to group.

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
+
+    return fetchTaxonomiesPage();
+  }
+
+  /**
+   * Retrieves a single taxonomy by its ID.
+   *
+   * This method supports optional pagination for the terms within the taxonomy.
+   * - If `termsPageSize` is provided, it limits the number of terms returned per request.
+   * - If omitted, the API default is used (returns all terms in a single request).
```
illiakovalenko: Should be a part of parameter's description

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
+   *
+   * The returned taxonomy object includes the system metadata, the initial set of terms,
+   * and a `fetchNext` function for retrieving additional terms if more are available.
```
illiakovalenko: It can be a part of "returns" statement below

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
+  ): Promise<ContentItemList<Taxonomy>> {
+    const fetchTaxonomiesPage = async (after: string = ''): Promise<ContentItemList<Taxonomy>> => {
+      debug.content(
+        'Fetching taxonomies (pageSize: %s, after: %s)',
+        taxonomyPageSize ?? 'API Default',
+        after
+      );
+
+      const variables: Record<string, any> = {
+        after: after || '',
```
illiakovalenko: The fallback is defined on parameter level above
```suggestion
        after,
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
+            debug.content(
+              'Fetching terms for taxonomy %s (pageSize: %s, after: %s)',
+              taxonomy.system.id,
+              termsPageSize ?? 'API Default',
+              termsAfter
+            );
+
+            const termVariables: Record<string, any> = {
+              after: after || '',
+              termsAfter: termsAfter || '',
```
illiakovalenko: The fallback is defined on parameter level above
```suggestion
              after,
              termsAfter,
```

packages/core/src/content/taxonomies.ts
```diff
@@ -0,0 +1,189 @@
...
+    results: {
+      /** The terms associated with the taxonomy, in a paginated format. */
+      terms: {
+        results: Term[];
+        cursor?: string | null;
+        hasMore: boolean;
+      };
+      /** The system metadata of the taxonomy. */
+      system: TaxonomySystem;
+    }[];
```
illiakovalenko: I suppose that "results" should represent _Taxonomy[]_ type, right?

packages/core/src/content/taxonomies.ts
```diff
@@ -0,0 +1,189 @@
...
+ * Variables:
+ * - minimumPageSize: The number of taxonomies to retrieve per page.
+ * - termsPageSize: The number of terms to retrieve per page within each taxonomy.
+ * - after: The cursor for fetching the next page of taxonomies.
+ * - termsAfter: The cursor for fetching the next page of terms within a taxonomy.
+ */
+export const GET_TAXONOMIES_QUERY = `
+  query GetAllTaxonomies(
+    $minimumPageSize: Int
+    $termsPageSize: Int
```
illiakovalenko: Can we align naming of these variables, please? You will update it in methods as well
First one can be simplified and named as $pageSize so it's expected that it belongs to taxonomies, second can be left as it's $termsPageSize

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
+        termsAfter: '', // Always provide for API requirements
+      };
+
+      if (taxonomyPageSize !== undefined) {
+        variables.minimumPageSize = taxonomyPageSize;
+      }
+
+      if (termsPageSize !== undefined) {
+        variables.termsPageSize = termsPageSize;
+      }
```
illiakovalenko: Can we introduce a shared function to get variables? so it will reduce amount of code.
This function should be introduced at the start of `fetchTaxonomiesPage` function.
(pageSize, after - belongs to taxonomies variables)
You will be able to reuse it twice

```ts
const createVariables = (after = '', termsAfter = '') => ({
      pageSize,
      after,
      termsAfter,
      termsPageSize,
    });
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +137,183 @@ export class ContentClient {
...
+        id,
+        termsPageSize ?? 'API Default',
+        termsAfter
+      );
+
+      const variables: Record<string, any> = { id, termsAfter };
+
+      if (termsPageSize !== undefined) {
+        variables.termsPageSize = termsPageSize;
+      }
```
illiakovalenko: can it be set on object level? so we don't need to check additionally

illiakovalenko (changes requested):
Looks really good!
Please, review my comments below and I will revisit again later

packages/core/src/content/taxonomies.ts
MenKNas: This is response structure so it's ok to keep it as it is. Inside the client method it is converted into `Taxonomy[]`

CHANGELOG.md
```diff
@@ -14,6 +14,12 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
+* `[core]` 
+  * `[Content - Name TBD] Support`:
```
illiakovalenko:
```suggestion
  * [Content - Name TBD] Support:
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
     const response = await this.get<LocalesQueryResponse>(GET_LOCALES_QUERY);
     return response.manyLocale;
   }
+
+  /**
+   * Retrieves all available taxonomies with optional pagination support.
+   * Note: Only taxonomies are paginated. Terms within each taxonomy are returned in full (no term pagination).
+   * @param {object} [options] - Optional pagination options.
+   * @param {number} [options.pageSize] - Optional. Limits the number of taxonomies returned per page. If not provided, defaults to Content Services API default (20).
```
illiakovalenko: For now let's leave it as TBD, we can't use this name
```suggestion
   * @param {number} [options.pageSize] - Optional. Limits the number of taxonomies returned per page. If not provided, defaults to [Name TBD] API default (20).
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+   * @param {string} [options.after] - Optional. Cursor for pagination. Used to fetch the next page of taxonomies.
+   * @returns A promise that resolves to a paginated list of taxonomies, including pagination metadata (`hasMore`, `cursor`) for developer-controlled pagination.
+   */
+  async getTaxonomies(options?: {
+    pageSize?: number;
+    after?: string;
+  }): Promise<ContentItemList<Taxonomy>> {
+    debug.content(
+      'Fetching taxonomies (pageSize: %s, after: %s)',
```
illiakovalenko: Just to make wording consistent
```suggestion
      'Getting taxonomies (pageSize: %s, after: %s)',
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+      options?.pageSize ?? 'API Default',
+      options?.after ?? ''
+    );
+
+    const variables: Record<string, any> = {};
+    if (options?.pageSize !== undefined) variables.pageSize = options.pageSize;
+
+    variables.after = options?.after ?? '';
```
illiakovalenko: Can we set props in variable, just for simplfication?
Can we set `pageSize` without checking for `undefined`?
```ts
    const variables: Record<string, any> = {
      pageSize: options?.pageSize,
      after: options?.after ?? '',
    };
```

packages/core/src/content/taxonomies.ts
```diff
@@ -0,0 +1,196 @@
...
+ * In getTaxonomies, terms are always returned as a full list (no pagination).
+ * In getTaxonomy, terms may be paginated.
+ */
+export interface TermList {
+  /** The list of terms in the current (or full) page. */
+  results: Term[];
+  /** The cursor to fetch the next page of terms, if available. (Used only in getTaxonomy) */
+  cursor?: string | null;
+  /** Indicates whether more terms are available after the current page. (Used only in getTaxonomy) */
+  hasMore: boolean;
```
illiakovalenko, line 24: Do we suppose that `hasMore` is optional as well? I noticed you check whether it's present

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+
+    const response = await this.get<TaxonomiesQueryResponse>(GET_TAXONOMIES_QUERY, variables);
+    const data = response?.manyTaxonomy;
+
+    return {
+      results: (data?.results ?? []).map((taxonomy) => ({
+        system: taxonomy.system,
+        terms: {
+          results: taxonomy.terms?.results ?? [],
+          cursor: undefined,
```
illiakovalenko: Can be removed
```suggestion
```

packages/core/src/content/taxonomies.ts
```diff
@@ -0,0 +1,196 @@
...
+  /** The minimum page size used for fetching terms (if provided). (Used only in getTaxonomy) */
+  minimumPageSize?: number;
```
illiakovalenko: This one is not present, right?
```suggestion
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+          hasMore: false,
+        },
+      })),
+      cursor: data?.cursor,
+      hasMore: data?.hasMore ?? false,
```
illiakovalenko, line 167: Does it mean that `hasMore` is optional? Should it be optional in our interfaces as well?

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+    const response = await this.get<TaxonomiesQueryResponse>(GET_TAXONOMIES_QUERY, variables);
+    const data = response?.manyTaxonomy;
+
+    return {
+      results: (data?.results ?? []).map((taxonomy) => ({
+        system: taxonomy.system,
+        terms: {
+          results: taxonomy.terms?.results ?? [],
+          cursor: undefined,
+          hasMore: false,
```
illiakovalenko: Can be removed as well, if it's optional

packages/core/src/content/taxonomies.ts
```diff
@@ -0,0 +1,196 @@
...
+};
+
+/**
+ * Represents a list of terms within a taxonomy.
+ * In getTaxonomies, terms are always returned as a full list (no pagination).
+ * In getTaxonomy, terms may be paginated.
+ */
+export interface TermList {
+  /** The list of terms in the current (or full) page. */
+  results: Term[];
```
illiakovalenko, line 20: Do we expect that `results` can be empty? In both Taxonomies and Terms

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+    variables.after = options?.after ?? '';
+
+    const response = await this.get<TaxonomiesQueryResponse>(GET_TAXONOMIES_QUERY, variables);
+    const data = response?.manyTaxonomy;
+
+    return {
+      results: (data?.results ?? []).map((taxonomy) => ({
+        system: taxonomy.system,
+        terms: {
+          results: taxonomy.terms?.results ?? [],
```
illiakovalenko: since we don't support pagination for terms, probably it would be more user friendly if terms results can be set in `terms` property, right? I don't need to access `results` and especially `cursor` or `hasMore`

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +136,89 @@ export class ContentClient {
...
+    debug.content(
+      'Getting taxonomy for id: %s (termsPageSize: %s, termsAfter: %s)',
+      id,
+      terms?.pageSize ?? 'API Default',
+      terms?.after ?? ''
+    );
+
+    const variables: Record<string, any> = { id };
+    if (terms?.pageSize !== undefined) variables.termsPageSize = terms.pageSize;
+    if (terms?.after !== undefined) variables.termsAfter = terms.after;
```
illiakovalenko: Can it be simplified?
```suggestion
    const variables: Record<string, any> = {
      id,
      termsPageSize: terms?.pageSize,
      termsAfter: terms?.after,
    };
```

illiakovalenko (changes requested):

packages/core/src/content/taxonomies.ts
MenKNas: `hasMore` will always be present in the response so it should stay required in the type

packages/core/src/content/content-client.ts
MenKNas: I have implemented this fallback option in the case `hasMore`  for some weird reason comes back as `undefined` or `null` from the API. In this case it would be better for anyone using our methods to have a default `false`  as the value of  `hasMore`

packages/core/src/content/taxonomies.ts
MenKNas: It can be empty

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +135,83 @@ export class ContentClient {
...
+   * @returns A promise that resolves to an object containing taxonomies, their terms, and pagination info.
+   */
+  async getTaxonomies(options?: { pageSize?: number; after?: string }) {
+    debug.content(
+      'Getting taxonomies (pageSize: %s, after: %s)',
+      options?.pageSize ?? 'API Default',
+      options?.after ?? ''
+    );
+
+    const variables: Record<string, any> = {
```
illiakovalenko: `any` is not allowed to be used, you can just remove explicit type
```suggestion
    const variables = {
```

packages/core/src/content/content-client.ts
```diff
@@ -129,4 +135,83 @@ export class ContentClient {
...
+    };
+  }): Promise<Taxonomy | null> {
+    debug.content(
+      'Getting taxonomy for id: %s (termsPageSize: %s, termsAfter: %s)',
+      id,
+      terms?.pageSize ?? 'API Default',
+      terms?.after ?? ''
+    );
+
+    const variables: Record<string, any> = {
```
illiakovalenko: any is not allowed to be used, you can just remove explicit type
```suggestion
    const variables = {
```

packages/core/src/content/content-client.test.ts
```diff
@@ -226,4 +227,220 @@ describe('content-client', () => {
...
+      const callArgs = requestStub.getCall(0).args;
+      expect(callArgs[0]).to.equal(GET_TAXONOMIES_QUERY);
+      expect(callArgs[1]).to.deep.equal({ pageSize: 2, after: 'cursor-taxonomies' });
+      expect(result.results[0].system.name).to.equal('Taxonomy 2');
+      expect(result.results[0].terms[0].name).to.equal('Term 2');
+      expect(result.cursor).to.equal('another-cursor');
+      expect(result.hasMore).to.be.false;
+    });
+
+    it('should handle empty or null responses', async () => {
```
illiakovalenko, line 329: Let's put it under "describe" and add one extra test to cover `null | undefined` _terms.results_

packages/core/src/content/content-client.test.ts
```diff
@@ -226,4 +227,220 @@ describe('content-client', () => {
...
+        termsPageSize: undefined,
+        termsAfter: undefined,
+      });
+      expect(result?.system.id).to.equal(taxonomyId);
+      expect(result?.terms.results[0].name).to.equal('Term 2');
+      expect(result?.terms.cursor).to.be.null;
+      expect(result?.terms.hasMore).to.be.false;
+    });
+
+    it('should handle null taxonomy response', async () => {
```
illiakovalenko, line 508: Let's put it under "describe" and add one extra test to cover `null | undefined` _terms.results_

illiakovalenko (approved):
Looks very good!
I left few minor comments, please, have a look
Feel free to merge then

PR #102, opened 2025-05-26
[cli] Content Services CLI Authentication: Basic commands

addy-pathania:
- [x] Changelog updated

## Description / Motivation
This PR introduces authentication to the CLI, allowing users to authenticate against Sitecore Content Services using the client credentials flow. It includes 
- persistent credential storage, 
- active tenant tracking
- support for multi-tenant management.

- Implemented Commands
   - `auth login`
      - Authenticates using client credentials only when `clientSecret` is present and stores:
         - `auth.json`: Access token info
         - `info.json`: Tenant metadata
         - `settings.json`: Tracks active tenant
         - Validates JWT claims against provided `tenantId` and `organizationId`.

   - `auth status`
      - Shows the currently active tenant registered under `~/.sitecore/sitecore-tools/settings.json`.
      - Automatically renews the token if expired.

   - `auth logout`
     - Logs out by clearing the active tenant and deleting its `auth.json`.

   - `auth list`
     - Lists all known tenants located under `~/.sitecore/sitecore-tools/{tenantId}`

- Token and tenant info stored under `~/.sitecore/sitecore-tools/{tenantId}-info.json`
- Auth metadata is under `~/.sitecore/sitecore-tools/{tenantId}-auth.json`
- `settings.json` tracks the active tenant globally under `~/.sitecore/sitecore-tools/settings.json`
- Token expiry is checked and token renewed automatically if expired.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/cli/src/scripts/auth/list.ts
+++ b/packages/cli/src/scripts/auth/list.ts
@@ -12,12 +12,12 @@ export const list: CommandModule = {
       return;
     }
 
-    console.log('\nKnown tenants:\n');
+    console.log('\n Known tenants:\n');
 
     tenants.forEach((tenant, index) => {
       console.log(`Tenant ${index + 1}:`);
       console.log(`  Tenant ID       : ${tenant.tenantId}`);
-      console.log(`  Tenant Name     : ${tenant.tenantName}`);
+      console.log(`  Tenant Name     : ${tenant.tenantName || 'N/A'}`);
       console.log(`  Organization ID : ${tenant.organizationId}`);
       console.log(`  Client ID       : ${tenant.clientId}`);
       console.log(`  Authority       : ${tenant.authority || 'N/A'}`);
--- a/packages/cli/src/scripts/auth/list.ts
+++ b/packages/cli/src/scripts/auth/list.ts
@@ -1,0 +1,25 @@
+﻿import { CommandModule } from 'yargs';
+import { getAllTenantsInfo } from '../../utils/auth/tenant-store';
+
+export const list: CommandModule = {
+  command: 'list',
+  describe: 'List all tenants',
+  handler: async () => {
+    const tenants = getAllTenantsInfo();
+
+    if (tenants.length === 0) {
+      console.info('\n No tenant info found.');
+      return;
+    }
+
+    console.info('\n Known tenants:');
+    console.table(
+      tenants.map((tenant) => ({
+        tenant_id: tenant.tenantId,
+        tenant_name: tenant.tenantName,
+        organization_id: tenant.organizationId,
+        client_id: tenant.clientId,
+      }))
+    );
+  },
+};
--- a/packages/cli/src/scripts/auth/list.ts
+++ b/packages/cli/src/scripts/auth/list.ts
@@ -8,18 +8,22 @@ export const list: CommandModule = {
     const tenants = getAllTenantsInfo();
 
     if (tenants.length === 0) {
-      console.info('\n No tenant information found.');
+      console.log('\n No tenant information found.');
       return;
     }
 
-    console.info('\n Known tenants:');
-    console.table(
-      tenants.map((tenant) => ({
-        tenant_id: tenant.tenantId,
-        tenant_name: tenant.tenantName,
-        organization_id: tenant.organizationId,
-        client_id: tenant.clientId,
-      }))
-    );
+    console.log('\nKnown tenants:\n');
+
+    tenants.forEach((tenant, index) => {
+      console.log(`Tenant ${index + 1}:`);
+      console.log(`  Tenant ID       : ${tenant.tenantId}`);
+      console.log(`  Tenant Name     : ${tenant.tenantName}`);
+      console.log(`  Organization ID : ${tenant.organizationId}`);
+      console.log(`  Client ID       : ${tenant.clientId}`);
+      console.log(`  Authority       : ${tenant.authority || 'N/A'}`);
+      console.log(`  Audience        : ${tenant.audience || 'N/A'}`);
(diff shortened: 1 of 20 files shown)


packages/cli/src/scripts/auth/login.ts
```diff
@@ -0,0 +1,102 @@
...
+        demandOption: false,
+        describe: 'Intended recipient (audience) of the token, usually the API base URL.',
+      })
+      .option('baseUrl', {
+        type: 'string',
+        demandOption: false,
+        describe: 'Base URL for the API, used to construct the audience.',
+      })
+      .group(
+        ['clientId', 'clientSecret', 'tenantId', 'organizationId, authority, audience', 'baseUrl'],
```
yavorsk: shouldn't 'organizationId, authority, audience' be separate elements of the group array?

packages/cli/src/scripts/auth/models.ts
```diff
@@ -0,0 +1,101 @@
...
+   * Exact ISO timestamp when the token expires
+   */
+  expires_at: string;
+
+  /**
+   * Secret used for client credentials flow and re-authenticate
+   */
+  clientSecret?: string;
+  /*
+    Refresh token used to obtain new access tokens
```
yavorsk:
```suggestion
   * Refresh token used to obtain new access tokens
```

packages/cli/src/utils/auth/renewal.ts
```diff
@@ -0,0 +1,88 @@
...
+  try {
+    if (authInfo.clientSecret) {
+      await renewClientToken(authInfo, tenantInfo);
+    } else {
+      // <TODO>: Implement Device auth token renewal.
+      throw new Error('\n Please use clientSecret for authentication.');
+    }
+    return { tenantId };
+  } catch (err) {
+    console.error(`\n Token renewal failed: ${(err as Error).message}`);
```
yavorsk: I think would be useful to expand this error message with the tenant id and notify user that we are going to logout and clear the active tenant

packages/cli/src/scripts/auth/list.ts
```diff
@@ -0,0 +1,29 @@
...
+
+    if (tenants.length === 0) {
+      console.log('\n No tenant information found.');
+      return;
+    }
+
+    console.log('\nKnown tenants:\n');
+
+    tenants.forEach((tenant, index) => {
+      console.log(`Tenant ${index + 1}:`);
```
yavorsk, line 18: wouldn't console.table be better here? just asking, this seems fine as well

packages/cli/src/scripts/auth/models.ts
```diff
@@ -0,0 +1,101 @@
...
+   * Auth authority/issuer URL (e.g., Sitecore identity endpoint)
+   */
+  authority?: string;
+  /**
+   * Base URL for the target Sitecore Content Management API
+   */
+  baseUrl?: string;
+}
+
+export interface Settings {
```
yavorsk: here 'Settings' seems too generic to me. 'TenantSettings' maybe?

packages/cli/src/scripts/auth/status.ts
```diff
@@ -0,0 +1,27 @@
+﻿import { CommandModule } from 'yargs';
+import { readTenantInfo } from '../../utils/auth/tenant-store';
+import { renewAuthIfExpired } from '../../utils/auth/renewal';
+
+export const status: CommandModule = {
+  command: 'status',
+  describe: 'Show current status of active tenant',
+  handler: async () => {
+    const context = await renewAuthIfExpired();
```
yavorsk, line 9: don't we need to call renewAuthIfExpired in the `list` command as well?

packages/cli/src/utils/auth/renewal.ts
```diff
@@ -0,0 +1,88 @@
...
+
+  const authInfo = await readTenantAuthInfo(tenantId);
+  if (!authInfo) return null;
+
+  const isValid = validateAuthInfo(authInfo);
+  if (isValid) {
+    return { tenantId };
+  }
+  const tenantInfo = await readTenantInfo(tenantId);
+  if (!tenantInfo) return null;
```
yavorsk, line 70: here do we have the same case as above for missing authInfo? in both cases user should login manually again?

yavorsk (changes requested):
Really great work!
see few comments and questions :)

packages/cli/src/scripts/auth/list.ts
addy-pathania: I was using `console.table` initially but the output becomes too wide or cluttered due to long keys/values, its not readable so I am falling back to the regular way.

packages/cli/src/scripts/auth/status.ts
addy-pathania: No, we don't. List command works irrespective of the auth status. Since we are just displaying the info of the known tenants.

packages/cli/src/scripts/auth/login.ts
addy-pathania: good catch.
should be separate strings.

packages/cli/src/utils/auth/renewal.ts
addy-pathania: Yes, I think they should. Do you think differently?

yavorsk (approved):
👍 👍

packages/cli/src/utils/auth/renewal.ts
yavorsk: no, i guess it is fine. I was wondering if we should show different messages for the reason to login again.

addy-pathania:
Closing this PR since this is pointing to dev.
Will open another one pointing to release/beta

PR #105, opened 2025-05-28
[core] Update getLocale(s) based on new schema structure

MenKNas:
- [x] Changelog updated

## Description / Motivation

This PR makes the following updates:

- Updates the GraphQL queries for locales `(GET_LOCALE_QUERY, GET_LOCALES_QUERY)` to reflect the new API response format, where id and label are now nested inside a system object.
- Refactors the corresponding content client methods `(getLocale, getLocales)` to handle the updated structure and return a flat `{ id, label }` object.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,13 +14,13 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* `[core]`
 * `[Content - Name TBD] Support`:
-* - Updated `getLocale/getLocales` based on new GQL Schemas ([#105](https://github.com/Sitecore/content-sdk/pull/105))
   - Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
     - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
     - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
-  - Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))   
+  - Introduced `getLocale/getLocales` functionality on content client: 
+    - Updated `getLocale/getLocales` based on new GQL Schemas ([#105](https://github.com/Sitecore/content-sdk/pull/105))   
+    - Initial implementation of `getLocale/getLocales` ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))
 * `[next.js]` Rework and simplify .env ([#89](https://github.com/Sitecore/content-sdk/pull/89)):
   - Introduced `.env.container.example`  
     - Intended for local development against a Sitecore container instance.
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -19,8 +19,7 @@ Our versioning strategy is as follows:
     - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
     - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
   - Introduced `getLocale/getLocales` functionality on content client: 
-    - Updated `getLocale/getLocales` based on new GQL Schemas ([#105](https://github.com/Sitecore/content-sdk/pull/105))   
-    - Initial implementation of `getLocale/getLocales` ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))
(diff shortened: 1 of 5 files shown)


CHANGELOG.md
```diff
@@ -18,7 +18,9 @@ Our versioning strategy is as follows:
   - Introduced get taxonomy/taxonomies functionality on content client ([#99](https://github.com/Sitecore/content-sdk/pull/99)):
     - Introduced `getTaxonomy` and `getTaxonomies` methods on Content Client class
     - Support pagination for the above methods by implementing an internal `fetchNext` method to handle this in each method
-  - Introduced get locales functionality on content client ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))   
+  - Introduced `getLocale/getLocales` functionality on content client: 
+    - Updated `getLocale/getLocales` based on new GQL Schemas ([#105](https://github.com/Sitecore/content-sdk/pull/105))   
+    - Initial implementation of `getLocale/getLocales` ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))
```
illiakovalenko: Let's add this PR reference to "initial implementation", because we need to add an essential information in Changelog, while API changes is not something interesting for customers, since this is feature is totally new
```suggestion
    - Initial implementation of `getLocale/getLocales` ([#74](https://github.com/Sitecore/content-sdk/pull/74))([#78](https://github.com/Sitecore/content-sdk/pull/78))([#105](https://github.com/Sitecore/content-sdk/pull/105))
```

packages/core/src/content/locales.ts
```diff
@@ -1,38 +1,45 @@
 /**
- * Represents the response structure for a query that retrieves a locale.
+ * Represents the "system" field in a locale response.
+ */
+export type SingleLocale = {
```
illiakovalenko: Can it be just `Locale`?
```suggestion
export type Locale = {
```

packages/core/src/content/locales.ts
```diff
@@ -1,38 +1,45 @@
 /**
- * Represents the response structure for a query that retrieves a locale.
+ * Represents the "system" field in a locale response.
```
illiakovalenko:
```suggestion
 * Represents the locale entity.
```

packages/core/src/content/locales.ts
```diff
@@ -1,38 +1,45 @@
...
+ */
+export type SingleLocale = {
+  id: string;
+  label: string;
+};
+
+/**
+ * Represents a locale result with a system field.
```
illiakovalenko:
```suggestion
 * Represents a locale query result.
```

packages/core/src/content/locales.ts
illiakovalenko, line 1: I added a few comments below to simplify the description. In general, it’s better to avoid mentioning internal structure details if they aren’t critical. Otherwise, any changes in the API (e.g. moving LocaleEntity out of the system field) would require us to constantly maintain and update the documentation - while this level of detail is usually not important or useful for developers consuming the API.

packages/core/src/content/locales.ts
```diff
@@ -1,38 +1,45 @@
...
+ */
+export type LocaleEntry = {
```
illiakovalenko:
```suggestion
export type LocaleItem = {
```

packages/core/src/content/locales.ts
```diff
@@ -1,38 +1,45 @@
 /**
- * Represents the response structure for a query that retrieves a locale.
+ * Represents the "system" field in a locale response.
+ */
+export type SingleLocale = {
+  id: string;
+  label: string;
```
illiakovalenko: Missing tsdoc

illiakovalenko (changes requested):

packages/core/src/content/content-client.ts
```diff
@@ -122,7 +122,7 @@ export class ContentClient {
     debug.content('Getting locale for id: %s', id);
 
     const response = await this.get<LocaleQueryResponse>(GET_LOCALE_QUERY, { id });
-    return response.locale;
+    return response.locale ? response?.locale?.system : null;
```
illiakovalenko: Very minor thing for simplification ;)
```suggestion
    return response.locale?.system || null;
```

packages/core/src/content/locales.ts
```diff
@@ -1,38 +1,47 @@
...
+ */
+export type Locale = {
+  /** The unique identifier of the locale. */
+  id: string;
+  /** The label of the locale. */
+  label: string;
+};
+
+/**
+ * Represents a locale query response.
```
illiakovalenko: Actually not exactly `locale query response`, since we have an interface below that represents `locale query response`
May be can be updated to describe like that?
```suggestion
 * A locale item included in a locale query response.
```

illiakovalenko (approved):
Looks good!
Few minor comments below, feel free to merge then

PR #121, opened 2025-06-12
[core][create-sitecore-jss] Fix for client context id causing runtime error

MenKNas:
- [x] Changelog updated

## Description / Motivation

This PR removes the requirement for `NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID` in client-side code. This issue used to cause runtime errors on the client; now this issue should be fixed by making the use of `clientContextId` (the client-side context ID) optional. A clear console message now alerts users when the context ID is missing.

Key changes:

- _Config validation_ – In define-config.ts, skip clientContextId checks when window exists.
- _GraphQL client_ – In client/utils.ts, fall back to an empty header instead of throwing if the context ID is absent in the browser.
- _Tests_ – Added new tests to cover the updates by introducing a new `utils.test.ts` file.

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/core/src/client/utils.ts
```diff
@@ -21,6 +22,14 @@ export const createGraphQLClientFactory = (options: GraphQLClientOptions) => {
       endpoint: `${options.api.local.apiHost}${options.api.local.path}`,
       apiKey: options.api.local.apiKey,
     };
+  } else if (typeof window !== 'undefined') {
+    // In browser, create a client that won't be used for real requests
+    console.warn(
+      'GraphQL client created without proper configuration - client-side requests may fail'
+    );
+    clientConfig = {
+      endpoint: '/api/graphql', // Dummy endpoint for browser initialization
```
art-alexeyenko, line 31: does this not cause any console errors on client?

packages/core/src/config/define-config.ts
```diff
@@ -116,16 +116,29 @@ const resolveConfig = (base: SitecoreConfig, override: SitecoreConfigInput): Sit
...
+      'Configuration error: at least one API configuration must be specified: ' +
+        'contextId (server-side), clientContextId (client-side), or local API settings (apiHost + apiKey)'
+    );
+  }
+
+  // Warn if middleware features might not work
+  if (!hasEdgeContextId && !hasClientContextId && hasLocalApi) {
+    console.warn(
+      'Warning: Redirects and Personalization middleware require Edge API configuration. ' +
+        'Consider setting SITECORE_EDGE_CONTEXT_ID or NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID.'
```
art-alexeyenko, line 140: will personalization work on client navigation, when NEXT_PUBLIC_..._CONTEXT_ID is not set?

art-alexeyenko (changes requested):
The app fails to build currently in the branch for PR. Please double check on your end (it's related to styles).
Please also check these comments

MenKNas:
> The app fails to build currently in the branch for PR. Please double check on your end (it's related to styles). Please also check these comments

Good catch, that was because the `dev` branch updates weren't pulled here. It should be good now.

packages/core/src/client/utils.ts
MenKNas: I can see there is a log coming from `nextjs\src\Bootstrap.tsx` L:31

`console.error('Client Edge API settings missing from configuration');` 

but that's just normal I guess, it doesn't have to do with the current changes.

packages/core/src/config/define-config.ts
MenKNas: Personalization on client navigation will work as long as you have a server-side `contextId`. As far as I can tell `NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID` is only needed for browser-side stuff. The warning only appears if neither context ID is set, which would break personalization and redirects.

art-alexeyenko (approved):
Approved, looks good.

PR #124, opened 2025-06-12
[core][nextjs][cli] Add component map generation

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Previous Content SDK releases came with a self-managed component-map.ts file that allowed greater flexibility at the cost of convenience. This PR re-introduces automatic map generation, bringing that convenience back.
- component-map.ts was moved from src/lib to .sitecore folder
- Introduce `sitecore-tools project component generate-map` CLI command
- CLI command is called on build out of the box
- App is scaffolded with default component-map.ts, that can be used even if CLI command is disabled on build
- CLI command supports `watch` mode that keeps track of components and re-generates component map upon changes
- CLI command can be configured via sitecore.cli.config and supports the following parameters:
  - `paths` - paths to register components from
  - `exclude` - filter out paths or individual components from being registered, via absolute, relative path, or regex pattern
  - `destination` - write component map to an alternative location
  - `packages` - register custom components from dependency packages, local file system, with support for named component imports. Components will be registered as SXA components (`import * as Component...`) by default.

Side effect of this PR is introducing `/config/cli` submodule. This allows us to use server-side logic, like 'fs', in CLI-related usecases, without breaking the sample app when it imports logic from `/config` in browser context, but throws because things like 'fs' are imported alongside.

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/nextjs/src/config/define-cli-config.ts
```diff
@@ -33,3 +35,17 @@ function addDefaultScaffoldTemplates(cliConfig: SitecoreCliConfigInput) {
 
   cliConfig.scaffold.templates.unshift(defaultTemplate, byocTemplate);
 }
+
+// /**
```
art-alexeyenko: This is a temporary measure - uncommenting this will cause nextjs app to fail at build time. `/config` submodule would try importing 'fs' module via `generateMap` and that will fail when nextjs generates client bundle and pages.
The solution is to separate `config` submodule into `config` and `cliConfig` - the latter being able to use server-side logic. This is in progress, but I didn't want this to block the PR, to get comments on overall implementation

packages/core/src/config/models.ts
```diff
@@ -214,6 +215,19 @@ export type SitecoreCliConfigInput = {
      */
     templates?: ScaffoldTemplate[];
   };
+  /**
+   * Configuration for the `sitecore-tools component generate-map` cli command
+   */
+  generateComponentMap?: {
```
art-alexeyenko: This is not very elegant, and it pains me. `generateComponentMap` ideally should only be configured with a single `command` entry that calls `generateMap` function directly (like the `build` command configuration works).
However, in that case CLI command logic (`generate-map.ts` under `cli` package) will have no access to `paths` argument. `paths` are required for `watch` logic. So we divide the config definition this way.

packages/create-sitecore-jss/src/templates/nextjs/package.json
```diff
@@ -58,11 +58,12 @@
     "typescript": "~5.8.3"
   },
   "scripts": {
-    "build": "cross-env NODE_ENV=production npm-run-all --serial sitecore-tools:build next:build",
+    "build": "cross-env NODE_ENV=production npm-run-all --serial sitecore-tools:generate-map sitecore-tools:build next:build",
```
stasmaxymov, line 61: What about calling 'generate-map' as part of 'build' process? We can have `runOnBuild`  or smth similar option on `generateComponentMap`, and when it is true - run generation on build. 

I just think we now have a very limiting concept of build commands with need to directly specify all commands.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -11,4 +11,10 @@ export default defineCliConfig({
       }),
     ],
   },
+  generateComponentMap: {
+    generator: generateMap,
```
stasmaxymov: I think providing function instances as config parameters doesn't look nice. I suggest to not provide generator in the cli.config and infer it elsewhere

packages/core/src/tools/templating/components.ts
```diff
@@ -29,10 +35,12 @@ export interface PackageDefinition {
  *  moduleName: 'ComponentName'
  * }
  * @param {string} path path to search
+ * @param {string[]} [exclude] paths and patterns to exclude from final result
  */
-export function getComponentList(path: string): ComponentFile[] {
+export function getComponentList(path: string, exclude?: string[]): ComponentFile[] {
```
stasmaxymov: It doesn't look like this function supports glob patterns (e.g. `./components/**/.tsx`)

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -0,0 +1,104 @@
...
+
+/**
+ * Arguments for the generateMap function.
+ * @typedef GenerateMapArgs
+ * @property {string[]} paths - Array of component paths to include in component map.
+ * @property {string} [destination='src/.sitecore'] - Destination folder path for the generated map.
+ * @property {PackageDefinition[]} [packages] - Optional array of package definitions to include in the map.
+ * @property {string[]} [exclude] - Optional array of paths to exclude from the map.
+ */
+export type GenerateMapArgs = {
```
stasmaxymov: AFAIR, there was a requirement to provide customization on component map template.

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -11,4 +11,10 @@ export default defineCliConfig({
       }),
     ],
   },
+  generateComponentMap: {
```
stasmaxymov: I would name this section "componentMap" to be more concise.

stasmaxymov (changes requested):

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
art-alexeyenko: I know and I agree 🥲 I describe the "why" in here:
https://github.com/Sitecore/content-sdk/pull/124#discussion_r2142762966

with the goal of supporting `watch` mode, and also framework-specific component-map-gen implementations, and also giving the users ability to configure generation process, this is the best compromise I could find. It still allows for customization, while providing a way to pass params into CLI logic.
But maybe I'm overthinking the implementation and there's an easier way to make it pretty that I did not see. What would you suggest?

packages/create-sitecore-jss/src/templates/nextjs/package.json
art-alexeyenko: I'm using this current approach for two main reasons:
- I think keeping build and generate-map CLI commands decoupled is better.
- If we remove the generate-map CLI command and only rung generateMap() as build command, we face challenges with watch mode. The conditions for its activation will not be as visible
- If we use both build and generate-map CLI commands, there would be duplicate configuration for the generation as part of `build` call and `generate-map` call - also not ideal


And in more general context, our larger goal with Content SDK is flexibility. This current approach allows it more easily, generate-map can still be disabled on build and component-map.ts can be managed manually for more advanced usecases.

packages/nextjs/src/tools/generate-map.ts
art-alexeyenko: right. added

packages/core/src/tools/templating/components.ts
art-alexeyenko: reworked it, the function is now much better, and more functional

packages/create-sitecore-jss/src/templates/nextjs/package.json
stasmaxymov: The main goals for Content SDK are simplicity and clarity, plugins in JSS V1 were much more flexible but not clear and simple. Here we have sort of fragmented experience, as you simply can't immediately tell for content sdk is doing on `build` step. Our main idea of introducing "build" command is to separate concerns and have designated place to control all operations.  

I do understand that current cli.config concepts are limiting, but that just means we should reconsider them. Example of such limiting concept is that build can only call listed commands, which can only be instances of functions.
And we can address that by expanding concept of 'sitecore-tools build` command

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
stasmaxymov: you can implicitly provide generator function instance in the definceCliConfig function as it is imported from framework specific module

packages/core/src/tools/templating/components.ts
stasmaxymov: looks good

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -11,4 +11,9 @@ export default defineCliConfig({
       }),
     ],
   },
+  componentMap: {
```
yavorsk, line 14: the property here is not required and can be removed, correct? bc the default value is set in the framework specific define-cli-config.

yavorsk (changes requested):
Looks Good! 
Just one comment from me and please add all affected packages in PR's title

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
art-alexeyenko: This was a temp thing, described here:
https://github.com/Sitecore/content-sdk/pull/124#discussion_r2142751077
now it's all done according to better practice and generator is set in package

packages/create-sitecore-jss/src/templates/nextjs/package.json
stasmaxymov: After a short call we agreed that it requires substantial rework of CLI to achieve desired behavior. Considering tight schedule we can take this as separate work

packages/create-sitecore-jss/src/templates/nextjs/sitecore.cli.config.ts
art-alexeyenko: yep, you're right, but I'm leaving it here for cli.config section to be more descriptive. Kind of a "show, don't tell" approach to configuration.

yavorsk: yea i see, but when we were introducing the cli config one of the discussed things was that it should be as clean as possible. thats why we moved the default scaffolding configuration to the framework specific define config.

yavorsk (approved):
approved in advance from me.

PR #125, opened 2025-06-13
[core][cli] Introduce `auth switch <tenantId>` command

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
- Adds `auth switch <tenantId>`
- The command verifies the current and the future tenants, performs the switch if verification successful.

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/cli/src/scripts/auth/switch.ts
```diff
@@ -0,0 +1,59 @@
...
+export type SwitchArgs = Pick<TenantInfo, 'tenantId'>;
+
+export const switchTenant: CommandModule<object, SwitchArgs> = {
+  command: 'switch <tenantId>',
+  describe: 'Switch into another tenant that you have logged into previously',
+  builder: (yargs: Argv<object>): Argv<SwitchArgs> =>
+    yargs.option('tenantId', {
+      type: 'string',
+      demandOption: true,
+      describe: 'Tenant ID to switch into.',
```
addy-pathania: Shouldn't `tenantId` be a required argument?
You can use the `requiresArg: true,` to make it required

packages/cli/src/scripts/auth/switch.ts
```diff
@@ -0,0 +1,59 @@
...
+    }),
+  handler: async (argv: SwitchArgs) => {
+    const tenantId = argv.tenantId;
+    if (!tenantId) {
+      console.error('Please provide tenant ID to switch into.');
+      return;
+    }
```
addy-pathania: If you use the `requiresArg` option above you would not need to make this check and yargs will do automatic error handling.

packages/cli/src/scripts/auth/switch.ts
```diff
@@ -0,0 +1,59 @@
...
+    const newTenantInfo = await readTenantInfo(tenantId);
+    if (!newTenantInfo) {
+      console.error(`Tenant info for ID '${tenantId}' not found in local storage.`);
+      console.error(
+        'Please ensure you have logged into the tenant by running the auth login command'
+      );
+      return;
+    }
+    setActiveTenant(tenantId);
+    const tenantContext = validateAndRenewAuthIfExpired();
```
addy-pathania, line 46: I believe it would always return null at line 49 because `validateAndRenewAuthIfExpired` attempts to read the auth info JSON file, which only exists for the currently active tenant. Since auth data is maintained per active tenant, it won’t be available for the tenant you're trying to switch into. If we are able to find a way to not read the auth info file in `validateAndRenewAuthIfExpired` then we are golden. Maybe pass a flag.

packages/cli/src/scripts/auth/switch.ts
```diff
@@ -0,0 +1,59 @@
...
+      return;
+    }
+
+    const newTenantInfo = await readTenantInfo(tenantId);
+    if (!newTenantInfo) {
+      console.error(`Tenant info for ID '${tenantId}' not found in local storage.`);
+      console.error(
+        'Please ensure you have logged into the tenant by running the auth login command'
+      );
+      return;
```
addy-pathania, line 43: process.exit should be more apt here?

addy-pathania (changes requested):
Looks good see comments.

packages/cli/src/scripts/auth/switch.ts
art-alexeyenko: The rest of the commands just do a 'return' and that stops executing the command. Should be enough, right?

addy-pathania: yeah should be.

packages/cli/src/scripts/auth/switch.ts
art-alexeyenko: I opted for an alternative approach. Instead of checking tenantInfo, I'm looking for tenantAuthInfo. This should ensure new tenant can be logged into.

addy-pathania (approved):
Looks good! 👍🏼

PR #129, opened 2025-06-17
[nextjs] Improve device detection and caching logic for Personalize middleware

addy-pathania:
- [x] Changelog updated

## Description / Motivation
This update improves isPrefetch logic to avoid skipping personalization for real mobile navigations.
- Refines the `isMobile` check to more accurately detect real user devices by considering known platform indicators (like iPhone, Mac, Linux, Windows) in addition to the `sec-ch-ua-mobile` header. This helps ensure that personalization is consistently applied across devices where standard mobile headers may not be set.
- We also introduce the `Cache-Control: no-store, no-cache, must-revalidate` header to non-prefetch responses, ensuring all requests bypass intermediary caches and always fetch fresh, personalized content from the server.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/nextjs/src/middleware/personalize-middleware.ts
+++ b/packages/nextjs/src/middleware/personalize-middleware.ts
@@ -131,7 +131,7 @@ export class PersonalizeMiddleware extends MiddlewareBase {
         // Disable preflight caching to force revalidation on client-side navigation (personalization WILL be influenced).
         // Note the reason we don't move this any earlier in the middleware is that we would then be sacrificing performance for non-personalized pages.
         res.headers.set('x-middleware-cache', 'no-cache');
-        res.headers.set('Cache-Control', 'no-cache, no-store, must-revalidate');
+        res.headers.set('Cache-Control', 'no-store, must-revalidate');
         return res;
       }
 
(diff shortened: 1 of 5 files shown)


packages/nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -131,6 +131,7 @@ export class PersonalizeMiddleware extends MiddlewareBase {
         // Disable preflight caching to force revalidation on client-side navigation (personalization WILL be influenced).
         // Note the reason we don't move this any earlier in the middleware is that we would then be sacrificing performance for non-personalized pages.
         res.headers.set('x-middleware-cache', 'no-cache');
+        res.headers.set('Cache-Control', 'no-cache, no-store, must-revalidate');
```
art-alexeyenko: I don't think this was present in JSS. Should it be moved over? And perhaps we should let users handle this - Cache-Control will be set for the whole request, regardless if personalization is present or not.

addy-pathania: So this was added to mitigate the issue of prefetch request being cached after next upgrade.
If we remove this we see the cached prefetch request which doesn't render personalization.

art-alexeyenko (approved):

PR #137, opened 2025-06-23
[core][cli] Enable available debug logs on build

MenKNas:
- [ ] Changelog updated

## Description / Motivation
This PR is introducing a change for debug logs not appearing during build execution. In more detail:

- Enabled DEBUG environment variables from `.env` files in CLI
- Added debug logging to multisite (site generation process)
- Debug logs now appear during build as expected (multisite-related logs since they are the only ones displayed on build time)

## Testing Details
- [ ] Unit Test Added

## Types of changes


packages/core/src/tools/generateSites.ts
```diff
@@ -40,6 +41,14 @@ export const generateSites = ({
     let sites: SiteInfo[] = [];
     const sitesFilePath = path.resolve(destinationPath ?? DEFAULT_SITES_DIST_PATH);
 
+    debug.multisite('Starting site generation process');
```
art-alexeyenko: A few points I want to add here:
- `site generation process` can be confusing. You can use `site information generation`, `fetching and saving site info` etc
- IMO there's too much unneeded info here. Users can find out whether context id and the rest of API configuration are present in sitecore.config already
- It might also be better to follow a similar approach we use in middleware. If multisite is disabled - log that operation is being skipped. If enabled - inform that operation started execution. Could be more informative this way

packages/cli/src/bin/sitecore-tools.ts
```diff
@@ -30,5 +30,13 @@ resolve('@sitecore-content-sdk/cli', { basedir: process.cwd() }, (error, project
   // in that case loading environment variables will be handled by the actual cli command
   processEnv(process.cwd());
 
+  // Re-enable debug logging after environment variables are loaded
+  // This ensures that DEBUG environment variables from .env files take effect
+  if (process.env.DEBUG) {
+    console.log(`Debug enabled: ${process.env.DEBUG}`);
```
art-alexeyenko: Could you make this more infomative please? I.e. "Debug enabled for scopes: [....]"

art-alexeyenko (changes requested):

packages/core/src/tools/generateSites.ts
MenKNas: Changed the rest. Regarding point number 2, that's the only useful piece of information I could find (apart from multisite being enabled or disabled). Is there something else that is probably useful to output in this case?

art-alexeyenko: @MenKNas let's not log context id and other stuff, but simply indicate whether site info generation was activated or not (and why - i.e. multisite is disabled in config)

art-alexeyenko: Also there's no need to add both debug and regular logs, I think

art-alexeyenko (approved):
Thank you! Look good and ready to roll

PR #131, opened 2025-06-18
[all] Remove jss references across content sdk

MenKNas:
- [x] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)
- [x] Changelog updated
- [ ] Upgrade guide entry added

## Description / Motivation
This PR removes all `jss` references across the entire code base.

_*does not include renames for things that are external dependencies like env vars, API endpoints,etc still reference jss_ 

## Testing Details

- [ ] Unit Test Added
- [x] Manual Test/Other (Please elaborate)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

PR #139, opened 2025-06-23
[core][nextjs] Remove SXA components from default template and allow for component-map with empty app

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
- Removes default SXA styles and components from `nextjs` template
- Going forward SXA components and styles will only be maintained as part of `xmcloud-foundation` repositories
- Adjusts component-map generation logic to ensure smooth operation with "empty" app
- Moves all the auxillary components (without corresponding Sitecore renderings) into separate folder

NOTE: should be merged after https://github.com/Sitecore/content-sdk/pull/131

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other (app builds, launches)

## Types of changes


packages/create-sitecore-jss/src/templates/nextjs/scripts/install-pre-push-hook.ts
art-alexeyenko, line 1: I believe this and other scripts files are no longer used, but they can be returned if that's not the case.

yavorsk (approved):
Looks good to me. The sample becomes leaner and leaner

PR #144, opened 2025-06-27
[core][nextjs] Add opt-out codegen option in sitecore config

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Implements the consent procedure for code generation via Sitecore config.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/core/src/config/define-config.ts
+++ b/packages/core/src/config/define-config.ts
@@ -53,7 +53,7 @@ export const getFallbackConfig = (): SitecoreConfig => ({
       timeout: 60,
     },
   },
-  codeGenerationOptOut: false,
+  disableCodeGeneration: false,
 });
 
 /**
--- a/packages/core/src/config/define-config.ts
+++ b/packages/core/src/config/define-config.ts
@@ -53,6 +53,7 @@ export const getFallbackConfig = (): SitecoreConfig => ({
       timeout: 60,
     },
   },
+  codeGenerationOptOut: false,
 });
 
 /**
--- a/packages/nextjs/src/tools/codegen/extract-files.ts
+++ b/packages/nextjs/src/tools/codegen/extract-files.ts
@@ -3,7 +3,6 @@ import {
   ExtractedFileType,
   resolveComponentImportFiles,
   sendCode,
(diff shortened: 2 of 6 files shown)


packages/core/src/config/define-config.ts
```diff
@@ -53,6 +53,7 @@ export const getFallbackConfig = (): SitecoreConfig => ({
       timeout: 60,
     },
   },
+  codeGenerationOptOut: false,
```
addy-pathania: doesn't matter but should we name this something like `disableCodeGeneration` or `skipCodeGeneration`?

Also do we need to add this as fallback value to nextjs defineConfig?

addy-pathania (commented):
Looks good, see one comment

packages/core/src/config/define-config.ts
art-alexeyenko: Not sure, I went for this one cause of the name of the task (and kind of cause of its nature). Let me check the naming with Product on Monday

illiakovalenko: Yeah, I would recommend to find something different, this name doesn't follow our naming convention

packages/nextjs/src/tools/codegen/extract-files.ts
```diff
@@ -36,8 +35,8 @@ export const extractFiles = (args: ExtractFilesConfig) => {
       debug.common('Skipping code extraction, not in deploy context');
       return;
     }
-    if (!validateConsent()) {
-      console.log(chalk.yellow('Skipping code extraction, consent not given'));
+    if (args.scConfig.codeGenerationOptOut) {
+      debug.common('Skipping code extraction, code generation has been opted out of');
```
illiakovalenko: This message might be udpated as well, together with property name, please, revisit

packages/core/src/config/define-config.ts
art-alexeyenko: @addy-pathania  @illiakovalenko renamed the option now

illiakovalenko (approved):
👍

PR #157, opened 2025-07-09
[nextjs][core] Add import-map generation for future rendering of generated components in Design Library

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Design Library needs to render and preview generated components seamlessly. Since components may import things from all over the place, we need an import map to dynamically add dependent values, classes etc when dynamically rendering generated components.

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other - map generated with no typescript errors

## Types of changes


packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+};
+
+/**
+ * Entry point function for generating import-map. Parses provided paths and outputs the modules and imports from those files into .sitecore/import-map.ts
+ * @param {WriteImportMapArgs} args include/exclude paths settings to be processed for import-map, and the Sitecore configuration
+ * @param {SitecoreConfig} scConfig Sitecore configuration from sitecore.config.ts
+ */
+export const writeImportMap = (args: WriteImportMapArgs, scConfig: SitecoreConfig) => {
+  return async () => {
+    if (scConfig.disableCodeGeneration) {
```
stasmaxymov, line 264: I'm not sure about controlling code-generation from application config, why don't have it on cli config?

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+      debug.common('Skipping import map generation. Code generation functionality is disabled.');
+      return;
+    }
+    if (!xmCloudDeploy()) {
```
stasmaxymov: Import map generation doesn't require XMC deploy context - it doesn't use any  specific stuff from it. Besides we want this command to be executed on regular build - because it generates new source code - users should be aware where it comes from. pretty much the same as component map

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+
+/**
+ * Gets a Map object with import modules and their respective exports present throughout the paths specified
+ * @param {string} paths paths to files to be processed for import-map
+ * @returns {Map<string, ImportModule>} collection of keys and values, where keys refer to modules being processed and values are collections of exports for each module
+ */
+export const getImportMap = (paths: string[]) => {
+  // make preparations for handling ts/js files
+  const appPath = process.cwd();
+  const tsConfig = ts.readConfigFile(path.resolve(appPath, 'tsconfig.json'), ts.sys.readFile);
```
stasmaxymov: what if this is JS only application?

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+
+/**
+ * Gets an import string and outputs import info with aliases, if present
+ * @param {string} importString import definition string
+ * @returns {Set<string,string>[]} object(s) with definition name (alias or import) and actual import
+ */
+export const parseImportString = (
+  importString: string
+): ImportWithAlias | ImportWithAlias[] | null => {
+  const namedImport = /\{.*\}/g;
```
stasmaxymov: Why don't we use AST to parse import clause? I believe it will be more robust than multiple regex

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+  const aliasImport = /^\s*([a-zA-Z0-9]+|\*) as (.+)\s*$/g;
+  const regularImport = /^\s*[a-zA-Z0-9]+\s*$/g;
+
+  if (namedImport.test(importString)) {
```
stasmaxymov: Each import can hold named, default and namespace import clauses at the same time. I'm not sure exclusive if is correct there

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+    return importString
+      .replace(/[import]*\s*\{/, '')
+      .replace(/\}/, '')
+      .split(',')
+      .reduce<ImportWithAlias[]>((acc, value) => {
+        const maybeAlias = aliasImport.exec(value);
+        if (maybeAlias) {
+          acc.push({
+            importName: maybeAlias[1].trim(),
+            alias: maybeAlias[2].trim(),
```
stasmaxymov: Why do we need aliases when parsing imports? Aliases don't change export/module identity so if one component has `import  { RichText } from '...'`
and another has `import {RichText as JssRichText} from '...'`
Those should result in the same entry on import map

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+          codeFileFullPath,
+          cliCompilerOptions,
+          tsHost
+        );
+        // get import path and check if its import target exists
+        const resolvedFile = resolvedModule?.resolvedModule?.resolvedFileName;
+        if (resolvedFile && imports) {
+          // if import path points to a file in local app - process import path to the file (i.e. ./myComponent)
+          // if it points to node_modules or a file in monorepo - parse import path as dependency module name (i.e. React)
+          const localModuleName =
```
stasmaxymov: Building a relative path for local imports might have unwanted behavior, because the same relative path is used on import map array. So a generated component will have an relative imports which work only from `.sitecore` folder perspective.

Besides, does this code account for aliases on tsconfig? I would expect that we need to use either full path from project root or apply relevant path alias from tsconfig when resolve module location

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+            resolvedFile.indexOf('node_modules') > -1 || resolvedFile.endsWith('.d.ts')
+              ? moduleName
+              : resolveLocalModulePath(resolvedFile, appPath);
+          // Set module import info in the map. If module import exists - add entries to existing entry
+          // Otherwise, add new entry
+          if (!importMapRecord.has(localModuleName)) {
+            importMapRecord.set(localModuleName, {
```
stasmaxymov: I see default and named exports. what about namespace?

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+              namedExports: new Map(),
+              defaultExports: new Map(),
+            });
+          }
+          // named imports in import statements have many values, default imports do not
+          if (Array.isArray(imports)) {
+            imports.forEach((value) => {
+              const importKey = value.alias || value.importName;
```
stasmaxymov: alias shouldn't be an import key - we need to always keep the real name that exists under module

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,399 @@
...
+  const importMapArray = Array.from(importMap);
+
+  // get import map entries after
+  const finalImportMap: ImportMapEntry[] = importMapArray.map(([modulePath, imports]) => {
+    const defaultExports = Array.from(imports.defaultExports).map(([exportAlias, exportName]) => {
+      const exportEntryName = exportAlias || exportName;
+      return {
+        originalName: exportName,
+        exportAlias: exportEntryName,
+        importMapValue: getImportValueAlias(exportEntryName, modulePath),
```
stasmaxymov: looks like we always apply alias for import variable name, but I think it will create unreadable import map, we should create alias only if it clashes with other name

stasmaxymov (changes requested):

packages/create-content-sdk-app/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -12,6 +12,11 @@ export default defineCliConfig({
       extractFiles({
         scConfig: config,
       }),
+      writeImportMap({
+        paths: ['src/components'],
+      },
+      config
```
illiakovalenko: Can we continue to follow our naming convention and set "scConfig" as a part of the first argument?

stasmaxymov:
One more thing, I see changelog is not updated. We probably should do that

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: For Content SDK we don't support js apps, I believe? Anyway, we do need fallback tsconfig values, true.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: namespace exports are considered `default` - they are marked as `wildcard` imports

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: My montivation for it was this. 
- If we pass 'disabled: true' into writeImportMap within CLI config, it's no different from just removing the call from there - so makes little sense
- But we also don't want the call to be lost from CLI config
- and finally, a conceptual difference. CLI is more for things that are "utils". Main config, to me, is for the global functionalities of the app, and codegen definitely fits this.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: Interesting. Disabled this part then.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: Addressed

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: After playing with AST, it's not that much robust unfortunately.
In any case, I streamlined the function a little with the use of AST, but some regex remains.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: function reworked, see above comment

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: Wrong assumption on my part on how aliases would be parsed by codegen renderings. Addressed.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: >Besides, does this code account for aliases on tsconfig?

Good call. Absolute paths are now being used - and if tsconfig is present, aliases will be accounted for.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: I added an index to keep track of all value names - please take a look

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: Could you please provide example where AST wasn't robust? I used babel for my PoC and relied on AST to parse import - it worked wonderfully

stasmaxymov: E.g. I have the following scenarios I doubt this code properly handles compound imports:
default +named imports: `import React, { ComponentType, useState as useBetterState } from 'react';`
this method will likely ignore a default import map

default + namespace imports (`import defaultExport, * as name from "module-name";`).

stasmaxymov: Sorry, I reviewed old changes once again 🤦. Let me check the new version

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,413 @@
...
+): ImportName[] | ImportName | null => {
+  const importClause = importNode.importClause?.getText();
+  const aliasImport = /^([a-zA-Z0-9]+) as .+$/g;
+  const namespaceImport = /^\* as (.+)$/g;
+
+  if (!importClause) {
+    console.warn('Cannot parse import string from: %s', importNode.getText());
+    return null;
+  }
+  if (/^\{.*\}$/.test(importClause)) {
```
stasmaxymov: It seems this code still doesn't work well with compound import declaration.
default import + namespace import
default import + named imports

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,413 @@
...
+    // import { a, b, c } from ..
+    // could use getChildren().map() but https://github.com/microsoft/TypeScript/issues/62112
```
stasmaxymov: Not useful link, as it explained there that it is not a bug

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,413 @@
...
+
+/**
+ * Gets an import string and outputs import info with aliases, if present
+ * @param {ts.ImportDeclaration} importNode import definition node
+ * @returns {ImportName[] | ImportName | null} object(s) with definition name
+ */
+export const getImportedValues = (
+  importNode: ts.ImportDeclaration
+): ImportName[] | ImportName | null => {
+  const importClause = importNode.importClause?.getText();
```
stasmaxymov: Still don't see why we can't rely just on AST, like you did partially for named imports. e.g.
if(importClause.name) -> has default import
if(importClause.namedBinding?.kind==Kind.NamespaceImport) -> has namespace import
if(isArray(importClause.namedBinding) -> has named imports

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,413 @@
...
+  return importValuesIndex.has(importName) && importValuesIndex.get(importName) !== moduleName
+    ? getImportValueAlias(importName, moduleName)
+    : importName;
+};
+
+// return alias-like name for an import value/variable name
+// this helps alleviate duplicate import names in import-map.ts
+const getImportValueAlias = (importValue: string, moduleName: string) => {
+  const prefix = crypto.hash('sha1', moduleName);
+  return `${prefix}_${importValue}`;
```
stasmaxymov: why prefix and not suffix?

packages/create-content-sdk-app/src/templates/nextjs/sitecore.cli.config.ts
```diff
@@ -1,17 +1,22 @@
...
+        scConfig,
       }),
       extractFiles({
-        scConfig: config,
+        scConfig,
       }),
+      writeImportMap({
+        paths: ['src/components'],
+      },
+      scConfig
```
stasmaxymov: If we keep scConfig - lets have passed as object property, so we can make it optional in future and provide it in centralized manner.

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: I see your point. I just don't like that our SC config and SC CLI config are so interconnected. So now it is hard to understand what options should go to CLI config and what should go to App config.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: no specific reason, I can make it a suffix

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,413 @@
+import * as ts from 'typescript';
+import path from 'path';
+import fs from 'fs';
+import { debug } from '@sitecore-content-sdk/core';
+import { getComponentList } from '@sitecore-content-sdk/core/tools';
+import { SitecoreConfig } from '@sitecore-content-sdk/core/config';
+import crypto from 'crypto';
+import { defaultImportEntries } from './default-import-map';
```
stasmaxymov: This way of integrating default map creates dependency between a generated map and default one. So user can't access a default map unless they like our heuristics for generated map. 
Integration point could be moved to later point, essentially to DesignLibrary.tsx component.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: in part, it's the discrepancy in getChildren() <-> forEachChild() logic in namedBindings. In other part, it is the depth we need to go to, to get correct import names (parsing aliases in particular, the import name node contains `x as y` always.
I feel the current approach with regex hits the good balance between readable and reliable.

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: Just realized this pattern was used in code extraction. I think it is a standard for now :). And later we can optimize it if needed

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: Alternatively you we can import and concat a default import map in generated file, to not statically insert a content.
`export default combineImportEntries(defaultImportEntries, importEntries)` 
where combineImportEntries - dedupes identical entries

This way we poeple who don't use code generation do not import these maps. And we have a more clear option for people to use default map when they don't want our generation mechanism

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,413 @@
...
+        // import path is extracted
+        const moduleName = childNode.moduleSpecifier.getText().replace(/['"]/g, '');
+        const resolvedModule = ts.nodeModuleNameResolver(
+          moduleName,
+          codeFileFullPath,
+          cliCompilerOptions,
+          tsHost
+        );
+        // get import path and check if its import target exists
+        const resolvedImportPath = resolvedModule?.resolvedModule?.resolvedFileName;
```
stasmaxymov, line 202: I just need your formal confirmation that this path resolution works correctly with tsconfig aliases and doesn't create issues with difference base for relative paths :)

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: we are providing just javascript packages with additional type sugar. We can't rule out usage of plain Javascript

art-alexeyenko: Just tested the logic with plain JS file - and it already works as expected. Latest commit adds the relevant unit test

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: It's more of an explanation of why one approach is used but not the other - still can be removed though

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: > I just need your formal confirmation that this path resolution works correctly with tsconfig aliases and doesn't create issues with difference base for relative paths :)

on it, boss. Let me add some unit tests to make sure.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: Addressed all the comments for that function. The current variant is the best it ever been - let's keep the discussion on it to this thread.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: I was convinced we combine the two in package - but this makes sense. Adjusted the thing and we now combine them in generated file.

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: Done, all good

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,421 @@
...
+    ].join('\n')
+  )
+  .join(',\n')}
+];
+
+export default combineImportEntries(defaultImportEntries, importMap);
+`;
+};
+
+export const combineImportEntries = (
```
stasmaxymov, line 389: Here you have just two arguments, but this function might be useful when we have multiple import maps, e.g, customer created separate map for each subproject.
Let's change incoming arguments to ...args: Array<ImportEntry[]> and allow to combine many maps

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: One more thing :)
You can use type predicates to check for node type and narrow down its type (isNamedImports, isNamespaceImport)
this way you can eliminate all string replacements and regex an easy traverse subnodes in case of named imports:
```
 if(ts.isNamedImports(node)){
    node.elements.forEach((element) => {
      const importName = element.propertyName?.text || element.name.text;
    }
  } 
```

packages/nextjs/src/tools/codegen/import-map.ts
stasmaxymov: but it seems in previous implementation you serialized it into generated import map, or did I understood it wrongly?

stasmaxymov (approved):
Overall changes look good. There are 2 comments left, but feel free to adress them in separate repo, they don't affect functionality

PR #161, opened 2025-07-22
[core] Duplicate dictionary requests in editing, preview, and design library modes

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Removed redundant dictionary fetches previously executed within `client.getPreview` and `client.getDesignLibraryData`.
The dictionary request is considered optional and should be initiated explicitly by consumers when needed.

## Testing Details

- [x] Unit Test Added

## Types of changes

  - This update introduces breaking changes in `EditingService`. However, we could consider it acceptable for the 1.1.0 release, as customers typically do not customize editing-related services, and the current implementation has a performance-impacting bug that this change resolves. Also dictionary is fetched by default via separate request so there are no changes for customers

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -12,6 +12,14 @@ Our versioning strategy is as follows:
 
 ## Unreleased
 
+### 🐛 Bug Fixes
+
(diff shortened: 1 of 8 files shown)


yavorsk (approved):
Looks great!

PR #163, opened 2025-07-25
[Design Library] Dynamic component rendering

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
This PR introduces support for dynamically rendering components generated by AI within the Design Library.

* Adds the ability to render a component based on AI-generated code.
* No breaking changes: If importMap is not provided, the code generation will not be enabled.

#### Error Handling

The following error scenarios are handled gracefully:

* Initialization Error - Occurs if an exception is thrown during function construction.
* Rendering Error - Occurs if an exception is thrown during component rendering.

In both cases, the error state can be cleared and retried when a new version of the generated code is received.

#### New "codegen" submodule

We have to introduce a separate submodule for "codegen" related functionality, since it utilizes "new Function" that is not allowed to be used in Edge Runtime. "editing" submodule is referenced by our middlewares and "sitecore-client"

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other - Tested using two application where one of them is iframe and the other one is parent

## Types of changes


packages/core/src/editing/codegen/preview.ts
```diff
@@ -0,0 +1,319 @@
...
+      const importNames = componentDepencencies.map((entry) => entry.name);
+      const importInstances = componentDepencencies.map((entry) => entry.value);
+
+      // get css file string from message and attach it to DOM
+      const style = document.createElement('style');
+      style.innerHTML = message.styles.content;
+      document.head.appendChild(style);
+
+      const exports: { Component: unknown } = { Component: null };
+
```
faiqamin, line 283: @illiakovalenko, a question. Do we need to add code validation here? Will there be any risks of malicious code injections?

illiakovalenko: @faiqamin We expect to get generated code from trusted source, I believe it should be done on backend side, since they transpile the code and it's generated by AI. On our side if the error is thrown by the generated code (initialization or render phase) we catch and render a proper message

faiqamin: Thanks. LGTM.

faiqamin (approved):

faiqamin (approved):

packages/core/src/editing/codegen/preview.ts
```diff
@@ -0,0 +1,319 @@
...
+        }
+
+        return;
+      }
+
+      console.debug('Component Library: message received', eventArgs);
+
+      const { message } = eventArgs;
+
+      const componentDepencencies = buildComponentDependencies(message.imports, importMap);
```
stasmaxymov: What if incoming message contains an import dependency that is not present in the import map? I suppose currently code will fail with prop of undefined exception or similarly. 
What if we consider it as error or warning here, to provide better context to user and DL app?

packages/core/src/editing/codegen/preview.ts
```diff
@@ -0,0 +1,319 @@
...
+
+      const importNames = componentDepencencies.map((entry) => entry.name);
+      const importInstances = componentDepencencies.map((entry) => entry.value);
+
+      // get css file string from message and attach it to DOM
+      const style = document.createElement('style');
```
stasmaxymov, line 277: we probably should keep track of style elements we insert. In case DL app will send multiple events with generated component we might end-up with many style elements inserted simultaneously

stasmaxymov (commented):
Changes look good. See couple of comments we might want  to address

packages/core/src/editing/codegen/preview.ts
illiakovalenko: Agree, done

stasmaxymov (approved):

PR #168, opened 2025-07-28
[nextjs] Add default import-map OOB

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
An intermediate solution to allow import map be imported iinto Layout.tsx in https://github.com/Sitecore/content-sdk/pull/163 without build failing.
Plus, fixes the syntax issue with missing commas in import-map

## Testing Details
- [ ] Unit Test Added

## Types of changes


packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -315,7 +315,7 @@ export const nextJsMapTemplate = (indexedImportMap: Map<string, ModuleExports>)
           .map(
             (namedExport) => `      { name: '${namedExport.name}', value: ${namedExport.value} }`
           )
-          .join(',\n')
+          .join(',\n') + ','
```
faiqamin, line 318:
```suggestion
          .filter(Boolean)
          .join(',\n') + ','
```
Should we add in `.filter(Boolean)` to remove empty lines when `outputExportEntries()` returns empty strings?

art-alexeyenko: `outputExportEntries` should not return empty strings, really. I will look for a solution in upcoming PR - but this is a good suggestion too.

faiqamin (approved):

PR #173, opened 2025-07-31
[cli] Code extraction extends XM Cloud Rendering Host build by several minutes

illiakovalenko:
- [x] Changelog updated

### Summary
This PR improves the behavior of the `sitecore-tools build` command by explicitly terminating the Node.js process after all build commands are executed. This prevents the process from hanging due to lingering background tasks.

### Motivation
The current implementation of `sitecore-tools build` allows any command to be executed during the build phase. However, we cannot guarantee that these commands will terminate cleanly - some may leave hanging processes, especially when tools like the TypeScript compiler (tscompiler) are initialized (e.g., during code extraction: [source](https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/tools/codegen/utils.ts#L74)).

Unfortunately, the compiler provides no public API to dispose or clean up such internal processes.

### Solution
To address this, we now explicitly call process.exit(0) after all build commands complete. This ensures a clean and predictable termination of the build process, avoiding CI/CD deadlocks and local development issues caused by lingering processes.

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -27,6 +27,7 @@ Our versioning strategy is as follows:
   - Updated `EditingService.fetchEditingData`:
     - Removed `siteName` parameter.
     - No longer requests and returns dictionary data.
+* `[cli]` Code extraction extends XM Cloud Rendering Host build by several minutes ([#173](https://github.com/Sitecore/content-sdk/pull/173))
 
 ## 1.0.0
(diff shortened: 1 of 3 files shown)


faiqamin (approved):
LGTM

PR #174, opened 2025-08-02
[core] Fix redirect regex processing to prevent over-escaping of question marks in regex patterns

faiqamin:
- [x] Changelog updated

## Description / Motivation

- Enhance `escapeNonSpecialQuestionMarks` to handle regex patterns correctly by avoiding over-escaping and improving language prefix removal logic. Maintain existing behavior for non-regex patterns with updated escaping function.

**Problem Summary**
- The redirect middleware was incorrectly processing regex patterns that contained special characters like `?` (quantifier) and anchors `(^, $)`. This caused redirect patterns like `^/meinezurich/?$` and `^/meinezurichtest/$` to fail matching their intended URLs.

**Root Cause**
The `escapeNonSpecialQuestionMarks` function was being applied to all patterns, including regex patterns that already contained special regex characters. This caused:
- Over-escaping: The `?` character in `^/meinezurich/?$` was being escaped to` \?`, creating a malformed regex that looked for a literal `\?` character instead of an optional `?`
- Pattern corruption: The resulting regex `/^\/meinezurich\/\?[/]?$/i` would not match `/meinezurich` or `/meinezurich/`
- Inconsistent behavior: Some patterns worked while others failed, depending on whether they contained special regex characters

**Solution**

Modified the `escapeNonSpecialQuestionMarks` function in `packages/core/src/utils/utils.ts` to:
- Detect regex patterns early: Check if the input contains `^` or `$` before any processing
- Skip over-escaping for regex patterns: Return regex patterns unchanged to preserve their special characters
- Preserve regex syntax: Don't escape `?` characters that are part of quantifiers `(?, *?, +?)` or negative lookaheads `((?!...))`
- Maintain backward compatibility: Non-regex patterns continue to use the existing logic with proper escaping

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -261,18 +261,35 @@ export class RedirectsMiddleware extends MiddlewareBase {
           }
 
           // process regex rules
-          // Modify the redirect pattern to ignore the language prefix in the path
-          // And escapes non-special "?" characters in a string or regex.
-          redirect.pattern = escapeNonSpecialQuestionMarks(
-            redirect.pattern.replace(new RegExp(`^[^]?/${language}/`, 'gi'), '')
-          );
+          // Check if the original pattern is already a regex pattern (starts with ^ and/or ends with $)
```
illiakovalenko: Looking at the problem I can see that "?" was escaped and handled in [escapeNonSpecialQuestionMarks](https://github.com/Sitecore/content-sdk/blob/bugfix/JSS-6811-redirect-with-the-question-mark-does-not-work/packages/core/src/utils/utils.ts#L204) method, does it mean that this function should be modified instead? And we need to add some extra checks there? I would propose to have a look, since the current implementation looks more complex

illiakovalenko (changes requested):

faiqamin:
fixing the test

illiakovalenko (approved):
👍

PR #175, opened 2025-08-04
[react][Codegen] Lazy load import map instead of direct import

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Optimize client bundle size in a post import-map world. Lazy loading it will reduce the main chunk size:
from
image
to
image
And move some of the lazy loaded content into separate chunk.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
+++ b/packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
@@ -6,7 +6,6 @@ import Head from 'next/head';
 import { Placeholder, Field, DesignLibrary, Page } from '@sitecore-content-sdk/nextjs';
 import Scripts from 'src/Scripts';
 import SitecoreStyles from 'src/components/content-sdk/SitecoreStyles';
-import importMap from '.sitecore/import-map';
 
 interface LayoutProps {
   page: Page;
@@ -22,6 +21,7 @@ const Layout = ({ page }: LayoutProps): JSX.Element => {
   const { route } = layout.sitecore;
   const fields = route?.fields as RouteFields;
   const mainClassPageEditing = mode.isEditing ? 'editing-mode' : 'prod-mode';
+  const importMapDynamic = () => import('.sitecore/import-map');
 
   return (
     <>
@@ -35,7 +35,7 @@ const Layout = ({ page }: LayoutProps): JSX.Element => {
       {/* root placeholder for the app, which we add components to using route data */}
       <div className={mainClassPageEditing}>
         {mode.isDesignLibrary ? (
-          <DesignLibrary importMap={importMap} />
+          <DesignLibrary importMapImport={importMapDynamic} />
         ) : (
           <>
             <header>
--- a/packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
+++ b/packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
@@ -35,7 +35,7 @@ const Layout = ({ page }: LayoutProps): JSX.Element => {
       {/* root placeholder for the app, which we add components to using route data */}
       <div className={mainClassPageEditing}>
         {mode.isDesignLibrary ? (
-          <DesignLibrary importMapImport={importMapDynamic} />
+          <DesignLibrary loadImportMap={importMapDynamic} />
         ) : (
           <>
             <header>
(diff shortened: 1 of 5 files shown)


packages/react/src/components/DesignLibrary.tsx
```diff
@@ -164,46 +168,73 @@ export const VariantGeneration = (props: VariantGenerationProps) => {
...
-        return;
+    let cancelled = false;
+    // since import map is loaded lazily, we only need to add preview event handler once the import map is loaded
+    // unsubscribe function for useEffect cleanup will also be returned once importMap promise has been resolved or rejected
+    let unsubscribe: (() => void) | undefined = undefined;
+
+    const init = async () => {
+      let importMap: codegen.ImportEntry[] = undefined;
+      try {
+        importMap = (await props.importMapImport()).default;
```
art-alexeyenko: @stasmaxymov could you clarify here: if import map is not resolved, should an error be thrown or and empty importMap passed into event handler?

packages/react/src/components/DesignLibrary.tsx
```diff
@@ -145,7 +149,7 @@ type VariantGenerationProps = {
   /**
    * The import map to be used in variant generation mode.
    */
-  importMap?: codegen.ImportEntry[];
+  importMapImport?: () => Promise<ImportMapImport>;
```
illiakovalenko: What do you think about renaming it to `loadImportMap`? That way, the function name clearly reflects its purpose and it's a common pattern to use "load" for dynamic/async components, files

packages/react/src/components/DesignLibrary.tsx
illiakovalenko: I would throw an error event since the import map resolver was passed intentionally, but something is happened

packages/react/src/components/DesignLibrary.tsx
```diff
@@ -164,46 +168,73 @@ export const VariantGeneration = (props: VariantGenerationProps) => {
...
+
+    init();
+    // return function that calls unsubsubribe - if the component was mounted correctly
+    return () => {
+      cancelled = true;
+      if (unsubscribe) {
+        unsubscribe();
+      }
+    };
+  }, [props.importMapImport]);
```
illiakovalenko: One optimization-related note: I see you added an importMapImport to the dependency array in useEffect, but we expect to execute `useEffect` only once, did you consider that?

illiakovalenko (changes requested):

packages/react/src/components/DesignLibrary.tsx
art-alexeyenko: true enough - and the dynamic import will not change much, so I removed the dependency

packages/react/src/components/DesignLibrary.tsx
art-alexeyenko: Yeah, that makes sense

packages/react/src/components/DesignLibrary.tsx
art-alexeyenko: Instead of throwing an error I'm indicating an error state through setInitError, but the details will be logged too.

packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
```diff
@@ -22,6 +21,7 @@ const Layout = ({ page }: LayoutProps): JSX.Element => {
   const { route } = layout.sitecore;
   const fields = route?.fields as RouteFields;
   const mainClassPageEditing = mode.isEditing ? 'editing-mode' : 'prod-mode';
+  const importMapDynamic = () => import('.sitecore/import-map');
```
illiakovalenko, line 24:
```suggestion
  const loadImportMap = () => import('.sitecore/import-map');
```

packages/react/src/components/DesignLibrary.test.tsx
```diff
@@ -378,31 +385,49 @@ describe('<DesignLibrary />', () => {
...
-          <DesignLibrary importMap={defaultImportMap} />
+          <DesignLibrary loadImportMap={errorImportMap} />
         </SitecoreProvider>,
         { container: document.body }
       );
 
-      fireEvent(null, () => {
-        throw new Error('Error rendering component');
+      await waitFor(() => {
+        callbackEvent('Error', null);
```
illiakovalenko: I believe that here we need to pass not an error but real component, since callback shouldn't throw an error in this test case
```suggestion
        callbackEvent(null, component);
```

illiakovalenko (approved):
👍

PR #176, opened 2025-08-05
[all]: Migrate eslint to version 9 and refactor accordingly

MenKNas:
- [x] Changelog updated

## Description / Motivation

This PR migrates ESLint from legacy configuration to modern flat config format, updating to the latest stable versions.

Core Changes:

- ESLint 9 Migration: Upgraded ESLint `8.57.0 → 9.0.0` and migrated from deprecated `.eslintrc` files to modern `eslint.config.mjs` flat config

- Package Dependencies: Updated all ESLint packages to the latest stable versions and added missing plugins (eslint-plugin-import, eslint-plugin-react-hooks)

- Configuration Consolidation: Replaced all package-level `.eslintrc` files with centralized flat config using file patterns for package-specific TypeScript and React rules

## Testing Details
- [ ] Unit Test Added
- [x] Manual Test/Other => Tested the correct linting of all packages with `yarn lint-packages` command after all the changes. Also tried some scenarios that triggered linting errors and were correctly caught by the linter.

## Types of changes


packages/create-content-sdk-app/package.json
```diff
@@ -38,6 +38,7 @@
     "minimist": "^1.2.8"
   },
   "devDependencies": {
+    "@stylistic/eslint-plugin-ts": "^2.10.0",
```
mg-aceik: If you're going to this effort, I don't understand why not use the latest v5 @stylistic/eslint-plugin? You're bringing in a whole lot of old dependencies here. 

Also I liked how un-opinionated it was about eslint config before, and now this is going to make it tricker to understand and remove your opinions when using this template.

(I meant to comment this on the `packages\create-content-sdk-app\src\templates\nextjs\package.json` but the same issue is there too).

MenKNas: @mg-aceik Thanks for the feedback. This is still a work in progress and not finalized yet.

MenKNas:
@Sitecore/jss-dev-reviewers 

Hello team,

This PR is now ready for a first round of review. Some things to notice and discuss further are the following:

- The approach I followed after migrating to `ESLint 9` is to have a centralized file `eslint.config.mjs` that follows the latest _**flat**_ syntax at the root of our project and is responsible for all our linting rules and logic. We can use this file to extend our linting rules for other packages without the obligation to maintain a dedicated file for each package. I find this approach more flexible and maintainable moving forward, but I am open to further discussion or other suggestions.

- In most cases I tried to keep our previous linting rules but also added some suggested/modern ones that will make our code easier to read. Of course this is personal preference so I would like to hear your opinion as well. We can adjust the rules according to our needs.

- Something weird happened after I resolved conflicts, and I can see all `.md` files added as changed because previously a final line for each file was removed. All files should be exactly the same as `dev` now so I think they can be ignored but it wouldn't hurt to have an extra quick look for those in case something was missed.

.github/workflows/publish.yml
```diff
@@ -65,4 +65,4 @@ jobs:
           echo //registry.npmjs.org/:_authToken=${NPM_TOKEN} > .npmrc
           npx lerna publish from-git --exact --canary --pre-dist-tag canary --preid canary --no-verify-access --yes
         env:
-          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
+          NPM_TOKEN: ${{ secrets.NPM_TOKEN }}
```
art-alexeyenko, line 68: probably an error here, removing empty line

tsconfig.json
```diff
@@ -17,5 +17,8 @@
     "noImplicitAny": true,
     "noUnusedLocals": true,
     "noUnusedParameters": true,
-  }
+  },
+  "exclude": [
+    "packages/create-content-sdk-app/src/templates/**/*"
```
art-alexeyenko, line 22: What does this change do?

eslint.config.mjs
```diff
@@ -0,0 +1,155 @@
...
+import jsdoc from 'eslint-plugin-jsdoc';
+import globals from 'globals';
+
+const __filename = fileURLToPath(import.meta.url);
+const __dirname = path.dirname(__filename);
+
+const compat = new FlatCompat({
+  baseDirectory: __dirname,
+  recommendedConfig: js.configs.recommended,
+  allConfig: js.configs.all,
```
art-alexeyenko: I'd like to confirm something. In original eslintrc we have the list of recommended configs being extended:
```
"extends": [
    "eslint:recommended",
    "plugin:jsdoc/recommended",
    "prettier",
    "prettier/@typescript-eslint"
  ],
```
How are these being replaced in flat config?

packages/create-content-sdk-app/src/templates/nextjs/package.json
```diff
@@ -42,16 +42,19 @@
...
+    "eslint": "^9.32.0",
+    "eslint-config-prettier": "^10.1.8",
+    "eslint-plugin-import": "2.32.0",
+    "eslint-plugin-jsdoc": "52.0.3",
+    "eslint-plugin-prettier": "^5.5.3",
+    "eslint-plugin-react": "7.37.5",
+    "eslint-plugin-react-hooks": "5.2.0",
     "npm-run-all2": "~8.0.1",
-    "prettier": "^3.5.3",
+    "prettier": "^2.8.8",
```
mg-aceik: Why the downgrade?

packages/create-content-sdk-app/src/templates/nextjs/.eslintrc
mg-aceik, line 1: Maybe I'm missing something but now the template won't include any eslint config?

packages/create-content-sdk-app/src/templates/nextjs/package.json
```diff
@@ -42,16 +42,19 @@
...
     "@types/react": "^19.1.3",
     "@types/react-dom": "^19.1.3",
-    "@typescript-eslint/eslint-plugin": "^8.32.0",
-    "@typescript-eslint/parser": "^8.32.0",
+    "@stylistic/eslint-plugin": "^5.2.2",
+    "@typescript-eslint/eslint-plugin": "8.39.0",
+    "@typescript-eslint/parser": "8.39.0",
     "cross-env": "~7.0.3",
-    "eslint": "^8.56.0",
-    "eslint-config-next": "^13.1.5",
```
mg-aceik, line 49: Have you considered replicating the same eslint packages and config you get from running `npx create-next-app@latest`?

.github/workflows/publish.yml
MenKNas: It was just removed before (not sure why) and re-added so it should work fine (looks exactly the same as `dev`)

tsconfig.json
MenKNas: This change helps us exclude `src/templates` and keeps the TS compiler and editor from type-checking our boilerplate template files which would produce spurious errors. ESLint already ignores them, but TS needs its own exclude.

eslint.config.mjs
MenKNas: The syntax has been updated and now accomodates all the previous configs plus new expanded rules

packages/create-content-sdk-app/src/templates/nextjs/.eslintrc
MenKNas: A new `eslint.config` file has also been introduced for templates, following the same latest flat config syntax.

packages/create-content-sdk-app/src/templates/nextjs/package.json
MenKNas: These have been updated.

eslint.config.mjs
```diff
@@ -0,0 +1,175 @@
+import { defineConfig, globalIgnores } from 'eslint/config';
+// import { FlatCompat } from '@eslint/eslintrc';
```
addy-pathania: is this comment intentional?

eslint.config.mjs
```diff
@@ -0,0 +1,175 @@
...
+      '@stylistic/space-before-function-paren': [
+        'error',
+        {
+          anonymous: 'never',
+          named: 'never',
+          asyncArrow: 'always',
+        },
+      ],
+      'react-hooks/rules-of-hooks': 'error',
+      'react-hooks/exhaustive-deps': 'warn',
```
addy-pathania, line 126: This looks repeated. Above you're setting this to `off`.

eslint.config.mjs
```diff
@@ -0,0 +1,175 @@
...
+        'error',
+        {
+          allowShortCircuit: true,
+          allowTernary: true,
+        },
+      ],
+      'spaced-comment': 'error',
+      curly: ['error', 'multi-line'],
+      'eol-last': ['error', 'always'],
+      'linebreak-style': ['error', 'windows'],
```
addy-pathania: should we remove this to avoid cross-OS issues?

addy-pathania (commented):
Great work. See some comments.

MenKNas:
> Great work. See some comments.

Thanks @addy-pathania, I addressed the comments and also expanded our `eslint.config` file slightly. Feel free to have a final look.

eslint.config.mjs
```diff
@@ -0,0 +1,172 @@
...
+      'no-new-wrappers': 'error',
+      'no-eval': 'error',
+      'dot-notation': 'error',
+      'no-trailing-spaces': 'error',
+      'brace-style': 'error',
+      quotes: ['error', 'single'],
+      radix: 'error',
+      'default-case': 'error',
+      eqeqeq: 'error',
+      'jsx-quotes': ['error', 'prefer-double'],
```
addy-pathania, line 113: In jsx we want double quotes?

eslint.config.mjs
```diff
@@ -0,0 +1,172 @@
...
+      '@typescript-eslint/typedef': 'error',
+      '@typescript-eslint/no-use-before-define': ['error', { functions: false, variables: false }],
+      '@stylistic/type-annotation-spacing': 'error',
+      '@stylistic/semi': 'error',
+
+      // Relaxed TS rules
+      '@typescript-eslint/no-require-imports': 'off',
+      '@typescript-eslint/no-explicit-any': 'off',
+      '@typescript-eslint/no-unused-expressions': 'off',
+      '@typescript-eslint/no-unused-vars': 'off',
```
addy-pathania, line 157: this is again listed twice or there is some reason?

eslint.config.mjs
```diff
@@ -0,0 +1,172 @@
...
+    rules: {
+      '@typescript-eslint/naming-convention': [
+        'error',
+        {
+          selector: 'typeLike',
+          format: ['PascalCase'],
+          custom: { regex: '^I[A-Z]', match: false },
+        },
+      ],
+      'no-undef': 'off',
```
addy-pathania, line 146: Since TypeScript has its own type checking, this rule might not be necessary. But its upto you to remove.

eslint.config.mjs
MenKNas: Yes, it's preferable because of attributes like `<Button label="OK" />` or similar. Prettier already handles this, but it doesn't hurt to enforce it extra.

eslint.config.mjs
MenKNas: I would keep this because core no-undef misfires in TypeScript, and the TS compiler already catches undefined identifiers. Leaving `'no-undef': 'off'` in your TS override prevents noisy false positives.

eslint.config.mjs
MenKNas: In fact there aren’t duplicate keys in the same rules block. What looks duplicated are the core JS rules vs the TS plugin rules.

In the base config we use core rules (e.g. no-unused-vars, no-unused-expressions) for JS.

In the TS override we explicitly turn off the TS equivalents (@typescript-eslint/no-*) to avoid overlap and keep the current behavior unchanged.

eslint.config.mjs
addy-pathania: make sense

addy-pathania (approved):
👍🏼

PR #178, opened 2025-08-11
[core] Restore back local connection (local.api creds)

MenKNas:
- [x] Changelog updated

## Description / Motivation
This PR restores the local API fallback when contextId is missing (when users want to use local connection first), ensuring correct Edge/local connection priority and updated error messages.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/core/src/config/define-config.ts
+++ b/packages/core/src/config/define-config.ts
@@ -14,8 +14,8 @@ export const getFallbackConfig = (): SitecoreConfig => ({
       edgeUrl: process.env.SITECORE_EDGE_URL || SITECORE_EDGE_URL_DEFAULT,
     },
     local: {
-      apiKey: process.env.NEXT_PUBLIC_SITECORE_API_KEY || '',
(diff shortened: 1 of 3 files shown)


packages/core/src/config/define-config.ts
```diff
@@ -14,8 +14,8 @@ export const getFallbackConfig = (): SitecoreConfig => ({
       edgeUrl: process.env.SITECORE_EDGE_URL || SITECORE_EDGE_URL_DEFAULT,
     },
     local: {
-      apiKey: '',
-      apiHost: '',
+      apiKey: process.env.NEXT_PUBLIC_SITECORE_API_KEY || '',
```
stasmaxymov: I'm not sure we should treat API Key as public thing (with **NEXT_PUBLIC** prefix). It should behave similarly to server context id. We also can read from either SITECORE_API_KEY or NEXT_PUBLIC_SITECORE_API_KEY to be flexible.

stasmaxymov (approved):

PR #187, opened 2025-08-19
[nextjs] Import map incorrectly handles duplicates when the same alias is used for both namespace and default exports

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
The import map generation fails when the same alias (e.g. React) is used for both a namespace import and a default import. This leads to a malformed import map output.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -158,8 +149,7 @@ export const getImportMap = (paths: string[]) => {
 
   // index to keep track of unique import values imported from different modules
   // helps avoid duplicate import names in the final import-map.ts file
-  // key = imported value name, value = imported module name
-  const importValuesIndex = new Map<string, string>();
+  const importsList: string[] = [];
```
art-alexeyenko, line 152: We can potentially use a Set here instead of an array. Depending on the size of an app, we'll have a lot of imports to go through, Set will be faster.

illiakovalenko: I changed it to use an array since the logic is already quite complex. Introducing additional maps or indexes would only make it harder to follow. This change should not affect performance, especially as it’s determined at build time - we’re not computing values, only setting and checking for occurrences

packages/nextjs/src/tools/codegen/test-data/import-map/multi-file-imports/D.tsx
```diff
@@ -0,0 +1,7 @@
+import * as React from '../fake-react';
+import { testClassInstance } from '../test-exports';
```
art-alexeyenko, line 2: to cover the full array of duplicating import name scenarios, let's modify this import to '../test-exports-2' and adjust the "should return map from multi-file imports with duplicate mixed exports" unit test accordingly.

art-alexeyenko (changes requested):
This is great, thank you Illia. I only have two comments to improve this

packages/nextjs/src/tools/codegen/import-map.ts
art-alexeyenko: It's not a big change, just use Set type instead of array for importsList and change related methods (maybe 6 lines changed in total?).
But regardless, it looks to be a negligible speed difference in our conditions (this is a short test procesing 10k entries)
image

art-alexeyenko (approved):

PR #190, opened 2025-08-22
[react] Send a `component:status` message during variant generation in Design Library

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Adds `component:status` events for `VariantGeneration` so the host can reliably track component lifecycle:

 - `READY` : fired once per uid when `VariantGeneration` has initialized, loaded the import map, and registered the preview handler.

 - `RENDERED` : fired once per successful preview render (per `renderKey`) after a valid component is set

**New:** we don't render custom error messages anymore, instead they should be sent to Design Library

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/react/src/components/DesignLibrary.test.tsx
+++ b/packages/react/src/components/DesignLibrary.test.tsx
@@ -6,7 +6,7 @@ import sinon from 'sinon';
 import { expect } from 'chai';
 import { Page, PageMode } from '@sitecore-content-sdk/core/client';
 import { LayoutServiceData } from '@sitecore-content-sdk/core/layout';
-import { fireEvent, render, waitFor, act } from '@testing-library/react';
+import { fireEvent, render, waitFor } from '@testing-library/react';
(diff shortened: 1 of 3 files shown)


packages/react/src/components/DesignLibrary.test.tsx
```diff
@@ -6,7 +6,7 @@ import sinon from 'sinon';
 import { expect } from 'chai';
 import { Page, PageMode } from '@sitecore-content-sdk/core/client';
 import { LayoutServiceData } from '@sitecore-content-sdk/core/layout';
-import { fireEvent, render, waitFor } from '@testing-library/react';
+import { fireEvent, render, waitFor, act } from '@testing-library/react';
```
illiakovalenko:
```suggestion
import { fireEvent, render, waitFor } from '@testing-library/react';
```

illiakovalenko (approved):

PR #192, opened 2025-08-25
[core] Content styles fail to load due to incorrect contextId resolution

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Fallback to client context id when server context id is not available (e.g. in a browser)

## Testing Details

- [x] Unit Test Added

## Types of changes


sc-hatimfarhathussain:
Hi @illiakovalenko , just a question, isnt app will failed to build if there's no Server-side context id provided in the first place? Hence why would we need this fallback if server-side context id is not provided.

illiakovalenko:
> Hi @illiakovalenko , just a question, isnt app will failed to build if there's no Server-side context id provided in the first place? Hence why would we need this fallback if server-side context id is not provided.

@sc-hatimfarhathussain  It will not fail. `getHeadLinks` can be called on both server and client side. In case it's called in the browser (our case) we don't have access to server context id, so we will fallback to client context id

sc-hatimfarhathussain:
> > Hi @illiakovalenko , just a question, isnt app will failed to build if there's no Server-side context id provided in the first place? Hence why would we need this fallback if server-side context id is not provided.
> 
> @sc-hatimfarhathussain It will not fail. `getHeadLinks` can be called on both server and client side. In case it's called in the browser (our case) we don't have access to server context id, so we will fallback to client context id

Ah, now i understand, thks man for the explanation.

yavorsk (approved):
Looks good!

PR #198, opened 2025-09-03
[nextjs][core] App router API editing routes

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Adds editing support for nextjs app router.
- Added new function `getEnforcedCorsHeaders` that is a more agnostic implementation of `enforceCors` and aims to replaces it later
- Refactored editing-config-middleware and editing-render-middleware to allow for better code reusability between app and pages router. No breaking changes.
- Moved common funtionality into /editing/utils.ts
- Added app router handlers editing and config API
  - New editing render handler passes preview data via query strings
  - No longer relies on nextjs logic, uses custom Content SDK-specific header to identify preview mode

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/core/src/client/sitecore-client.ts
```diff
@@ -404,8 +404,11 @@ export class SitecoreClient implements BaseSitecoreClient {
       return null;
     }
     // If we're in Pages preview (editing) mode, prefetch the editing data
-    const { site, itemId, language, version, variantIds, layoutKind, mode } =
-      previewData as EditingPreviewData;
+    const { site, itemId, language, version, layoutKind, mode } = previewData as EditingPreviewData;
+
+    const variantIds = Array.isArray(previewData.variantIds)
+      ? previewData.variantIds
+      : previewData.variantIds.split(',');
```
art-alexeyenko, line 411: There's no preview data for nextjs draft mode (used in app router), so preview data is passed via query string. This required an adjustment to how variantIds are handled

packages/nextjs/src/client/sitecore-nextjs-client.ts
```diff
@@ -73,6 +76,22 @@ export class SitecoreNextjsClient extends SitecoreClient {
...
+  /**
+   * Get design library page details for Design Library mode of your app
+   * @param {PreviewData} designLibData preview data set in 'library' mode of the app
+   * @param {FetchOptions} [fetchOptions] Additional fetch fetch options to override GraphQL requests
+   * @returns {Page} preview page for Design Library
+   */
+  async getDesignLibraryData(
+    designLibData: PreviewData,
+    fetchOptions?: FetchOptions
+  ): Promise<Page> {
```
art-alexeyenko, line 88: This is needed to make typescript happy when preview data from search params is passed in app router sample. We employ similar approach for regular getPreview method

art-alexeyenko:
@stasmaxymov there's a large increase for nextjs package size. I think we should move app router specific logic to a separate package, later on.

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx
```diff
@@ -11,22 +12,24 @@ import Providers from 'src/Providers';
...
   // Set preview to false until preview mode is integrated
-  const preview = { enabled: false, data: {} };
+  // const preview = { enabled: false, data: {} };
 
   // Fetch the page data from Sitecore
   let page;
-  if (preview.enabled) {
-    if (isDesignLibraryPreviewData(preview.data)) {
-      page = await client.getDesignLibraryData(preview.data);
+if (isCSDKPreview(await headers())) {
```
Comment (stasmaxymov) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: why don't you use `draftMode`? Using `headers` likely force page into dynamic rendering effectively killing SSG

Comment (stasmaxymov) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: `CSDK` abbreviation looks clunky, I suggest to simplify it like `isPreview`

Comment (stasmaxymov) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: See some details here: https://nextjs.org/docs/app/getting-started/partial-prerendering#dynamic-rendering

packages/nextjs/src/editing/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+
+  return secret;
+};
+
+/**
+ * Parses query string and its parameters to required editing parameters
+ * @param {{ [key: string]: string | undefined }} query query string values
+ * @returns {EditingRenderQueryParams} editing parameters
+ */
+export const getEditingParams = (query: {
```
stasmaxymov: we re-form query params into Editing params, maybe `mapEditingParams` name would be more clear

packages/nextjs/src/editing/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+      };
+  return params;
+};
+
+/**
+ * Filters out Next.js preview cookies from a cookie string or array
+ * @param {string | string[] | null} cookies cookie header value
+ * @returns {string[] | null} filtered cookies
+ */
+export const getFilteredCookies = (cookies: string | string[] | null) => {
```
stasmaxymov: Method name doesn't say anything about Next or preview, making method purpose unclear, let's name it as `extractNextPreviewCookies`

packages/nextjs/src/editing/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+  );
+  return filteredCookies;
+};
+
+/**
+ * Gets the Next.js preview cookies to enable preview mode
+ * @param {string} site current site name
+ * @returns {string[]} list of cookies to set
+ */
+export const getNextPreviewCookies = (site: string) => {
```
stasmaxymov: those are not Next cookies, but our own ones

packages/nextjs/src/editing/editing-render-middleware.ts
```diff
@@ -200,124 +142,59 @@ export class EditingRenderMiddleware extends RenderMiddlewareBase {
...
-        // Cache the preview data for 3 seconds to ensure the page is rendered with the correct preview data not the cached one
-        {
-          maxAge: 3,
-        }
-      );
-    }
+    const previewDataParams = getEditingParams(query as { [key: string]: string });
+
+    // With the editing optimization work done, we can cut the dependency on nextjs preview mode
+    // @TODO in next major release: pass preview data via query string, variantIds will be a single string too
```
stasmaxymov: passing query string parameters in Pages router won't work for SSG projects. Next simply doesn't provide them to getStaticProps function.

CHANGELOG.md
```diff
@@ -14,62 +14,63 @@ Our versioning strategy is as follows:
...
+  - Base template ([#191](https://github.com/Sitecore/content-sdk/pull/191))
+  - Editing and preview support ([#198](https://github.com/Sitecore/content-sdk/pull/198))
+  - `[nextjs]` Support component-level data fetching in 404/500 pages ([#199](https://github.com/Sitecore/content-sdk/pull/199))
+- `[react]` Add `component:status` events for VariantGeneration ([#190](https://github.com/Sitecore/content-sdk/pull/190))
 
 ## 1.1.0
 
 ### 🎉 New Features & Improvements
 
-* `[nextjs]` Support component-level data fetching in 404/500 pages ([#199](https://github.com/Sitecore/content-sdk/pull/199))
```
illiakovalenko, line 26: This feature is not related to App Router, can you move it back, please? It was in a right place

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx
Comment (art-alexeyenko) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: draft mode cookies are not populated unless we do a .redirect() call, and so it doesn't work with our current optimized approach

Comment (stasmaxymov) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: Hmm, if I remember correctly after setting draftmode you can access a draft cookie using `await cookie()` next helper.
See this part in PoC: https://github.com/stasmaxymov/content-sdk/blob/ecef7ae364def03ce25e5d22a5caff65e405c55d/samples/nextjs/src/app/api/editing/render/route.ts#L53C3-L53C14

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx
```diff
@@ -11,22 +12,24 @@ import Providers from 'src/Providers';
...
+  searchParams: Promise<{ [key: string]: string | string[] | undefined }>;
 };
 
-export default async function Page({ params }: PageProps) {
+export default async function Page({ params, searchParams }: PageProps) {
   const { path } = await params;
 
   // Set preview to false until preview mode is integrated
-  const preview = { enabled: false, data: {} };
+  // const preview = { enabled: false, data: {} };
```
Comment (addy-pathania) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: is this comment intentional?

packages/nextjs/src/editing/editing-render-middleware.ts
```diff
@@ -135,17 +78,27 @@ export class EditingRenderMiddleware extends RenderMiddlewareBase {
       body,
     });
 
-    if (!enforceCors(req, res, EDITING_ALLOWED_ORIGINS)) {
+    const corsHeaders = getEnforcedCorsHeaders({
+      requestMethod: req.method,
+      headers: req.headers,
+      presetCorsHeader: headers['Access-Control-Allow-Origin'] as string,
```
addy-pathania, line 84: `presetCorsHeader` can be string[] too.
We should normalize it and pass it cleanly.

packages/nextjs/src/editing/utils.ts
```diff
@@ -0,0 +1,302 @@
...
+    : (req as NextApiRequest).headers.host;
+  // use https for requests with auth but also support unsecured http rendering hosts
+  return `${useHttps ? 'https' : 'http'}://${host}`;
+};
+
+/**
+ * Gets the Content-Security-Policy header value
+ * @returns Content-Security-Policy header value
+ */
+export const getSCPHeader = () => {
```
addy-pathania: i think we should rename this to getCSPHeader. Don't know why we named it SCP.

packages/nextjs/src/route-handler/editing-config-route-handler.ts
```diff
@@ -0,0 +1,143 @@
...
+            'Content-Type': 'application/json',
+          },
+        });
+      }
+
+      if (secret !== getEditingSecret()) {
+        debug.editing(
+          'invalid editing secret - sent "%s" expected "%s"',
+          secret,
+          getEditingSecret()
```
addy-pathania, line 70: any benefit from logging these 2? I think we should just log the message.

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -0,0 +1,210 @@
...
+    // init query string values
+    const query: EditingRenderQueryParams = {} as EditingRenderQueryParams;
+    req.nextUrl.searchParams.forEach((value, key) => {
+      query[key] = value;
+    });
+
+    const expectedCorsHeaders = getEnforcedCorsHeaders({
+      requestMethod: req.method,
+      headers: req.headers,
+      presetCorsHeader: req.headers?.get('Access-Control-Allow-Origin') as string,
```
addy-pathania, line 62: `Access-Control-Allow-Origin` is a response header, it won't be available on req.
We should directly pass undefined.

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -0,0 +1,210 @@
...
+        { status: 401 }
+      );
+    }
+
+    const responseHeaders: { [key: string]: string } = expectedCorsHeaders;
+
+    // Validate secret
+    const secret = query[QUERY_PARAM_EDITING_SECRET];
+    if (secret !== getEditingSecret()) {
+      debug.editing('invalid editing secret - sent "%s" expected "%s"', secret, getEditingSecret());
```
addy-pathania, line 115: again, any value in logging secret?

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -0,0 +1,210 @@
...
+      const html = await getEditingRequestHtml(
+        requestUrl,
+        propagatedQsParams,
+        propagatedHeaders,
+        convertedCookies,
+        dataFetcher
+      );
+
+      // disable draft mode after page is rendered
+      await draft.disable();
```
addy-pathania: If an exception is thrown draft mode disable will be skipped, maybe we should move it to the finally block.

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -0,0 +1,210 @@
...
+      debug.editing(
+        'invalid origin host - set allowed origins in JSS_ALLOWED_ORIGINS environment variable'
+      );
+      return new Response(
+        `<html><body>Requests from origin ${req.headers.get('origin')} not allowed</body></html>`,
+        { status: 401 }
+      );
+    }
+    debug.editing('preflight request');
+    return new Response(null, { status: 204, headers: expectedCorsHeaders });
```
addy-pathania, line 76:
```suggestion
    return new Response(null, { status: 204, headers: ...(expectedCorsHeaders ?? {}) });
```

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -0,0 +1,210 @@
...
+        { status: 400, headers: responseHeaders }
+      );
+    }
+
+    let rawCookies = headers.get('cookie') || '';
+
+    // Set Preview mode identifier cookie, if the page is rendered in Sitecore Preview mode
+    if (mode === LayoutServicePageState.Preview) {
+      const previewCookies = getNextPreviewCookies(query.sc_site);
+      rawCookies = rawCookies.concat(previewCookies.join('; '));
```
addy-pathania: Please check if this way of concatenating cookies is correct. This might cause duplication and corruption. Rather we should only append the cookies which we want the browser to store.

addy-pathania (commented):
Great work Artem, see some comments.

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx
Comment (art-alexeyenko) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: no, should be removed. good catch

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx
Comment (art-alexeyenko) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: In regards to this part
> cookieStore.set('__prerender_bypass', cookieStore.get('__prerender_bypass')

draft mode relies on `_prerender_bypass` cookie now, but it will not neccessarily do so in the future. As long as the draft mode functionality is obfuscated and internalized within nextjs, it is more reliable to rely on headers we manager ourselves IMO

packages/nextjs/src/editing/editing-render-middleware.ts
art-alexeyenko: Cookie headers can be, but `Access-Control-Allow-Origin` header is only a single string value as per spec https://fetch.spec.whatwg.org/#http-access-control-allow-origin

packages/nextjs/src/route-handler/editing-config-route-handler.ts
art-alexeyenko: It's done for parity with original config/editing middlewares

packages/nextjs/src/route-handler/editing-render-route-handler.ts
art-alexeyenko: It's done for parity with original config/editing middlewares

packages/nextjs/src/route-handler/editing-render-route-handler.ts
art-alexeyenko: They will never be null at this point of implementation as we do falsly check on line 56 above

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx
Comment (art-alexeyenko) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/app/[[...path]]/page.tsx]: reverted to draft mode

packages/nextjs/src/route-handler/editing-render-route-handler.ts
art-alexeyenko: I'd leave this for now - I need to clarify NextRequest behavior with vercel. NextRequest will have some response headers (like Set-Cookie), and considering response is not available in API routes, CORS headers should be somewhere in request too.

packages/nextjs/src/route-handler/editing-render-route-handler.ts
art-alexeyenko: logic has been reworked for nextjs's `cookies()`

addy-pathania (approved):
👍🏼

PR #201, opened 2025-09-05
[nextjs] Preserve default locale in external absolute urls

MenKNas:
- [x] Changelog updated

## Description / Motivation
Fixes an issue where the default locale was stripped from external absolute URLs in redirects.

Example:
Before: `/en-hk/ra` → `https://example.com/this-is-en`
After: `/en-hk/ra` → `https://example.com/en/this-is-en`

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -14,66 +14,63 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* Next.js App Router support:
-  * Robots.txt and Sitemap.xml support ([#197](https://github.com/Sitecore/content-sdk/pull/197))
-  * Base template ([#191](https://github.com/Sitecore/content-sdk/pull/191))
-* `[react]` Add `component:status` events for VariantGeneration ([#190](https://github.com/Sitecore/content-sdk/pull/190))
-
-### 🐛 Bug Fixes
-
-* `[nextjs]` Preserve default locale in external absolute urls ([#201](https://github.com/Sitecore/content-sdk/pull/201))
+- Next.js App Router support:
+  - Robots.txt and Sitemap.xml support ([#197](https://github.com/Sitecore/content-sdk/pull/197))
+  - Base template ([#191](https://github.com/Sitecore/content-sdk/pull/191))
+- `[react]` Add `component:status` events for VariantGeneration ([#190](https://github.com/Sitecore/content-sdk/pull/190))
 
 ## 1.1.0
 
 ### 🎉 New Features & Improvements
 
-* `[nextjs]` Support component-level data fetching in 404/500 pages ([#199](https://github.com/Sitecore/content-sdk/pull/199))
-* Migration from ESlint 8 -> ESLint 9 and introduction of the new Flat Config file ([#176](https://github.com/Sitecore/content-sdk/pull/176))
-* Code generation for Design Library enablers:
+- `[nextjs]` Support component-level data fetching in 404/500 pages ([#199](https://github.com/Sitecore/content-sdk/pull/199))
+- Migration from ESlint 8 -> ESLint 9 and introduction of the new Flat Config file ([#176](https://github.com/Sitecore/content-sdk/pull/176))
+- Code generation for Design Library enablers:
   - `[core]` `[nextjs]` Add import-map generation ([#157](https://github.com/Sitecore/content-sdk/pull/157))([#167](https://github.com/Sitecore/content-sdk/pull/167))([#170](https://github.com/Sitecore/content-sdk/pull/170))([#171](https://github.com/Sitecore/content-sdk/pull/171))([#175](https://github.com/Sitecore/content-sdk/pull/175))([#177](https://github.com/Sitecore/content-sdk/pull/177))([#187](https://github.com/Sitecore/content-sdk/pull/187))
     - New `writeImportMap()`, `combineImportEntries()` methods and `defaultImportEntries` export available from `@sitecore-content-sdk/nextjs/codegen`
   - Dynamic component rendering ([#163](https://github.com/Sitecore/content-sdk/pull/163))
   - Updated API endpoint to new Edge Platform format ([#162](https://github.com/Sitecore/content-sdk/pull/162))
   - Ensure editing state is enabled in Design Library mode ([#181](https://github.com/Sitecore/content-sdk/pull/181))
-* `[core]` Ensure displayName paths are properly UTF-8 encoded. ([#179](https://github.com/Sitecore/content-sdk/pull/179))
-* `[react]` Enhanced the Design Library cache buster format to hh-dd-mm-yyyy ([#188](https://github.com/Sitecore/content-sdk/pull/188))
-* `[nextjs]` Optimization for editing render middleware: issue an internal server request for fetching page data during editing instead of doing temporary redirect ([#195](https://github.com/Sitecore/content-sdk/pull/195)) ([#196](https://github.com/Sitecore/content-sdk/pull/196))
+- `[core]` Ensure displayName paths are properly UTF-8 encoded. ([#179](https://github.com/Sitecore/content-sdk/pull/179))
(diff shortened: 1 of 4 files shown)


packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -144,7 +144,7 @@ export class RedirectsMiddleware extends MiddlewareBase {
           const targetParts = existsRedirect.target.split('/');
           const urlFirstPart = targetParts[1];
 
-          if (this.locales.includes(urlFirstPart)) {
+          if ( !REGEXP_ABSOLUTE_URL.test(existsRedirect.target) && this.locales.includes(urlFirstPart) ) {
```
illiakovalenko:
```suggestion
          if (!REGEXP_ABSOLUTE_URL.test(existsRedirect.target) && this.locales.includes(urlFirstPart)) {
```

CHANGELOG.md
```diff
@@ -19,6 +19,10 @@ Our versioning strategy is as follows:
   * Base template ([#191](https://github.com/Sitecore/content-sdk/pull/191))
 * `[react]` Add `component:status` events for VariantGeneration ([#190](https://github.com/Sitecore/content-sdk/pull/190))
 
+### 🐛 Bug Fixes
+
+* `[nextjs]` Preserve default locale in external absolute urls ([#201](https://github.com/Sitecore/content-sdk/pull/201))
```
illiakovalenko: This fix should go under 1.1.0 release, so let's move it down below

packages/nextjs/src/middleware/redirects-middleware.test.ts
```diff
@@ -1713,6 +1713,51 @@ describe('RedirectsMiddleware', () => {
         expect(finalRes).to.deep.equal(res);
         expect(finalRes.status).to.equal(res.status);
       });
+
+      it('should not strip default locale from external absolute URLs', async () => {
```
illiakovalenko: I suppose that this is not only about "default" language, but about all the languages if you provide proper configuration to "sitecore.config.ts" -> "redirects.locales". Can you double check, please? By default "en" is set

packages/nextjs/src/middleware/redirects-middleware.ts
illiakovalenko: From my understanding in case you have an "absolute" url - you can't reach this "else" branch, since on [line 140](https://github.com/Sitecore/content-sdk/pull/201/files#diff-7f5b8504a3f63fc359d33ee438a9f96733b1b39a1a7d0cbe70fed44fffd14e02R140) there is a check for "absolute" format, so here you have a relative url in all the cases. Can you double check, please? May be there is a different problem

illiakovalenko (changes requested):

packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -137,14 +137,32 @@ export class RedirectsMiddleware extends MiddlewareBase {
...
+                301,
+                'Moved Permanently'
+              );
+            case REDIRECT_TYPE_302:
+              return this.createRedirectResponse(existsRedirect.target, res, 302, 'Found');
+            case REDIRECT_TYPE_SERVER_TRANSFER:
+              return this.rewrite(existsRedirect.target, req, res, true);
+            default:
+              return res;
+          }
```
illiakovalenko: Let’s avoid copy-pasting existing code, as we’re aiming to follow best practices. Instead, you can simply set
```
url = existsRedirect.target
```
So, in the end, the function will execute exactly the [same code](https://github.com/Sitecore/content-sdk/blob/dfb767f933edf2cfdee89df410df720d0efa0700/packages/nextjs/src/middleware/redirects-middleware.ts#L200-L212).
If this approach doesn’t work for some reason, you should introduce a new method that accepts url and executes this switch block

packages/nextjs/src/middleware/redirects-middleware.test.ts
```diff
@@ -1713,6 +1744,55 @@ describe('RedirectsMiddleware', () => {
...
+          req,
+          res
+        );
+
+        const urlVal =
+          typeof (finalRes as any).url === 'string'
+            ? (finalRes as any).url
+            : (finalRes as any).url?.href;
+
+        expect(urlVal).to.equal(externalUrl);
```
illiakovalenko:
```suggestion
        expect(finalRes.url).to.equal(externalUrl);
```

packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -137,14 +137,32 @@ export class RedirectsMiddleware extends MiddlewareBase {
...
         } else {
           const isUrl = isRegexOrUrl(existsRedirect.pattern) === 'url';
           const targetParts = existsRedirect.target.split('/');
           const urlFirstPart = targetParts[1];
 
-          if (this.locales.includes(urlFirstPart)) {
+          if (
+            !REGEXP_ABSOLUTE_URL.test(existsRedirect.target) &&
```
illiakovalenko: This change should be reverted, since it will never be executed

illiakovalenko (changes requested):

packages/nextjs/src/middleware/redirects-middleware.test.ts
```diff
@@ -548,7 +575,7 @@ describe('RedirectsMiddleware', () => {
         expect(siteResolver.getByHost).to.be.calledWith(hostname);
         // eslint-disable-next-line no-unused-expressions
         expect(fetchRedirects.called).to.be.true;
-        expect(finalRes).to.deep.equal(res);
```
illiakovalenko, line 551: Is there a specific reason for removing this check across all the tests? It was originally added to ensure that the mock response created before test execution was actually used in the middleware and returned as the final response

CHANGELOG.md
```diff
@@ -14,63 +14,64 @@ Our versioning strategy is as follows:
 
 ### 🎉 New Features & Improvements
 
-* Next.js App Router support:
-  * Robots.txt and Sitemap.xml support ([#197](https://github.com/Sitecore/content-sdk/pull/197))
-  * Base template ([#191](https://github.com/Sitecore/content-sdk/pull/191))
-* `[react]` Add `component:status` events for VariantGeneration ([#190](https://github.com/Sitecore/content-sdk/pull/190))
+- Next.js App Router support:
```
illiakovalenko: Please, revert all the unnecessary changes in CHANGELOG

packages/nextjs/src/middleware/redirects-middleware.test.ts
illiakovalenko: @MenKNas This one is not resolved

packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -342,6 +337,41 @@ export class RedirectsMiddleware extends MiddlewareBase {
...
+          isExternal
+        );
+      default:
+        return res;
+    }
+  }
+
   /**
    * Helper function to create a redirect response and remove the x-middleware-next header.
    * @param {NextURL} url The URL to redirect to.
```
illiakovalenko: The parameter's type should be updated
```suggestion
   * @param {NextURL | string} url The URL to redirect to.
```

illiakovalenko (approved):

packages/nextjs/src/middleware/redirects-middleware.test.ts
MenKNas: Restored back `expect(finalRes).to.deep.equal(res)` only when the middleware should return the same response (preview/disabled/no-redirect). For redirects/rewrites Next returns a new NextResponse, so object equality is fragile. We assert behavior (status + key headers like location/x-sc-rewrite) instead.

illiakovalenko (approved):
👍

PR #115, opened 2025-06-04
[core] [templates/nextjs] - Add metadata generation support for mono-repo setup

MiguelMinoldo:
This commit allows metadata generation to query Sitecore packages without filtering out workspaces, enabling support for mono-repo setups.

- [x] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)
- [ ] Changelog updated
- [x] Upgrade guide entry added

## Description / Motivation
This PR adds support for mono-repo setups by allowing metadata generation to query Sitecore packages without filtering out workspaces. 

This change is necessary to ensure that metadata can be generated correctly in environments where multiple workspaces are used, such as in a mono-repo setup. 

It addresses the issue of metadata generation not accounting for workspaces, which can lead to incomplete or incorrect metadata.

## Testing Details
- **Unit Test Added**: A new unit test has been added to verify that the `generateMetadata` function correctly handles the `allowWorkspaces` property when set to `true`.
- **Manual Test/Other**: The changes have been manually tested in a mono-repo setup to ensure that metadata is generated correctly without filtering out workspaces.
- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #203, opened 2025-09-08
[core] [templates/nextjs] - Add metadata generation support for mono-repo setup

art-alexeyenko:
This is a slight rework of community PR https://github.com/Sitecore/content-sdk/pull/115
This commit allows metadata generation to query Sitecore packages without filtering out workspaces, enabling support for mono-repo setups.

- [x] Changelog updated

## Description / Motivation
This PR adds support for mono-repo setups by allowing metadata generation to query Sitecore packages without filtering out workspaces. 

This change is necessary to ensure that metadata can be generated correctly in environments where multiple workspaces are used, such as in a mono-repo setup. 

It addresses the issue of metadata generation not accounting for workspaces, which can lead to incomplete or incorrect metadata.

## Testing Details
- **Unit Test Added**: A new unit test has been added to verify that the `generateMetadata` function correctly handles the `allowWorkspaces` property when set to `true`.
- **Manual Test/Other**: The changes have been manually tested in a mono-repo setup to ensure that metadata is generated correctly without filtering out workspaces.
- [x] Unit Test Added

## Types of changes


packages/core/src/editing/metadata.ts
```diff
@@ -15,13 +15,17 @@ const trackedScopes = [
...
  */
-export function getMetadata(): Metadata {
+export function getMetadata(allowWorkspaces: boolean = false): Metadata {
   const metadata: Metadata = { packages: {} };
 
   let queryResult: Package[] = [];
   try {
-    queryResult = JSON.parse(execSync('npm query [name*=@sitecore] --workspaces false').toString());
+    queryResult = JSON.parse(
+      execSync(`npm query [name*=@sitecore] --workspaces ${allowWorkspaces}`).toString()
```
addy-pathania, line 27: I think in npm, --workspaces is a flag (boolean). Passing --workspaces false can still be interpreted as “set the flag”. I think we should only include the flag when `allowWorkspaces === true`.

art-alexeyenko: Passing `--workspaces` false is how we do it right now in codebase - just explicit instruction to not use workspaces, for safety. I'd leave this as is.

addy-pathania (approved):

PR #206, opened 2025-09-16
[Chore] Cherrypick "[core] [templates/nextjs] - Add metadata generation support for mono-repo setup (#203)(#115)"

art-alexeyenko:
Cherrypick for community PR into upcoming release.


packages/core/src/tools/generateMetadata.ts
```diff
@@ -19,17 +19,22 @@ export type GenerateMetadataConfig = {
...
  * Generate application metadata
  * @param {GenerateMetadataConfig} config - Optional configuration for generating metadata.
- * If not provided, the default '.sitecore/metadata.json' will be used.
+ * If not provided, the default '.sitecore/metadata.json' will be used and allowWorkspaces will be set to false.
  * @returns {Promise<void>} A promise that resolves when the metadata generation is complete.
  */
 export const generateMetadata = (config?: GenerateMetadataConfig): (() => Promise<void>) => {
   return async () => {
-    const metadata: Metadata = getMetadata();
+    const metadata: Metadata = getMetadata(!!config?.allowWorkspaces);
```
MenKNas, line 37: Do we need to coerce this as a boolean since it's already typed as a boolean? Also, is there a chance that config might be undefined (hence the use of the optional chaining) ?

art-alexeyenko: that's exactly the usecase. `allowWorkspaces` and config itself might be undefined, so we coearce to get boolean always

MenKNas (approved):

art-alexeyenko:
pipeline does not complete, will re-create PR

PR #208, opened 2025-09-16
[Design Library] Unite capabilities of library-metadata and library-variant generation modes

addy-pathania:
- [x] Changelog updated

## Description / Motivation
- Unified behavior: `mode=library` and `mode=library-metadata` now optionally enable DL variant generation with `generation=variant`.

- Removed `DesignLibraryMode.VariantGeneration`. `isVariantGeneration` is honored only when `isDesignLibrary` is true and `generation=variant` query string is passed to the editing render endpoint.

Rendering flow:

 - Always render the real component first.

 - Post `component:status` `READY` on mount.

 - Non-generation: post `RENDERED` after first paint and after each `component:update` event.

 - DL Variant Generation on: 
   - lazily load `importMap` (via `loadImportMap`), 
   - send `component:generation:import-map` and `component:generation:component-props`,
   - listen for `component:generation:component-preview`, 
   - then switch to the generated component and continue handling `component:update`.
   - On setup errors, post `component:generation:component-preview-error`

Migration notes:

- To enable generation, pass `generation=variant` alongside `mode=library | library-metadata;` provide `loadImportMap`.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/react/src/components/DesignLibrary.tsx
```diff
@@ -146,143 +102,131 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
...
+    }
+
+    const unsubUpdate = addComponentUpdateHandler(rendering, (updated) => {
+      setPropsState({ fields: updated.fields, params: updated.params });
+      setRenderKey((k) => k + 1);
+    });
+
+    // useEffect will cleanup event handler on re-render
+    return () => unsubUpdate && unsubUpdate();
+  }, [isDesignLibrary, rendering?.uid]);
```
MenKNas: Let's add missing dependencies here. Especially, rendering is important to avoid memory leaks since the handler won't be cleaned up when it changes.

packages/react/src/components/DesignLibrary.tsx
```diff
@@ -146,143 +102,131 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
...
 
-  // eslint-disable-next-line react-hooks/rules-of-hooks
   useEffect(() => {
+    // Send a rendered event only as effect of a component update command
+    if (renderKey === 0) return;
+
+    postToDL(getDesignLibraryStatusEvent(DesignLibraryStatus.RENDERED, rendering.uid));
+  }, [renderKey, isDesignLibrary, rendering?.uid, postToDL]);
```
MenKNas: Same here, let's make sure dependencies are consistent.

CHANGELOG.md
```diff
@@ -21,6 +21,7 @@ Our versioning strategy is as follows:
   - Internationalization support ([#202](https://github.com/Sitecore/content-sdk/pull/202))
 * `[react]` Add `component:status` events for VariantGeneration ([#190](https://github.com/Sitecore/content-sdk/pull/190))
 * `[nextjs]` Support component-level data fetching in 404/500 pages ([#199](https://github.com/Sitecore/content-sdk/pull/199))
+* `[react]` `[core]` Unite capabilities of library | library-metadata with library-variant-generation modes. `isVariantGeneration` is honored only when `isDesignLibrary`(library | library-metadata) is true and `generation=variant` query string is passed to the editing render endpoint
```
MenKNas: PR number is missing here. Also, we could move this up the list to maintain chronological order

packages/react/src/components/DesignLibrary.tsx
```diff
@@ -146,143 +102,131 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
...
     };
-  }, []);
-
-  // eslint-disable-next-line react-hooks/rules-of-hooks
-  useEffect(() => {
-    // Send a rendered event only as effect of a component update command
-    if (renderKey === 0) return undefined;
-
-    const renderedEvent = getDesignLibraryStatusEvent(DesignLibraryStatus.RENDERED, rendering.uid);
+  }, [rendering?.uid]);
```
MenKNas: Again, let's make sure dependencies are consistent

packages/react/src/components/DesignLibrary.tsx
```diff
@@ -146,143 +102,131 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
...
+ * Renders the **real** Sitecore component for `library` / `library-metadata` modes and,
+ * when generation is enabled (`page.mode.designLibrary.isVariantGeneration === true`),
+ * wires the **variant generation** handshake so the parent (DL Studio) can send
+ * generated code to preview and iterate on.
+ * @param {DesignLibraryProps} props
+ * @param {() => Promise<{ default: import('../codegen').ImportEntry[] }>} [props.loadImportMap] Optional async loader that resolves to the import-map used to resolve the generated component’s imports. Required when `isVariantGeneration` is true.
+ * @returns {JSX.Element} The preview surface, or `null` when not in Design Library mode.
  */
-export const VariantGeneration = (props: VariantGenerationProps) => {
+export const DesignLibrary = ({ loadImportMap }: DesignLibraryProps): JSX.Element => {
```
MenKNas: Is there a need to type this as `JSX.Element` ?

packages/react/src/components/DesignLibrary.tsx
```diff
@@ -146,143 +102,131 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
...
-    if (renderKey === 0) return undefined;
-
-    const renderedEvent = getDesignLibraryStatusEvent(DesignLibraryStatus.RENDERED, rendering.uid);
+  }, [rendering?.uid]);
 
-    console.debug('Component Library: sending event', renderedEvent);
-
-    window.top?.postMessage(renderedEvent, '*');
-  }, [renderKey, rendering?.uid]);
+  if (!isDesignLibrary) return null;
```
yavorsk: wouldn't it be better to move this check at the beginning before all the useEffects for visibility?

packages/core/src/editing/models.ts
```diff
@@ -74,8 +74,10 @@ export enum DesignLibraryMode {
   Normal = 'library',
   /** Metadata mode */
   Metadata = 'library-metadata',
-  /** Variant generation mode */
-  VariantGeneration = 'library-variant-generation',
+}
+
+export enum DesignLibraryVariantGeneration {
```
yavorsk, line 82: maybe add ts-doc for consistency

yavorsk (approved):
Looks good to me,
few minor comments, approved in advance

packages/react/src/components/DesignLibrary.tsx
addy-pathania: We just want to check if the `uid` changed but it makes sense to add the whole rendering. These are only dependencies we need to for the effects to re-run. Same for the following comments.

CHANGELOG.md
addy-pathania: Ahh, good catch.
Updated.

packages/react/src/components/DesignLibrary.tsx
addy-pathania: Yeah there is no need, will let it infer.

packages/react/src/components/DesignLibrary.tsx
addy-pathania: Yes makes sense.

PR #210, opened 2025-09-18
[react][nextjs] App Router ServerPlaceholder implementation

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Main focus of this PR is ServerPlaceholder implementation:
- Handle both server and react client components seamlessly
- Prefer passing props via, well, props instead of react context.

Also features a heavy refactor of old Placeholder component for better readability and navigate-ability. No breaking changes should occur there.

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/react/src/components/Placeholder/ServerPlaceholder.tsx
```diff
@@ -0,0 +1,111 @@
...
+} from './PlaceholderCommon';
+import React from 'react';
+import { PlaceholderMetadata } from './PlaceholderMetadata';
+import { ComponentRendering } from '@sitecore-content-sdk/core/layout';
+
+export type ServerPlaceholderProps = PlaceholderProps & {
+  componentMap: ComponentMap;
+};
+
+export const ServerPlaceholder = (props: ServerPlaceholderProps) => {
```
addy-pathania: can we add some tsDoc?

packages/react/src/components/Placeholder/ServerPlaceholder.tsx
```diff
@@ -0,0 +1,111 @@
...
+  if (!props.componentMap) {
```
addy-pathania, line 21: do we need this check? since componentMap is required

packages/react/src/components/Placeholder/ServerPlaceholder.tsx
```diff
@@ -0,0 +1,111 @@
...
+      delete finalProps[prop];
+      return finalProps;
+    },
+    { ...props }
+  );
+
+  const placeholderRenderings = getPlaceholderRenderings(
+    props.rendering,
+    props.name,
+    props.page.mode.isEditing
```
addy-pathania: we should add optional chaining to avoid runtime errors.

packages/react/src/components/Placeholder/ServerPlaceholder.tsx
```diff
@@ -0,0 +1,111 @@
...
+      if (!isEmpty) {
+        const errorBoundaryKey = rendered.type + '-' + index;
+
+        rendered = wrapErrorBoundary(rendered, props, errorBoundaryKey, dynamic);
+      }
+
+      // if in edit mode then emit shallow chromes for hydration in Pages
+      if (props.page.mode.isEditing) {
+        return (
+          <PlaceholderMetadata key={finalPhProps.name} rendering={rendering}>
```
addy-pathania: is finalPhProps.name unique always?

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+import { ComponentType } from 'react';
+import { ServerPlaceholder } from '../components/Placeholder';
+import { ComponentRendering } from '@sitecore-content-sdk/core/layout';
+import { ComponentMap } from '../components/sharedTypes';
+import { Page } from '@sitecore-content-sdk/core/client';
+
+export type ComponentProps = {
+  rendering: ComponentRendering;
+  placeholders: Record<string, React.ReactNode>;
+  componentMap?: ComponentMap;
```
addy-pathania: If ServerPlaceholder requires a componentMap, shouldnt it be required here too?

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+};
+
+export type WrapperProps = {
+  page: Page;
+  componentMap?: ComponentMap;
```
addy-pathania: same

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+  rendering: ComponentRendering;
+};
+
+export const withServerPlaceholder = <T extends ComponentProps, W extends T & WrapperProps>(
+  Component: ComponentType<T>,
+  placeholders: string[]
+) => {
+  return (props: W) => {
+    const phProps: Record<string, unknown> = {};
```
addy-pathania, line 24: consider wrapping this with useMemo since it gets re-created on every render.

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+    for (const placeholder of placeholders) {
+      phProps[placeholder] = (
+        <ServerPlaceholder
+          name={placeholder}
+          rendering={props.rendering}
+          page={props.page}
+          componentMap={props.componentMap}
```
addy-pathania, line 31: if componentMap is optional here it would throw an error.

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+        />
+      );
+    }
+    const propsCopy: T = { ...props };
+    delete propsCopy.componentMap;
```
addy-pathania: just a suggestion but i think we should avoid this mutate-then-delete patterns. We can use destructuring instead:
const { componentMap: _omit, ...rest } = props;
const propsCopy: T = rest as T;

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+  componentMap?: ComponentMap;
+};
+
+export type WrapperProps = {
+  page: Page;
+  componentMap?: ComponentMap;
+  rendering: ComponentRendering;
+};
+
+export const withServerPlaceholder = <T extends ComponentProps, W extends T & WrapperProps>(
```
addy-pathania, line 19: i think we should consider adding displayName to the returned component for better DevTools developer experience
  const wrappedName = Component.displayName || Component.name || 'Component';
 ...
   Wrapped.displayName = `withServerPlaceholder(${wrappedName})`;
   return Wrapped;

addy-pathania (commented):
Great work Artem.
See comments and suggestions.
Also, please update the changelog.

stasmaxymov:
@art-alexeyenko I noticed PR doesn't have changes for this part of the ticket:
>Implement conditional exports for placeholders in package.json to simplify usage and transition, or provide a wrapper component that switches between server and client implementations based on context.

Were there any problems with it?

packages/react/src/enhancers/withServerPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+  Component: ComponentType<T>,
+  placeholders: string[]
```
stasmaxymov: Not sure `placeholders` prop is required. I have it in PoC implementation, but probably just coppied from original `withPlaceholder` implementation. Can we take list of placeholders from the component itself, and property `rendering: ComponentRendering;`?
Or even better have placeholders prop as kind of a filter, when you skip it - all rendered, and when provided - only mentioned are rendered.

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
```diff
@@ -0,0 +1,260 @@
...
+  rendered: React.ReactElement<{
+    [attr: string]: unknown;
+  }>,
+  placeholderProps: PlaceholderProps,
+  renderingKey: string,
+  isDynamic: boolean = false,
+  extraProps: { [attr: string]: unknown } = {}
+) => {
+  // assign type based on passed element - type='text/sitecore' should be ignored when renderEach Placeholder prop function is being used
+  const type = rendered.props.type === 'text/sitecore' ? rendered.props.type : '';
```
stasmaxymov: do we still need this `type`? looks like artifact of `chromes` age

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
```diff
@@ -0,0 +1,260 @@
...
+
+  const disableSuspense = placeholderProps.disableSuspense || false;
+
+  return (
+    <ErrorBoundary
+      data-testid="error-boundary"
+      key={renderingKey}
+      errorComponent={placeholderProps.errorComponent}
```
stasmaxymov: Could you test errorComponent and componentLoadingMessage with server components? Cause it might an error there. 
`ErrorBoundary` has to be a client component to capture errors.
`errorComponent` and `componentLoadingMessage` - are actually functions (React.FC).
So it seems you pass non-serializable prop to a client component.

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
```diff
@@ -0,0 +1,260 @@
...
+      componentLoadingMessage={placeholderProps.componentLoadingMessage}
+      type={type}
+      isDynamic={isDynamic}
+      disableSuspense={disableSuspense}
+      {...rendered.props}
```
stasmaxymov: these extended props are potential problems, as they can have non-serializable fields and ErrorBoundary is a client component

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
```diff
@@ -0,0 +1,260 @@
...
+/**
+ * Wraps rendered component(s) with an error boundary.
+ * @param {React.ReactElement} rendered React elements to be wrapped
+ * @param {PlaceholderProps} placeholderProps props of placeholder under which the component is being rendered
+ * @param {string} renderingKey unique key for the error boundary
+ * @param {boolean} isDynamic whether error boundary wraps any lazy components
+ * @param {object} [extraProps] extra non-rendered component props to be passed to the error boundary
+ * @returns {React.ReactElement} wrapped component
+ */
+export const wrapErrorBoundary = (
```
stasmaxymov: I don't see much value in this function to be honest you can simply use errorboundary component as JSX in placeholder components directly

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
```diff
@@ -0,0 +1,260 @@
...
+      ? ((component as ReactModule)[exportName] as ComponentType)
+      : (component as ReactModule).default ||
+        (component as ReactModule).Default ||
+        (component as ComponentType);
+
+  // all dynamic elements will have a separate render prop
+  return {
+    component: renderedComponent,
+    dynamic: !!(renderedComponent as LazyComponentType).render?.preload,
+    isClient: !!(component as ReactModule).isClient,
```
stasmaxymov: We agreed to change convention on module to `componentType`

packages/react/src/components/Placeholder/models.ts
```diff
@@ -0,0 +1,112 @@
...
+//
+export const nonSerializedProps = [
+  'renderEmpty',
+  'render',
+  'renderEach',
+  'errorComponent',
+  'componentLoadingMessage',
+  'modifyComponentProps',
+  'componentMap',
+  'pageContext',
```
stasmaxymov: there are no `pageContext` on placeholder props, it seems it was renamed to `page`

packages/react/src/components/Placeholder/Placeholder.tsx
```diff
@@ -0,0 +1,166 @@
...
+  getComponentForRendering,
+  getPlaceholderRenderings,
+  getRenderedComponentProps,
+  renderEmptyPlaceholder,
+  wrapErrorBoundary,
+} from './PlaceholderCommon';
+import { ComponentRendering } from '@sitecore-content-sdk/core/layout';
+import { PlaceholderMetadata } from './PlaceholderMetadata';
+
+export class PlaceholderComponent extends React.Component<PlaceholderProps> {
```
stasmaxymov, line 16: Considering you had to change this file, maybe it is a good time to make it functional component 😃

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
```diff
@@ -0,0 +1,260 @@
+﻿import React, { ComponentType } from 'react';
```
stasmaxymov, line 1: `PlaceholderCommon` filename was fine when this file contained a base React component, let's rename it to show that it contains mostly utilities

stasmaxymov (commented):
Nice work! 
See some suggestions and questions in comments

packages/react/src/components/Placeholder/Placeholder.tsx
art-alexeyenko: Mercy, please. Refactoring all the placeholder stuff was bad enough.
I'll try, but focus on other points first

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
art-alexeyenko: It just makes reading Placeholder/ServerPlaceholder logic easier, so no one needs to scroll past this large ErrorBoundary block.
I'm fine reverting it if you insist.

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
art-alexeyenko: It just makes reading Placeholder/ServerPlaceholder logic easier, so no one needs to scroll past this large ErrorBoundary block.
I'm fine reverting it if you insist.

packages/react/src/enhancers/withServerPlaceholder.tsx
art-alexeyenko: we'll reuse props between server and regular placeholders, so componentMap needs to be optional. I've adjusted componentMap to be madatory in HOC, but optional in other places

packages/react/src/components/Placeholder/ServerPlaceholder.tsx
art-alexeyenko: should be uid here, adjusted

packages/react/src/components/Placeholder/ServerPlaceholder.tsx
art-alexeyenko: made componentMap optional - we need to have it so for compatibilty between client/server placeholder

packages/react/src/enhancers/withServerPlaceholder.tsx
art-alexeyenko: useMemo can't be used with server components unfortunately:
https://github.com/vercel/next.js/discussions/68071#discussioncomment-10134807

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
art-alexeyenko: Adjusted the thing, and it seems to be working ok. Custom error component works OK too

packages/react/src/components/Placeholder/index.ts
```diff
@@ -0,0 +1,14 @@
...
+
+const AppPlaceholder = rsc ? ServerPlaceholder : Placeholder;
+
+// exporting actual implementation names for inner usage to avoid uncertainty
+export { ServerPlaceholder } from './ServerPlaceholder';
+export { Placeholder, PlaceholderComponent } from './Placeholder';
+export { PlaceholderMetadata } from './PlaceholderMetadata';
+export { PlaceholderProps } from './models';
+export * from './placeholder-utils';
+// exporting the server/client implementation for outer usage
```
art-alexeyenko: @stasmaxymov check the conditional placeholder export here and the root `index.ts`
I'm not confident enough to offer a conditional "cat in the bag" Placeholder for page router users yet - so I'd rather we intriduce it in next major release and use AppPlaceholder for app router for now.

stasmaxymov: conditional exports are on package.json, not index.ts, like
```
    {
      "name": "my-package",
      "exports": {
        ".": {
          "react-server": "./lib/server.js",
          "default": "./lib/client.js"
        }
      }
    }
```

and with this way Pages router users won't get react-server implementation

packages/react/src/components/ErrorBoundary.tsx
```diff
@@ -7,10 +7,11 @@ type ErrorComponentProps = {
   [prop: string]: unknown;
 };
 
+// @TODO: remove type in next major release
 export type ErrorBoundaryProps = {
   children: ReactNode;
   page: Page;
-  type: string;
+  type?: string;
```
stasmaxymov: is error boundary exported from core package, maybe we can just remove it?
If not, it would be great to mark it as obsolete with tsdoc

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
stasmaxymov: I don't see any changes to `errorComponent` and `componentLoadingMessage`. Besided it seems Errorboundary is not marked with "use client" and considering that nothing in import chain that leads to ErrorBoundary is marked with "use client", makes me think that ErrorBoundary is a server component, which is not possible as it is a class component.
We should look at it together to understand this mystery.

stasmaxymov: Just to be sure, is Layout.tsx still marked with use client on your test bench?

packages/react/src/components/Placeholder/placeholder-utils.tsx
```diff
@@ -0,0 +1,258 @@
...
+ * @returns {React.ReactElement} wrapped component
+ */
+export const wrapErrorBoundary = (
+  rendered: React.ReactElement<{ [attr: string]: unknown }>,
+  placeholderProps: PlaceholderProps,
+  renderingKey: string,
+  isDynamic: boolean = false
+) => {
+  const disableSuspense = placeholderProps.disableSuspense || false;
+  const serializableProps = nonSerializedProps.reduce(
```
stasmaxymov: This nonSerialized props hack is strange from this helper perspective for two reasons:
- it is applied regardless it is client or server context
- ErrorBoundary component doesn't expect arbitrary key value props and don't need all this stuff from placeholderProps

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
stasmaxymov: I want to convince you, not force 😃 
to me rendering a JSX is essentially calling a function and if util just call JSX just with a bit shorthand syntax - there is no benefit considering sacrificing clarity a bit. 
So if after all changes to errorboundary call we end-up with helper that just calls component - I suggest to get rid of it.

packages/react/src/components/Placeholder/placeholder-utils.tsx
art-alexeyenko: Realistically we will pass some modification of PlaceholderProps here, so we know the props names to be passed. It can be communicated better in code however, true that

packages/react/src/components/Placeholder/index.ts
art-alexeyenko: We need to preserve esm/cjs exports while using this approach and these are the aproaches I tried:
```
 "exports": {
    ".": {
      "default": {
        "import": "./dist/esm/index.js",
        "require": "./dist/cjs/index.js"
      },
      "react-server": {
        "import": "./dist/esm/server.js",
        "require": "./dist/cjs/server.js"
      },
      "types": "./types/index.d.ts"
    },
    "./package.json": "./package.json"
  },
```
and
```
  "exports": {
    ".": {
      "import": {
        "default" : "./dist/esm/index.js",
        "react-server": "./dist/esm/server.js"
      },
      "require": {
        "default" : "./dist/esm/index.js",
        "react-server": "./dist/esm/server.js"
      },
      "types": "./types/index.d.ts"
    },
    "./package.json": "./package.json"
  },
```

Neither works. Even when I use `rsc-env` directly in app, it always returns rsc=false. I've reverted to using ServerPlaceholder directly for now.

packages/react/src/components/Placeholder/placeholder-utils.tsx
art-alexeyenko: Adjusted function to be more generic

packages/react/src/components/Placeholder/PlaceholderCommon.tsx
art-alexeyenko: Okay I'm convinced - moving the logic out of the function, there are enough server/client differences to warrant it now :)

packages/react/src/components/Placeholder/index.ts
art-alexeyenko: Placeholder is dynamically decided at app level now.

packages/react/src/components/Placeholder/placeholder-utils.tsx
stasmaxymov: It seems we are not on the same page regarding props passed to error boundary. 
Error boundary component expects only this props:
```
export type ErrorBoundaryProps = {
  children: ReactNode;
  page: Page;
  type: string;
  isDynamic?: boolean;
  errorComponent?: React.ComponentClass<ErrorComponentProps> | React.FC<ErrorComponentProps>;
  rendering?: ComponentRendering;
  componentLoadingMessage?: string;
  disableSuspense?: boolean;
};
```
There is no need to dump `passThroughProps` into error component. Pass only props expected by component

stasmaxymov: and thus no need for all this pick serializable prop helpers

packages/react/src/components/EditingReadyComponent.tsx
```diff
@@ -0,0 +1,14 @@
+'use client';
+import React, { useEffect } from 'react';
+
+/**
+ * Component notifies the Pages when all of react page content is loaded.
+ * This ensures additional Pages markup can be applied at the correct time.
+ */
+export const EditingReadyComponent = () => {
```
stasmaxymov: We already have a way to notify Page's canvas that they need to reconsider chromes, helper called `resetChromes` can we use that?

stasmaxymov: And if we use that, I'd prefer to not use this component in Pages router

packages/react/src/components/Placeholder/ServerPlaceholder.tsx
```diff
@@ -0,0 +1,128 @@
...
+
+      if (!isEmpty) {
+        const errorBoundaryKey = rendered.type + '-' + index;
+        const disableSuspense = props.disableSuspense || false;
+        const passThroughProps = pickSerializableProps(rendered.props);
+        rendered = (
+          <ErrorBoundary
+            data-testid="error-boundary"
+            key={errorBoundaryKey}
+            errorComponent={props.errorComponent}
```
stasmaxymov, line 67: I remember we talked that passing errorComponent directly creates problems in server context and we either need to address typing to accept `ReactNode` or try to have a server wrapper (HOC) for error boundary that will render error component into JSX and pass to Error boundary

art-alexeyenko: It works okay if:
- component is imported from a separate file
- the imported file is marked 'use client`

I believe it's something we'd need to document

packages/react/src/components/EditingReadyComponent.tsx
art-alexeyenko: using resetChromes now in a useEffect

stasmaxymov: I see, but this app router specific workaround still affects pages router template

stasmaxymov (approved):

PR #221, opened 2025-09-22
[nextjs] Normalize import-map keys from absolute paths to relative paths.

addy-pathania:
- [x] Changelog updated

## Description / Motivation
This PR fixes a path mismatch where the import-map stored absolute build paths (e.g. /workspace/tmp/builds/...) that don’t exist causing “Missing module …” at runtime.
We now:
 - Keep external deps as-is (e.g. react).
 - For local files:
    • keep bare/aliased specifiers (e.g. components/…), stripping extension.
    • convert relative imports to app-relative specifiers via `toAppSpecifier`, normalizing slashes.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/nextjs/src/tools/codegen/import-map.ts
+++ b/packages/nextjs/src/tools/codegen/import-map.ts
@@ -5,7 +5,7 @@ import { debug } from '@sitecore-content-sdk/core';
 import { getComponentList } from '@sitecore-content-sdk/core/tools';
 import { SitecoreConfig } from '@sitecore-content-sdk/core/config';
 import crypto from 'crypto';
-import { isBare, stripExtension, toAppSpecifier, toPosix } from './utils';
+import { isNodeModuleImport, stripExtension, getRelativeImportPath, toPosixPath } from './utils';
 
 let _getComponentList = getComponentList;
 const aliasImport = /^([a-zA-Z0-9]+) as .+$/;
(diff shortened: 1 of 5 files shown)


packages/nextjs/src/tools/codegen/utils.ts
```diff
@@ -244,3 +244,18 @@ export const sendCode = async ({
   }
   return file.path;
 };
+
+// Normalize path separators to POSIX-style "/" for cross-platform consistency.
+export const toPosix = (p: string) => p.replace(/\\/g, '/');
```
art-alexeyenko: for readability, let's rename this to something like "getPosixPath"

packages/nextjs/src/tools/codegen/utils.ts
```diff
@@ -244,3 +244,18 @@ export const sendCode = async ({
...
+export const stripExtension = (p: string) => p.replace(/\.(tsx?|jsx?|mjs|cjs)$/, '');
+
+// Convert an absolute file path into an app-relative module specifier (POSIX), optionally stripping a leading "src/".
+export const toAppSpecifier = (absFile: string, appPath: string) => {
+  let rel = toPosix(path.relative(appPath, absFile));
+  if (rel.startsWith('src/')) rel = rel.slice(4);
```
art-alexeyenko: not sure we need to slice `src/` part. OOB Content SDK has aliases to to handle sub-src folders, but not all custom Content SDK apps will have it

packages/nextjs/src/tools/codegen/utils.ts
```diff
@@ -244,3 +244,18 @@ export const sendCode = async ({
   }
   return file.path;
 };
+
+// Normalize path separators to POSIX-style "/" for cross-platform consistency.
+export const toPosix = (p: string) => p.replace(/\\/g, '/');
+export const stripExtension = (p: string) => p.replace(/\.(tsx?|jsx?|mjs|cjs)$/, '');
+
+// Convert an absolute file path into an app-relative module specifier (POSIX), optionally stripping a leading "src/".
+export const toAppSpecifier = (absFile: string, appPath: string) => {
```
art-alexeyenko: what do you think of calling it `getRelativeImportPath` ?

packages/nextjs/src/tools/codegen/utils.ts
```diff
@@ -244,3 +244,18 @@ export const sendCode = async ({
...
+  let rel = toPosix(path.relative(appPath, absFile));
+  if (rel.startsWith('src/')) rel = rel.slice(4);
+  return stripExtension(rel);
+};
+
+// Determine if a specifier is bare/aliased (i.e., not relative "./" or "../" and not absolute "/").
+export const isBare = (name: string) =>
```
art-alexeyenko: similarly here, what do you think of giving it more specific name, i.e. `isRelativeAppImport` ? Or maybe `isNodeModuleImport` ?

art-alexeyenko (approved):
Great approach, thank you for seeing it through Addy. Leaving some comments on naming, mainly.

packages/nextjs/src/tools/codegen/utils.ts
yavorsk: maybe we need unit tests for the new methods?

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -187,75 +193,92 @@ export const getImportMap = (paths: string[]) => {
...
+        if (!resolvedImportPath) {
+          console.warn('[Codegen] Could not resolve a file for import %s', moduleName);
+          return;
+        }
 
-          imports.named.forEach((importEntry) => {
-            const sameModuleNamedAlready = moduleEntry!.namedExports.has(importEntry);
+        let importModuleName: string;
+
+        // if import path points to a file in local app - process import path to the file (i.e. ./myComponent)
```
yavorsk: i think its better to move these comments inside the if() similar to the rest of the cases

yavorsk (approved):
few comments, approved in advance,

packages/nextjs/src/tools/codegen/utils.ts
addy-pathania: i just added it as a safety check but i will remove it if this will be handled.

packages/nextjs/src/tools/codegen/utils.ts
addy-pathania: Yeah i was not too particular about naming these but these names makes sense.
Updated all.

PR #230, opened 2025-09-29
[core][nextjs] Implementation of separate component map for app router

MenKNas:
- [x] Changelog updated

## Description / Motivation
The current implementation adds dual component map generation for Next.js App Router projects to prevent environment poisoning between server and client components.

Key Changes

- Dual Maps: Generates both component-map.ts (all components) and component-map.client.ts (client-safe only)
- Component Type Detection: Auto-detects 'use client' directives and explicit componentType exports ('server'|'client'|'universal')
- Router Auto-Detection: Automatically identifies App Router vs Pages Router from project structure
- Environment Separation: Server uses regular map, client uses client-safe map
- Backward Compatible: Pages Router projects continue to generate single map

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/core/src/tools/templating/components.ts
+++ b/packages/core/src/tools/templating/components.ts
@@ -171,7 +171,7 @@ export function detectComponentType(filePath: string): ComponentType {
       // Check for explicit componentType export (improved detection)
       if (ts.isVariableStatement(node)) {
         const hasExportModifier = node.modifiers?.some(
-          (modifier: ts.Modifier) => modifier.kind === ts.SyntaxKind.ExportKeyword
+          (modifier: ts.ModifierLike) => modifier.kind === ts.SyntaxKind.ExportKeyword
         );
 
         if (hasExportModifier) {
--- a/packages/core/src/tools/templating/components.ts
+++ b/packages/core/src/tools/templating/components.ts
@@ -1,8 +1,12 @@
 import * as glob from 'glob';
+import * as fs from 'fs';
 
 const componentNamePattern = /^[\/]*(.+[\/\\])*(.+)\.[jt]sx?$/;
 
 const componentPathPattern = /^([\/]*.+[\/\\].+)\..+$/;
+
+export type ComponentType = 'server' | 'client' | 'universal';
+export type RouterType = 'app' | 'pages';
 
 /**
  * Describes a file that represents a component definition
@@ -17,6 +21,11 @@ export interface ComponentFile {
   importPath: string;
   moduleName: string;
   componentName: string;
+  componentType?: ComponentType;
+}
+
+export interface ComponentFileWithType extends ComponentFile {
+  componentType: ComponentType;
 }
 
 /**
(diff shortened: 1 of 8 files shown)


packages/core/src/tools/templating/components.ts
```diff
@@ -72,3 +81,66 @@ export function getComponentList(paths: string[], exclude?: string[]): Component
...
+
+  return 'pages';
+}
+
+export function detectComponentType(filePath: string): ComponentType {
+  try {
+    const content = fs.readFileSync(filePath, 'utf-8');
+
+    // Check for 'use client' directive
+    if (content.includes("'use client'") || content.includes('"use client"')) {
```
stasmaxymov: I think simple "includes" check is not enough. Checking that first line is not complete either. I advise AST based checks, there are some example of similar stuff in the repo (e.g. import-map stuff)
It will also help with `componentType` export

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -18,23 +22,119 @@ export const generateMap: GenerateMapFunction = ({
   exclude,
   componentImports,
   mapTemplate = nextjsMapTemplate,
+  routerType,
```
stasmaxymov: I think explicit option to enable/disable client component map should be named like `clientComponentMap: boolean` instead of  `routerType`. Because you might want to use single component map in case you use only server or client components in project

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -18,23 +22,119 @@ export const generateMap: GenerateMapFunction = ({
...
+  componentImports?: ComponentImport[]
+): string => {
+  const wildcardImports: string[] = [];
+  const namedImports: string[] = [];
+  const componentMapEntries: string[] = [];
+
+  components.forEach((component) => {
+    wildcardImports.push(`import * as ${component.moduleName} from '${component.importPath}';`);
+    // Add fallback componentType if not exported by the component
+    wildcardImports.push(
```
stasmaxymov: It seems these `componentType` assignments are mixed between imports which are going to look strange in component map. 
At minimum, we should separated them from imports section. 
But I even would argue that we don't need them in the current form. 

The only scenario where we need to explicitly set those is when component uses 'use client' directive, so we need only to modify such component entries and we can do that inline when add them to component map, e.g.:
`['Promo', {...Promo, componentType: 'client' }]`

faiqamin:
Looks good @MenKNas. Do we need to add some tests in here? Maybe to cover the map generation correctness or the classification rules?

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -18,23 +22,119 @@ export const generateMap: GenerateMapFunction = ({
   exclude,
   componentImports,
   mapTemplate = nextjsMapTemplate,
```
stasmaxymov, line 36: we have a way to override templating from cli.config, we probably need to extend it to work with client map too

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -18,23 +22,119 @@ export const generateMap: GenerateMapFunction = ({
...
+      console.error(
+        `Component Map generation failed. Error writing to file ${destination}:`,
+        error
+      );
+      throw error;
+    }
   }
 };
 
+const nextjsMapTemplateWithTypes = (
```
stasmaxymov: Now we end-up with 3 copies of very similar templating functions, is there a way to combine them?

stasmaxymov (changes requested):
Overall looks good, please see some improvement comments

MenKNas:
Hello guys, PR update:

@faiqamin updated tests for `components.ts` to include new scenarios
@stasmaxymov addressed your comments and refactored some parts

stasmaxymov (approved):

PR #233, opened 2025-10-03
[cli][nextjs] Fix client map generation control issue

MenKNas:
- [ ] Changelog updated

## Fixes

- **CLI Option Handling**: Fixed `clientComponentMap` option not being passed from `sitecore.cli.config.ts` to the generator, allowing users to explicitly control dual component map generation

- **Template Formatting**: Cleaned up generated component map files to remove excessive empty lines when no custom components exist, maintaining one clean line between imports and exports for readability

## Testing Details
- [ ] Unit Test Added
- [x] Manual Test/Other => Testing: Verified by setting `clientComponentMap: false` in `sitecore.cli.config.ts` and confirming only `component-map.ts` was generated while `component-map.client.ts` was correctly skipped, then tested auto-detection still works when the option is omitted.

## Types of changes

### DIFF
--- a/packages/cli/src/scripts/project/component/generate-map.test.ts
+++ b/packages/cli/src/scripts/project/component/generate-map.test.ts
@@ -87,4 +87,27 @@ describe('generate-map CLI', () => {
     expect(generatorStub.firstCall.args[0]).to.deep.equal(args);
     expect(consoleLogStub.calledWithMatch(/Generating component map/)).to.be.true;
   });
+
+  it('should pass clientComponentMap: false to generator when explicitly set to false', () => {
+    const generatorStub = sinon.stub();
+    const args = {
+      paths: ['src'],
+      destination: 'dest',
+      componentImports: ['pkg'],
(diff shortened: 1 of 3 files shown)


packages/cli/src/scripts/project/component/generate-map.test.ts
```diff
@@ -73,6 +73,7 @@ describe('generate-map CLI', () => {
       destination: 'dest',
```
illiakovalenko, line 73: Since the client component map was being generated even when it the setting was set to false, perhaps it would make sense to add an extra unit test to ensure the issue you fixed doesn’t regress in the future?

illiakovalenko (approved):
Looks good 👍 
Approved in advance, please, have a look at one comment below

packages/cli/src/scripts/project/component/generate-map.test.ts
MenKNas: Good point, will update the test before merging.

PR #243, opened 2025-10-14
[create-content-sdk-app] Prevent CloudSDK re-initialization on client-side navigation

MenKNas:
- [x] Changelog updated

## Description / Motivation
The issue was that CloudSDK was re-initializing on every route change due to Bootstrap component unmounting/remounting with each navigation, causing unnecessary overhead and potential side effects.

### Solution
Moved Bootstrap component from `[[...path]]/page.tsx` to `[site]/layout.tsx`. This keeps the component mounted across navigations since the site-level layout has a stable React key that only changes when the site changes.

With this fix, CloudSDK now initializes once per site and only re-initializes when switching between sites. Client-side navigation within the same site skips initialization entirely.

### Changes
- Created `app/[site]/layout.tsx` with Bootstrap component
- Updated `Bootstrap.tsx` to accept `siteName` prop and use `useRef` for initialization tracking
- Removed Bootstrap from `page.tsx`

## Testing Details
- [ ] Unit Test Added
- [x] Manual Test/Other 

Verified fix by adding test navigation component with Next.js Links to sample app. Navigated between multiple pages and confirmed:

- Initial page load: CloudSDK initializes once
- Client-side navigation: Using Console logs, it shows "already initialized - skipping" with no unmount messages
- Full page reload: Re-initializes correctly (expected behavior for new app instance)
- Cross-site navigation: Re-initializes with new site name (expected behavior)

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -20,15 +20,16 @@ Our versioning strategy is as follows:
   - Editing and preview support ([#198](https://github.com/Sitecore/content-sdk/pull/198))
   - Internationalization support ([#202](https://github.com/Sitecore/content-sdk/pull/202)) ([#214](https://github.com/Sitecore/content-sdk/pull/214))
   - Client-server component map separation ([#230](https://github.com/Sitecore/content-sdk/pull/230))([#232](https://github.com/Sitecore/content-sdk/pull/232))([#234](https://github.com/Sitecore/content-sdk/pull/234))([#235](https://github.com/Sitecore/content-sdk/pull/235))([#241](https://github.com/Sitecore/content-sdk/pull/241))
-  - Prevent CloudSDK re-initialization on client-side navigation ([#243](https://github.com/Sitecore/content-sdk/pull/243))
+ 
 * [nextjs] Slim down sample even more ([#225](https://github.com/Sitecore/content-sdk/pull/225))
 * Mark client components with `use client` directive. ([#226](https://github.com/Sitecore/content-sdk/pull/226))
 * Add LLMs and Copilot instruction files for improved AI guidance ([#239](https://github.com/Sitecore/content-sdk/pull/239))
 
-### 🐛 Bug Fixes	
+### 🐛 Bug Fixes
 
 * `[template/nextjs]` `[template/next-app-router]` Fix imports for SSR/SSG ([#229](https://github.com/Sitecore/content-sdk/pull/229))
 * `[template/next-app-router]` Guard static params generation and harden not-found routes for XM Cloud ([#242](https://github.com/Sitecore/content-sdk/pull/242))
+* `[template/next-app-router]` Prevent CloudSDK re-initialization on client-side navigation ([#243](https://github.com/Sitecore/content-sdk/pull/243))
 
 ### 🧹 Chores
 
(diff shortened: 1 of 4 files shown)


CHANGELOG.md
```diff
@@ -20,11 +20,12 @@ Our versioning strategy is as follows:
   - Editing and preview support ([#198](https://github.com/Sitecore/content-sdk/pull/198))
   - Internationalization support ([#202](https://github.com/Sitecore/content-sdk/pull/202)) ([#214](https://github.com/Sitecore/content-sdk/pull/214))
   - Client-server component map separation ([#230](https://github.com/Sitecore/content-sdk/pull/230))([#232](https://github.com/Sitecore/content-sdk/pull/232))([#234](https://github.com/Sitecore/content-sdk/pull/234))([#235](https://github.com/Sitecore/content-sdk/pull/235))([#241](https://github.com/Sitecore/content-sdk/pull/241))
+  - Prevent CloudSDK re-initialization on client-side navigation ([#243](https://github.com/Sitecore/content-sdk/pull/243))
```
illiakovalenko: Can be moved under "bug fixes" section, since we already have a few entries there

CHANGELOG.md
```diff
@@ -20,11 +20,12 @@ Our versioning strategy is as follows:
...
 * [nextjs] Slim down sample even more ([#225](https://github.com/Sitecore/content-sdk/pull/225))
 * Mark client components with `use client` directive. ([#226](https://github.com/Sitecore/content-sdk/pull/226))
 * Add LLMs and Copilot instruction files for improved AI guidance ([#239](https://github.com/Sitecore/content-sdk/pull/239))
 
-### 🐛 Bug Fixes
+### 🐛 Bug Fixes
```
illiakovalenko:
```suggestion
### 🐛 Bug Fixes
```

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/Bootstrap.tsx
```diff
@@ -1,48 +1,47 @@
...
+const Bootstrap = ({ siteName }: { siteName: string }): JSX.Element | null => {
+  const searchParams = useSearchParams();
+  const sdkInitialized = useRef(false);
+  const currentSiteName = useRef<string | null>(null);
 
   useEffect(() => {
-    if (!page) {
-      return;
-    }
+    const isPreviewMode = searchParams.has('sc_itemid') || searchParams.has('sc_mode');
```
illiakovalenko: Can we check if "draftMode" is enabled and rely on that? Draft mode should be enabled when rendering app in preview or editing mode
```tsx
layout.tsx

import { draftMode } from 'next/headers';
import { Suspense } from 'react';

export default async function SiteLayout({
  children,
  params,
}: {
  children: React.ReactNode;
  params: Promise<{ site: string }>;
}) {
  const { site } = await params;
  const { isEnabled } = await draftMode();
  return (
    <>
      <Suspense>
        <Bootstrap siteName={site} isPreviewMode={isEnabled} />
      </Suspense>
      {children}
    </>
  );
}

```

Another problem that when you rely on `searchParams` and wrap Bootstrap component by `Suspense` this component is rendered at the very end of the cycle, so `CDPPageView` is rendered earlier which is trying to fire `events` but `CloudSDK` is not initialized yet. You can check console in the browser and you will see warning message.

By relying on `draftMode` you are able to remove `Suspense` so the order will remain correct as it was.
You need to test it

packages/create-content-sdk-app/src/templates/nextjs-app-router (beta)/src/Bootstrap.tsx
```diff
@@ -1,48 +1,47 @@
...
- * @param props
- */
-const Bootstrap = (props: SitecorePageProps): JSX.Element | null => {
-  const { page } = props;
-
-  // Browser ClientSDK init allows for page view events to be tracked
+const Bootstrap = ({ siteName }: { siteName: string }): JSX.Element | null => {
+  const searchParams = useSearchParams();
+  const sdkInitialized = useRef(false);
+  const currentSiteName = useRef<string | null>(null);
```
illiakovalenko: Do we need to keep these extra states? When `site` is changed - Bootstrap is fully re-rendered. I believe when you will navigate within the same site - Bootstrap will not be re-rendered. Can we remove it? Please, double check. `useEffect` should be able to avoid extra re-render cycles

illiakovalenko (changes requested):

illiakovalenko (approved):
Looks good! 👍

PR #244, opened 2025-10-15
[AI Component Generation] A new CLI command to add a new component variant

illiakovalenko:
- [x] Changelog updated

## Description / Motivation

* Introduced a new `sitecore-tools project component add <variant-id>` CLI command to add a generated variant to the application:
  * Token-based authentication is used to access the component variants source
  * Introduced an `override` flag to allow overwriting existing variants.
* Introduce `scConfig` property in `sitecore.cli.config`:
  * The `scConfig` property is used to pass the Sitecore configuration to the CLI commands.
  * Build commands now receive arguments (e.g. `scConfig`) to allow access to the CLI specific configuration settings.
  * Now it's optional to pass `scConfig` manually to the command constructor.
  * These changes were introduced as optional for customers and will be fully adopted and marked as required in the next major release

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/cli/src/scripts/project/component/add.ts
```diff
@@ -0,0 +1,181 @@
...
+      }
+    }
+
+    const registryUrl = getComponentRegistryUrl({
+      variantId,
+      contextId,
+      targetPath: targetPath as string,
+    });
+
+    execSync(`npx shadcn@latest add ${registryUrl}${overwrite ? ' --overwrite' : ''}`, {
```
stasmaxymov: I think we should pin certain version of shadcn, as latest could take a new major version our call is not compatible with

packages/core/src/tools/generateSites.ts
```diff
@@ -29,46 +30,50 @@ export type GenerateSitesConfig = {
...
- * @param {string} config.destinationPath - The optional path where the generated sites file will be written. Defaults to '.sitecore/sites.json'.
  * @returns {Promise<Function>} - A promise that resolves to an asynchronous function that fetches site information and writes it to a file.
  */
 export const generateSites = ({
-  scConfig,
+  scConfig: deprecatedScConfig,
   destinationPath,
-}: GenerateSitesConfig): (() => Promise<void>) => {
-  return async () => {
+}: GenerateSitesConfig = {}): (() => Promise<void>) => {
```
stasmaxymov, line 38: Lets wrap scConfig in obj (`{ scConfig: SitecoreConfig}`). Build commands are executed in certain CLI context which currently consists of `SitecoreConfig` but could be extended with cli specific, or command specific stuff

packages/core/src/tools/codegen/component-variant.ts
stasmaxymov, line 78: I think we should also cover 401 error, it will be returned when token is invalid or expired

packages/cli/src/scripts/project/component/add.ts
```diff
@@ -157,13 +168,14 @@ export async function handler(argv: AddArgs) {
...
+    const variantSpecUrl = getComponentVariantSpecUrl({
       variantId,
-      contextId,
       targetPath: targetPath as string,
+      edgeUrl,
+      token,
     });
 
-    execSync(`npx shadcn@latest add ${registryUrl}${overwrite ? ' --overwrite' : ''}`, {
+    execSync(`npx shadcn@^3.4.2 add ${variantSpecUrl}${overwrite ? ' --overwrite' : ''}`, {
```
stasmaxymov: We might need quotes around `variantSpecUrl`, depending on what values are present in token

stasmaxymov (approved):

PR #245, opened 2025-10-17
[core][nextjs] Enable component variants in component-map & also send variants files for code extraction

addy-pathania:
- [x] Changelog updated

## Description / Motivation
1) Component map (generate-map)
   - Variant grouping: Neighbor files in the same folder with a `BaseName.suffix.tsx` pattern are grouped under a single component    key and merged via object spread.
   - Introduced a flag `includeVariants` to switch between group variants and one entry per file. True by default.
   - Client map support preserved: Works for both App Router (two maps) and Pages Router (single map) paths.

Example:
`src/components/PromoBlock/PromoBlock.tsx` (base)
`src/components/PromoBlock/PromoBlock.but-cooler.tsx` (variant file)
//component-map
`['PromoBlock', { ...PromoBlockbutcooler, ...PromoBlock }]`

2) Code extraction (extractFiles)
   - When variants are not included the behavior is identical to previous releases:
      - Component map emits one entry per file.
      - Code extraction sends one item per file with `variantNames` as  labels.
   - Existing custom templates and componentImports are unchanged.
   

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/cli/src/scripts/project/component/generate-map.test.ts
```diff
@@ -5,7 +5,7 @@ import * as generateMapModule from './generate-map';
 import * as loadConfigModule from '../../../utils/load-config';
 import * as watchItemsModule from '../../../utils/watch-items';
 
-describe('generate-map CLI', () => {
+describe.only('generate-map CLI', () => {
```
art-alexeyenko: likely a leftover from local testing

packages/nextjs/src/utils/utils.ts
```diff
@@ -77,3 +77,12 @@ export const parseRewriteHeader = (headers: Headers) => {
...
   return { site, locale };
 };
+
+/**
+ * Filters out variant components (names containing a '.').
+ * e.g., "Teaser.Variant1" → removed, "Teaser" → kept
+ * @param {Array} components Array of components to filter
+ * @returns {Array} Components without variant suffixes
+ */
+export const stripVariants = <T extends { componentName: string }>(components: T[]): T[] =>
```
art-alexeyenko: I believe just checking for a dot will not reliably identify variant component files.
We should be able to simplify the whole approach to no longer require this function.

packages/core/src/tools/templating/utils.ts
```diff
@@ -100,3 +101,107 @@ export function getItems<Item>(settings: GetItemsSettings<Item>): Item[] {
...
+
+  return componentGroups;
+}
+
+/**
+ * Produces entries where each base component key aggregates its variants via object spread.
+ * @param {ComponentSource[]} components list of components
+ * @returns {ComponentMapEntry[]} list of map entries
+ */
+export function groupComponentsWithVariants(components: ComponentSource[]): ComponentMapEntry[] {
```
art-alexeyenko: maybe this can be simplified and this function can be cut too.
i.e. we can perform the grouping in generateMap function itself and have a components list with:
- components grouped with a variant
- components with no variants
You already do a similar logic, but in buildNextMapContent. I believe this should be done at the "getting components list" step insted - I will outline my prosposal more in my next comment

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -123,70 +27,55 @@ type TemplateOptions = {
...
-        `['${component.moduleName}', {...${component.moduleName}, componentType: 'client'}]`
-      );
-    } else {
-      componentMapEntries.push(`['${component.moduleName}', ${component.moduleName}]`);
-    }
-  });
+  const entries: string[] = [];
+
+  // Build per-entry imports/values via grouping
+  const groups = grouping(components);
```
art-alexeyenko: logically, grouping components at the templating step for component-map is too late. The template function should be as simple as possible and all the transformed components entries should just be passed into it already.
I propose an alterative approach:
- modify `ComponentFile` type or introduce a inherited type that contains groups for component-map. Add a variant array into it that will be populated when we retrieve components for component-map. This will describe components passed into `buildNextjsMapContent`
- `buildNextjsMapContent` will check if that array is empty, and output a spread for a map entry if it isn't
- This should provide a simpler and a more direct flow that is easier to manage

packages/cli/src/scripts/project/component/generate-map.ts
```diff
@@ -62,7 +62,7 @@ export function handler(argv: GenerateMapCliArgs) {
     return;
   }
   const componentMapGenerator = cliConfig.componentMap.generator;
-  const { paths, destination, componentImports, exclude, clientComponentMap } =
+  const { paths, destination, componentImports, exclude, clientComponentMap, enableVariantsInMap } =
```
art-alexeyenko: if variants are enabled by default, let's rename `enableVariantsInMap` into `ignoreVariants` so that CLI function call can have fewer required params

art-alexeyenko (commented):
A good start and we can streamline the implementation further for better maintainability.
component-map generation is complex, but can be narrowed down to two steps:
- getting components to be registered, based on app type, context etc
- generating the component-map file for those components
To sum up the simpler approach I propose in individual comments:
- The decision to process/not process variants and the grouping of variants should be tightly coupled together and both can be done in the first step
- Either the `getComponents...` methods themselves or the logic around them can perform the filtering and grouping of variant (i.e.  for filtering, `getComponents` method itself accepts an `exclude` pattern list which can remove variants from consideration when variants flag is off)
- the resulting components can be passed into the function that builds component map file (this function should remain as straightforward as possible)

This should help reduce the amount of code for this feature and make the extra complication to component map logic be less complex

addy-pathania:
> A good start and we can streamline the implementation further for better maintainability. component-map generation is complex, but can be narrowed down to two steps:
> 
> * getting components to be registered, based on app type, context etc
> * generating the component-map file for those components
>   To sum up the simpler approach I propose in individual comments:
> * The decision to process/not process variants and the grouping of variants should be tightly coupled together and both can be done in the first step
> * Either the `getComponents...` methods themselves or the logic around them can perform the filtering and grouping of variant (i.e.  for filtering, `getComponents` method itself accepts an `exclude` pattern list which can remove variants from consideration when variants flag is off)
> * the resulting components can be passed into the function that builds component map file (this function should remain as straightforward as possible)
> 
> This should help reduce the amount of code for this feature and make the extra complication to component map logic be less complex

- I deliberately kept previous component pre-processing logic stable (i.e. no changes to `ComponentFile` contract since it is used in multiple places and that change can overflow) to avoid breaking other tool consumers. We isolated the new logic (grouping) near the map-building concerns, so:
  - it’s easy to unit test in isolation,
  - it doesn’t affect other code paths that use component discovery.
- The current design already has two clean stages:
  - discovery → `getComponentList`
  - transformation → grouping helpers
  - render → template
 
In short: the approach minimized surface area change and kept backwards compatibility risk low.
I would like to push back on your proposal of putting the grouping logic directly inside `generateMap`. That function is already the orchestrator (deciding app/pages router, dual-map vs single-map, custom templates, destinations, error handling). In lining grouping there couples orchestration with transformation and will bloat it over time, making it even harder to test, maintain and evolve. 
Although, i agree with your comment about keeping the template function as simple as possible so I propose a middle ground. Introduce a more focused pre-processor that sits next to `generateMap`, not in it, something like:
```
const entries = await resolveEntriesForMap(paths, exclude, { includeVariants });
const content  = buildNextjsMapContent(entries, ...);
```
This way decision to process variants and the grouping itself happen before templating.

packages/cli/src/scripts/project/component/generate-map.ts
addy-pathania: `includeVariants`?

art-alexeyenko:
@addy-pathania 
> Introduce a more focused pre-processor that sits next to generateMap, not in it, something like: ...

Yeah, I like this. Let's go with this.

packages/cli/src/scripts/project/component/generate-map.ts
art-alexeyenko: `ignoreVariants` might be better. It is a flag that is false by default. If this flag is taken from sitecore.cli.config like it is right now:
```
const { paths, destination, componentImports, exclude, clientComponentMap, enableVariantsInMap } =
    cliConfig.componentMap;
```
then having it false by default is easier. we don't need to explicitly mention it in config, and users will only add it when disabling variants.

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -196,43 +181,118 @@ export const componentMap = new Map<string, NextjsContentSdkComponent>([
...
   components: ComponentFileWithType[],
+  includeVariants: boolean,
   componentImports?: ComponentImport[]
 ): string => {
-  return nextjsUnifiedTemplate(components, componentImports, {
-    headerComment:
-      "Below are built-in components that are available in the app, it's recommended to keep them as is",
+  // We don’t pre-process here because this is the legacy “custom template” hook.
+  // This function exists purely as our default client template implementation.
+  const entries = prepareComponentsForMap(components, { includeVariants });
```
art-alexeyenko: is this a duplicate call? the prepare function is being called in generateMap before this too

packages/nextjs/src/tools/generate-map.ts
```diff
@@ -196,43 +181,118 @@ export const componentMap = new Map<string, NextjsContentSdkComponent>([
...
-  return nextjsUnifiedTemplate(components, componentImports, {
-    headerComment: 'Client-safe component map for App Router',
-    isClientMap: true,
-  });
+  if (shouldGenerateClientMap) {
+    // App Router case, main map
+    const componentsWithTypes = getComponentListWithTypes(paths, exclude);
+    let mainContent: string;
+    if (mapTemplate) {
+      mainContent = mapTemplate(componentsWithTypes, includeVariants, componentImports);
```
art-alexeyenko: should components be prepared in this clause too?
also, maybe the getComponentList/getComponentListWithTypes and prepareComponentsForMap can be bundled together so as to not repeat code and ensure all the required prep steps are always followed

packages/nextjs/src/tools/generate-map.ts
addy-pathania: its not duplicate since client-map components will be different so we need to process them again.

packages/core/src/tools/codegen/utils.ts
```diff
@@ -48,179 +107,234 @@ const netlifyDeploy = () => !!process.env.NETLIFY && !!process.env.BUILD_ID;
...
+      const modName = ident.text;
+      const spec = nameSpaceImports.get(modName);
+      if (!spec) return;
+      const fileAbs = resolveWithTs(spec, mapPath, compilerOptions);
+      if (!fileAbs) return;
+      results.push({
+        componentKey,
+        moduleName: modName,
+        importPath: spec,
+        filePath: toPosixPath(fileAbs),
```
art-alexeyenko, line 226: please double check this works on both Windows and UNIX systems

packages/core/src/tools/codegen/extract-files.ts
```diff
@@ -42,6 +48,9 @@ function _extractFiles(args: ExtractFilesConfig) {
     authority: process.env.SITECORE_AUTH_AUTHORITY,
     audience: process.env.SITECORE_AUTH_AUDIENCE,
   };
+
+  const enableVariantsInMap = args.enableVariantsInMap ?? true;
```
art-alexeyenko: what will happen if `enableVariantsInMap = true` is passed into extractFiles, but `enableVariantsInMap = false` is passed into `componentMap` ?

packages/core/src/tools/codegen/extract-files.ts
```diff
@@ -54,31 +63,74 @@ function _extractFiles(args: ExtractFilesConfig) {
...
+        });
+      } else {
+        console.error(chalk.red('resolveComponentImportFiles: unexpected return shape'));
+        return;
+      }
+
+      const fileDispatches: Promise<string | null>[] = [];
+
+      for (const { componentKey, filePath } of items) {
+        if (!fs.existsSync(filePath)) {
```
art-alexeyenko: this check is already present in `sendCode`

packages/core/src/tools/codegen/utils.ts
```diff
@@ -48,179 +107,234 @@ const netlifyDeploy = () => !!process.env.NETLIFY && !!process.env.BUILD_ID;
...
-          const componentValue = importStringsMap[componentImport];
-          componentImportsMap.set(componentKey, componentValue);
-        }
-      });
-    } else if (node.getChildCount() > 0) {
-      ts.forEachChild(node, (childNode) => {
-        traverseNewStatement(childNode);
+  });
+
+  // 2) Find the map literal: new Map([...]) and collect entries
```
art-alexeyenko, line 205: you also need to look for map.set() calls. Nothing prevents users from adding entries outside of the `new` call

packages/core/src/tools/codegen/utils.ts
```diff
@@ -48,179 +107,234 @@ const netlifyDeploy = () => !!process.env.NETLIFY && !!process.env.BUILD_ID;
...
+  const results: ResolvedImport[] = [];
+
+  const addEntry = (componentKey: string, ident: ts.Expression) => {
```
art-alexeyenko: please add comments or more descriptive names to `addEntry` and `visit` functions

packages/core/src/tools/codegen/utils.ts
addy-pathania: Yes it should but i have added a bit more safer version.

packages/core/src/tools/codegen/extract-files.ts
```diff
@@ -54,31 +63,74 @@ function _extractFiles(args: ExtractFilesConfig) {
...
-            path: mapEntry[1],
-            type: ExtractedFileType.Component,
-          },
-          token: accessToken,
-          targetUrl,
-        })
-      );
+      let items: Array<{ componentKey: string; filePath: string }> = [];
+
+      if (resolvedImports && typeof (resolvedImports as ResolveResult).imports !== 'undefined') {
```
art-alexeyenko: this should ideally be done under `resolveComponentImportFiles`. I also feel extraLabels overcomplicates the logic, it's very hard to follow what is being done here (but `resolveComponentImportFiles` refactor is very welcome)

packages/core/src/tools/codegen/utils.test.ts
```diff
@@ -115,174 +157,239 @@ describe('codegen-utils', () => {
...
     });
 
-    it('should return JS imports with absolute paths from component-map.ts', () => {
+    it('returns identifier imports with absolute paths from default .sitecore/component-map.ts', () => {
       const appPath = './src/tools/codegen/test-data/extract-components/regular-imports';
-      const imports = codegenUtils.resolveComponentImportFiles(appPath);
-      expect(Array.from(imports)).to.deep.equal([
+      const res = codegenUtils.resolveComponentImportFiles(appPath);
+
+      const actual = res.imports.map((i) => ({ ...i, filePath: norm(i.filePath) }));
```
art-alexeyenko: I feel like this and other tests don't assert the return data for resolveComponentImportFiles fully, only select parts of it. It's better to test the full returned object, it will provide more insight into how function works and make it more reliable.

packages/core/src/tools/codegen/utils.test.ts
```diff
@@ -115,174 +157,239 @@ describe('codegen-utils', () => {
...
             './src/tools/codegen/test-data/extract-components/with-path-aliases/src/components/TestComponent.tsx'
-          ),
-        ],
+          )
+        ),
       ]);
     });
 
-    it('should ignore imports starting with "node:", ending with ".d.ts" and containing "node_modules"', () => {
+    it('ignores non-relevant imports (node:, *.d.ts, node_modules) in the map entries', () => {
```
art-alexeyenko, line 404: This test passes because node_modules import is not being used in test component map.
If you add
```
  ['fs', fs],
  ['index', index],
```
in component map init under node-modules-imports, it will fail

packages/nextjs/src/tools/generate-map.ts
addy-pathania: Updated to prepare components for custom templates too.
I think bundling functions together would be out of scope now and should be done as a separate refactor.

packages/core/src/tools/codegen/extract-files.ts
addy-pathania: i have refactored and simplified the extract-files function a bit so now it should seem simple to understand.

packages/core/src/tools/codegen/extract-files.ts
addy-pathania: everything will be handled correctly, just that in the above case the separate variant files will not be sent for extraction provided if there are any since its variant generation is false for generate-map.

art-alexeyenko: it would be better to have a single point of configuration for this - or alternatively don't pass `enableVariantsInMap` into extract-files. It will just parse component-map with assumption variants may be present, always.

packages/core/src/tools/codegen/extract-files.ts
```diff
@@ -54,31 +59,49 @@ function _extractFiles(args: ExtractFilesConfig) {
...
+        fileDispatches.push(
+          sendCode({
+            file: {
+              name: componentKey,
+              path: filePath,
+              type: ExtractedFileType.Component,
+            },
+            token: accessToken,
+            targetUrl,
+            extraLabels,
```
art-alexeyenko: What do you think about adding extraLabels as an optional field in ExtractedFile type instead?

packages/nextjs/src/tools/generate-map.ts
art-alexeyenko: I don't think it's out of scope now. We have 4 different usecases for getting components for component-map. If ever, in the future, someone adds 2 more and forgets to call `prepareComponentsForMap` for them, we'll have a broken logic. This needs to be prevented.

packages/core/src/tools/codegen/extract-files.ts
addy-pathania: yes that's a good idea.

packages/core/src/tools/codegen/extract-files.ts
addy-pathania: sounds good. Removed from extract-files.

packages/nextjs/src/tools/generate-map.ts
addy-pathania: I have introduced a small helper `collectComponents` that returns { raw, entries }. All call sites (main map and client map) will use it, so there’s a single place that performs `prepareComponentsForMap`. The custom templates gets both the raw list (componentsWithTypes) and the pre-processed entries. That allows existing templates to keep using the raw list while new templates can rely on the prepared shape. So the preparation step isn’t skipped.

art-alexeyenko (approved):
Good to go with unit test adjustments.
Solid job, Addy, these refactors are a pain to implement, but very welcome to have.

PR #248, opened 2025-10-20
[nextjs][App Router] FEAAS and BYOC support

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
- Adds server wrappers for byoc and feaas components
- Ensures server/client component maps in App Router apps use correct implementations
- Server implementations fetch server props on their own instead of using  `getComponentServerProps`
- Refactors FEAAS/BYOC implementations for better visibility and navigation (it was getting a bit out of hand)
- Backwards compatible with pages router

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/nextjs/src/tools/generate-map.ts
addy-pathania: Please update this file and test file as per latest changes in dev

packages/react/src/components/FEaaS/BYOCServerWrapper.tsx
```diff
@@ -0,0 +1,22 @@
...
+import { fetchBYOCComponentServerProps } from './feaas-utils';
+import { BYOCWrapper } from './BYOCWrapper';
+import { nonSerializedPlaceholderProps } from '../Placeholder/models';
+
+export const BYOCServerWrapper = async (props: BYOCServerWrapperProps) => {
+  const params = props.rendering?.params || {};
+  // only pass serializable props to the client BYOC component
+  const serializableProps = nonSerializedPlaceholderProps.reduce(
+    (finalProps, prop) => {
+      delete finalProps[prop];
```
addy-pathania: i think we should avoid this pattern of mutating via delete in a reduce. We can do something like

```
export const BYOCServerWrapper = async ({
  rendering,
  ...rest
}: BYOCServerWrapperProps) => {
  const params = rendering?.params ?? {};
  
  const serializableProps = Object.fromEntries(
    Object.entries(rest).filter(([key]) => 
      !nonSerializedPlaceholderProps.includes(key as keyof typeof rest as any)
    )
  );
  ...
  ```

File: packages/react/src/components/FEaaS/FEaaSSeverWrapper.tsx [hunk 2]
```diff
@@ -0,0 +1,28 @@
...
+ * @returns {Promise<JSX.Element>} rendered FEaaSWrapper component
+ */
+export const FEaaSServerWrapper = async (props: FEaaSServerWrapperProps) => {
+  const params = props.rendering?.params || {};
+  const isPageStateNormal = props.page?.mode.isNormal;
+  // only pass serializable props to the client FEaaS component
+  const serializableProps = nonSerializedPlaceholderProps.reduce(
+    (finalProps, prop) => {
+      delete finalProps[prop];
+      return finalProps;
```
addy-pathania: same comment as above

packages/react/src/components/FEaaS/feaas-utils.ts
```diff
@@ -0,0 +1,119 @@
...
+
+/**
+ * Fetches server component props required for server rendering, based on rendering params.
+ * @param {BYOCComponentParams} params component params
+ */
+export async function fetchBYOCComponentServerProps(
+  params: BYOCComponentParams
+): Promise<BYOCServerProps> {
+  const fetchDataOptions: FEAAS.DataOptions = params.ComponentDataOverride
+    ? JSON.parse(params.ComponentDataOverride)
```
addy-pathania, line 18: should we add a guard here to check if the passed (params.ComponentDataOverride) is valid as per the type? this can be a bad input.

addy-pathania (approved):
Some minor comments, otherwise looks good to me.

packages/react/src/components/FEaaS/feaas-utils.ts
art-alexeyenko: I think this part was left as it is now initially, in order for the error to be propped up and handled by BYOC error handling. We want the error about invalid input in params to be visible to user so they can fix it.

PR #251, opened 2025-10-23
[core] Fix registration of component logs in console

addy-pathania:
- [x] Changelog updated

## Description / Motivation
Only log registration of components when in registering in component-map otherwise silence them. 

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/core/src/tools/templating/components.ts
```diff
@@ -113,7 +119,10 @@ function _getComponentList(paths: string[], exclude?: string[], silent?: boolean
         .filter((path: string) => path.match(componentNamePattern))
         .map((filePath: string) => {
           const name = filePath.match(componentNamePattern)![2];
-          !silent && console.debug(`Registering Content SDK component ${name}`);
+          const isVariant = name.includes('.');
+          if (!silent && (!isVariant || includeVariants)) {
+            console.debug(`Registering Content SDK component ${name}`);
```
art-alexeyenko: it would be better to remove this log from here altogether and instead log registered components in import-map and generate-map, considering all the modifications we do on the initial list

addy-pathania: We dont need to log it for import map. If we dont need to to log it we can just pass the silent flag.

art-alexeyenko: >We dont need to log it for import map. If we dont need to to log it we can just pass the silent flag.

I think this relates more to the separation of concerns. When we get a list of components and their paths in this method, we're not neccessarily registering them anywhere, so the debug log is misleading. Some other functions in future can use getComponents to just make a list or something.
That's why I think loggin belongs outside of here.

addy-pathania: updated

art-alexeyenko (approved):

PR #253, opened 2025-10-23
[core][App Router] Fix components from client component map not being code extracted

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Fixes the issue where component-map.client.ts was not considered for code extraction

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other - stuff works (Image is client component)
image

## Types of changes


packages/core/src/tools/codegen/extract-files.ts
```diff
@@ -128,7 +140,7 @@ function _extractFiles(args: ExtractFilesConfig = {}) {
         )
       );
     } catch (error) {
-      console.error(chalk.red('Error during code extraction:', error));
+      console.error(chalk.red('Error during code extraction:', error, error.stack));
```
illiakovalenko, line 143: Probably it's a leftover? We don't really need to show stack in logs

packages/core/src/tools/codegen/extract-files.test.ts
```diff
@@ -276,6 +276,44 @@ describe('extract-files', () => {
           );
           expect(msgs).to.include(successMsg);
         });
+
+        xit('should call sendCode for each component path and package.json when client component map present', async () => {
```
illiakovalenko:
```suggestion
        it('should call sendCode for each component path and package.json when client component map present', async () => {
```

packages/core/src/tools/codegen/extract-files.test.ts
illiakovalenko, line 1: We need to add unit tests to cover main scenarios:
* New `clientComponentMapPath` parameter
* Absolute/relative paths
* An edge case when path is present but client map doesn't exist

illiakovalenko (approved):
Looks good! 👍
I’ve left some comments below - the PR is approved in advance so it’s not blocked, but please make sure to address the feedback

PR #270, opened 2025-10-29
[react]Replace rsc-env dependency with private conditional imports

stasmaxymov:
- [ ] Changelog updated

## Description / Motivation
AppPlaceholder used `rsc-env` library to check if current context is RSC. However `rsc-env` doesn't produnce CommonJS target, and it affects NextJs Pages router template, due to mixed usage of CommonJs and ESM.
Replace `rsc-env` with local solution using condition private imports and 'react-server' conventional condition.

## Testing Details

- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/react/package.json
+++ b/packages/react/package.json
@@ -88,10 +88,10 @@
     "#rsc-env": {
       "react-server": {
         "import": "./dist/esm/rsc-utils/rsc.js",
-        "require": "./dist/cjs/rsc-utils/rsc.js"
+        "default": "./dist/cjs/rsc-utils/rsc.js"
       },
       "import": "./dist/esm/rsc-utils/no-rsc.js",
-      "require": "./dist/cjs/rsc-utils/no-rsc.js"
+      "default": "./dist/cjs/rsc-utils/no-rsc.js"
     }
   }
 }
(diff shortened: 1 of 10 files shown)


packages/react/package.json
```diff
@@ -75,14 +75,23 @@
...
-  ]
+  ],
+  "imports": {
+    "#rsc-env": {
+      "react-server": {
+        "import": "./dist/esm/rsc-utils/rsc.js",
+        "require": "./dist/cjs/rsc-utils/rsc.js"
+      },
+      "import": "./dist/esm/rsc-utils/no-rsc.js",
+      "require": "./dist/cjs/rsc-utils/no-rsc.js"
```
addy-pathania: do we want to add a default here so tools that don’t pass conditions still pick “no-RSC” safely?

packages/react/src/rsc-utils/rsc.ts
```diff
@@ -0,0 +1 @@
+export const rsc = true;
```
addy-pathania, line 2: can we add a comment above why we  are doing this?

addy-pathania (approved):
Looks great

PR #273, opened 2025-11-02
[cli] Refactor component handling: rename variant to component, add target path and override validation

illiakovalenko:
- [ ] Changelog updated - Entries will be added once it will be planned to be released

## Description / Motivation
This PR enhances the `sitecore-tools component add` CLI command with improved validation, better error handling, and more generic terminology to support future extensibility.

### Validation Improvements

**Simplified Configuration Validation**
- Removed the requirement for Edge URL during CLI execution
- Now only validates `contextId` and `apiKey` as required fields

**Enhanced Path Security**
- Added comprehensive validation for target paths to prevent security issues:
  - Blocks absolute paths (starting with `/`)
  - Prevents path traversal attacks (using `..`)
  - Validates filename format: `{componentName}.{variantName}.{extension}`
- Example valid path: `components/Promo.WithText.ts`

### Terminology

**Component Naming Standardization**
- Renamed all "component variant" terminology to "component" throughout the CLI
- Made the `add` command more generic to support future component types

### Bug Fixes

**shadcn Override Flag Handling**
- Implemented workaround for known shadcn bug related to the `--override` flag
- Manually handles file overwriting with confirmation prompts

### Improved Developer Experience

**Better Error Messages**
- Enhanced 401 error message: Now explicitly states "The token is incorrect or expired **or the component ID is incorrect**"
- Clarifies that authentication failures can be caused by invalid componentId, not just token issues
- Added clearer validation error messages for path-related issues
- Added practical examples in help text
- `Improved prompt validation by handling invalid input inline with retry logic instead of terminating the CLI process

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/cli/src/scripts/project/component/add.test.ts
```diff
@@ -94,60 +90,61 @@ describe('add command', () => {
...
       });
     execSyncStub = sandbox.stub(childProcess, 'execSync');
     generateMapStub = sandbox.stub(generateMapModule, 'handler');
     inquirerStub = sandbox.stub(inquirer, 'prompt');
-
+    unlinkSyncStub = sandbox.stub(fs, 'unlinkSync');
+    existsSyncStub = sandbox.stub(fs, 'existsSync');
     addModule.unitMocks({
-      getComponentVariantSpec: getComponentVariantSpecStub,
+      getComponentSpec: getComponentSpecStub,
```
stasmaxymov, line 108: This approach for exporting a special mocking function from module looks quite messy for me. 
Why can't we use `sinon.stub` or `sinon.replace` on module itself (like one line above)? It worked well for me in AppPlaceholder tests.

packages/cli/src/scripts/project/component/add.ts
```diff
@@ -97,17 +97,56 @@ export function args(yargs: Argv<AddArgs>) {
...
+ * @param {string} [targetPath] - The target path to validate.
+ * @returns {boolean | string} True if the target path is valid, the error message otherwise.
+ */
+const validateTargetPath = (targetPath?: string) => {
+  if (!targetPath) {
+    return true;
+  }
+
+  // Check for unsafe absolute path starting with "/"
+  if (targetPath.startsWith('/')) {
```
stasmaxymov: There is a `path.isAbsolute(path)` method, I think it is better to use to check if path is absolute. It also has platform specific logic.

packages/cli/src/scripts/project/component/add.ts
```diff
@@ -97,17 +97,56 @@ export function args(yargs: Argv<AddArgs>) {
...
+
+  // Check for path traversal attempts
+  if (targetPath.includes('..')) {
+    return 'Target path cannot contain ".." (path traversal)';
+  }
+
+  const filename = path.basename(targetPath);
+
+  // Validate filename matches {componentName}.{variantName}.{extension} pattern
+  const filenamePattern = /^[^.]+\.[^.]+\.[a-zA-Z0-9]+$/;
```
stasmaxymov: Should we always expect a variant name in the provided custom path? It feels like too restrictive to me, and likely will need to change once we support not only variants but new components

packages/cli/src/scripts/project/component/add.ts
```diff
@@ -162,20 +206,43 @@ export async function handler(argv: AddArgs) {
...
+          message: `File already exists: ${targetPath}. Overwrite?`,
+          default: false,
+        });
+
+        if (!shouldOverwrite.overwrite) {
+          return;
+        }
+      }
+
+      fs.unlinkSync(resolvedFilePath);
```
stasmaxymov: There is a chance of loosing file if shadcn command fails due to some other reasons, e.g. dependency installation. The failsafe way to do it would be to move file to temp folder or rename it in place and then either remove or restore depending of shadcn results

stasmaxymov (commented):
@illiakovalenko  thanks for the changes, there are few comments, please have a look.

packages/cli/src/scripts/project/component/add.test.ts
illiakovalenko: @stasmaxymov It’s a slightly different case - here we need to stub methods from a third-party npm package. Regular stubbing or replacement doesn’t work well in this scenario, so we need to use proxyquire or apply a mocking workaround

packages/cli/src/scripts/project/component/add.ts
illiakovalenko: Good catch, totally right

packages/cli/src/scripts/project/component/add.test.ts
stasmaxymov: Interesting, I see there is a stub for inquirer and it seems it works:
`inquirerStub = sandbox.stub(inquirer, 'prompt');`
Is it due to the way how core package is built?

stasmaxymov (approved):

packages/cli/src/scripts/project/component/add.test.ts
illiakovalenko: @stasmaxymov I don't remember exact details, but most likely because of commonjs and monorepo setup

PR #276, opened 2025-11-03
[chore] Harmonize react Placeholder unit tests

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Updates BYOC and Placeholder unit tests in react package:
- Ensure AppPlaceholder is tested in server context by default
- Disabled BYOC and FEAAS tests re-enabled
- Ensure tests run okay and assert in a sane manner

## Testing Details
- [x] Unit Tests Updated (duh)

## Types of changes
- [x] Chore


packages/react/src/components/FEaaS/BYOCWrapper.test.tsx
```diff
@@ -257,19 +257,20 @@ describe('Error handling', () => {
...
       params: { ComponentName: 'NonExistentComponent' },
       components: {},
       fetchedData: {},
     };
 
     const wrapper = render(<BYOCComponent {...props} />);
-
-    expect(wrapper.queryAllByText('This component was not registered').length).to.equal(1);
+    waitFor(() => {
+      expect(wrapper.queryAllByText('This component was not registered').length).to.equal(1);
```
illiakovalenko: this assertion isn’t being executed - I can change its value to anything and the test still passes.
Most likely you forgot to add `await` before `waitFor`

packages/react/src/components/Placeholder/models.ts
```diff
@@ -24,6 +24,7 @@ export interface AppComponentProps {
     [name: string]: string;
   };
   rendering: ComponentRendering;
+  disableSuspense?: boolean;
```
illiakovalenko: Not sure I follow why our components should get a `disableSuspense` prop - how is this supposed to be used?

packages/react/src/components/FEaaS/BYOCWrapper.test.tsx
```diff
@@ -283,9 +284,11 @@ describe('Error handling', () => {
       fetchedData: {},
     };
     const wrapper = render(<BYOCComponent {...props} />);
-    const text = wrapper.container.querySelector('div')?.innerText;
-
-    expect(text).to.contain('Custom missive for NonExistentComponent');
-    expect(text).to.contain('This component was not registered');
+    waitFor(() => {
```
illiakovalenko: Same as above

packages/react/src/components/Placeholder/AppPlaceholder.test.tsx
```diff
@@ -557,7 +559,7 @@ describe('App Placeholder logic', () => {
```
illiakovalenko: Remove unused variables: https://github.com/Sitecore/content-sdk/pull/276/files#diff-b2e8fcc007988592cdf84412dfd39c0534c2510e7e5485c30125b7129935b1f8R557-R558

packages/react/src/components/Placeholder/placeholder-utils.tsx
```diff
@@ -155,6 +155,7 @@ export function getAppComponentProps<T extends BasePlaceholderProps>(
       ...getSXAParams(componentRendering),
     },
     rendering: componentRendering,
+    disableSuspense: placeholderProps.disableSuspense,
```
illiakovalenko: The same question as for "models" change

illiakovalenko (changes requested):

packages/react/src/components/Placeholder/models.ts
art-alexeyenko: We passed *all* props to components with old Placeholder implementation, but we filter them with AppPlaceholder implementation,  as it needs to be safe while not passing non-serializable props.
disableSuspense is serializable, but was cut from propagation somewhere along the line, and was not passed into child placeholders.

packages/react/src/components/Placeholder/AppPlaceholder.test.tsx
art-alexeyenko: They are used down the file:
image

packages/react/src/components/Placeholder/placeholder-utils.tsx
art-alexeyenko: and the same answer :)

packages/react/src/components/FEaaS/BYOCWrapper.test.tsx
art-alexeyenko: So the miracle did not happen, a shame. I was hoping react test tools worked okay with feaas now.
I reworked these two unit tests to focus more on our logic, than FEAAS one.

packages/react/src/components/FEaaS/BYOCWrapper.test.tsx
art-alexeyenko: I'll mark this resolved as it's the same solution as above

packages/react/src/components/Placeholder/AppPlaceholder.test.tsx
illiakovalenko: @art-alexeyenko nope, you can just remove them, since they are not used in any other places, you can stub without defining a variable

packages/react/src/components/Placeholder/models.ts
illiakovalenko: @art-alexeyenko This isn’t the correct approach to pass all props to components as it was done in the old Placeholder. Moving forward - possibly in the next major release - we plan to remove these extra properties.
From my understanding, no developers will be accessing `disableSuspense` inside the component itself

illiakovalenko (approved):

PR #279, opened 2025-11-10
[core][nextjs]: Add comments/fallbacks for multisite config in App Router

MenKNas:
- [ ] Changelog updated

## Description / Motivation
After investigating, my conclusion is that **multisite** cannot be disabled in App Router due to route structure constraints ([site] segment required). I have documented the limitations and the correct approach for single-site setups.

Changes:
- Added `MULTISITE_CONFIGURATION.md` file that explains the current setup and how to work with it
- Added warning comment in config
- Added fallback for missing site cookie in Preview/Editing modes
- Added runtime warning when disabled in App Router

_Key finding: For single-site setups, keep multisite.enabled: true and configure one site — achieves the same result without breaking the app._

## Testing Details
- [ ] Unit Test Added

## Types of changes

### DIFF
--- a/packages/nextjs/src/middleware/multisite-middleware.ts
+++ b/packages/nextjs/src/middleware/multisite-middleware.ts
@@ -66,8 +66,11 @@ export class MultisiteMiddleware extends MiddlewareBase {
       if (!isSitecorePreview) {
         if (!this.config.enabled) {
           this.shouldWarnWhenDisabled(res);
-          debug.multisite('skipped (multisite middleware is disabled globally)');
-          return res;
+          if (this.shouldSkipWhenDisabled()) {
+            debug.multisite('skipped (multisite middleware is disabled globally)');
+            return res;
+          }
+          // Continue execution if shouldSkipWhenDisabled returns false (App Router case)
         }
 
         if (this.disabled(req, res)) {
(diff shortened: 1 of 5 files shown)


packages/nextjs/src/middleware/multisite-middleware.ts
```diff
@@ -79,7 +89,16 @@ export class MultisiteMiddleware extends MiddlewareBase {
...
+        // Fallback to hostname resolution if cookie is missing (shouldn't happen in normal flow)
+        const siteCookie = req.cookies.get(SITE_KEY)?.value;
+        if (!siteCookie) {
+          debug.multisite(
+            'warning: sc_site cookie missing in preview mode, falling back to hostname resolution'
+          );
+          siteName = this.siteResolver.getByHost(hostname).name;
+        } else {
+          siteName = siteCookie;
+        }
```
yavorsk: Do we need these changes? I think resolve by host won't work for sitecore preview

packages/nextjs/src/middleware/multisite-middleware.ts
```diff
@@ -63,6 +65,14 @@ export class MultisiteMiddleware extends MiddlewareBase {
 
       if (!isSitecorePreview) {
         if (!this.config.enabled) {
+          // Warn if multisite is disabled in App Router - this will break regular requests
```
yavorsk: potentially this check can be extracted as a method and which has override in app-router-multisite-middleware, instead of using `this.isAppRouter`. might be better

ref-docs/nextjs/middleware/MULTISITE_CONFIGURATION.md
```diff
@@ -0,0 +1,90 @@
+# Multisite Configuration in App Router
```
yavorsk, line 1: explanation looks ok, but maybe check with product if it is better to have these notes in the doc site

yavorsk (approved):
Overall looks good. I am not sure about the additional md file, maybe check with product on that. approved in advance

PR #283, opened 2025-11-13
[Chore] API Surface verification and reporting

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Adds API surface verification to the Pull Request build workflow to catch unintended API changes before merge.
### Problem
Breaking API changes could slip through without explicit validation, risking compatibility issues for consumers.
### Solution
Adds an API Extractor verification step that:
* Validates the public API surface against committed API reports
* Fails the build if there are uncommitted API changes
* Ensures breaking changes are intentional and documented

Added "Verify API surface" step to `.github/workflows/pull_request_build.yml`:
* Runs `yarn api-extractor:verify` which executes API Extractor across all packages (except of CLI and create-content-sdk-app)
* Uses the existing api-extractor:verify script that validates against committed API reports

## Testing Details

- [ ] Unit Test Added

## Types of changes


packages/core/src/client/sitecore-client.ts
```diff
@@ -37,11 +38,13 @@ export enum ErrorPage {
 
 /**
  * Page mode name
+ * @internal
  */
 type PageModeName = LayoutServicePageState | DesignLibraryMode;
```
yavorsk, line 43: LayoutServicePageState, DesignLibraryMode and PageMode are all public, so shouldn't this be public as well?

yavorsk (approved):
Looks great, very useful addition. Approved, see only one comment.

PR #293, opened 2025-11-24
[core][nextjs] Client import map for App Router

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
As Content SDK continues leveraging App Router and RSC features, so does Design Library.
This PR adds a separation for server/client import-maps in order for Design Library to be able to render both client and server components.

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other - verified generation of import maps and building of the app

## Types of changes


packages/react/src/enhancers/withLoadImportMap.tsx
```diff
@@ -23,8 +23,14 @@ export function withLoadImportMap<T extends WithLoadImportMapProps>(
...
-
-    return <Component {...props} loadImportMap={props.loadImportMap || loadImportMapContext} />;
+    const loadClientImportMap = props.loadImportMap || loadImportMapContext;
+    return (
+      <Component
+        {...props}
+        loadClientImportMap={loadClientImportMap}
+        loadImportMap={loadClientImportMap}
+      />
+    );
```
illiakovalenko: We don't need to apply any of those changes, right?
We still need to pass only `loadImportMap` and we still need fallback to `props.loadImportMap`

packages/core/src/tools/codegen/import-map.ts
illiakovalenko, line 1: I see you’ve started moving some Next.js/React-specific logic into the nextjs package, which actually makes sense. Can we move more? instead of continuing to add more Next.js-specific logic here? Currently it's not extendable. At the moment we have a hardcoded Next.js template with React-specific logic. Even the client/server map separation is largely a Next.js-specific concern
@stasmaxymov What are your thoughts on this?

packages/nextjs/src/tools/codegen/import-map.ts
```diff
@@ -0,0 +1,18 @@
+import {
+  writeImportMap as writeImportMapCore,
+  WriteImportMapArgs,
+} from '@sitecore-content-sdk/core/tools';
+import { detectRouterType } from '../templating/utils';
+
+/**
+ * Entry point function for generating import-map. Parses provided paths and outputs the modules and imports from those files into .sitecore/import-map.ts
```
illiakovalenko, line 12: Let's add an explanation about client/server maps and differences in Pages vs App Router usage

packages/react/src/components/DesignLibrary/DesignLibraryApp.tsx
```diff
@@ -18,6 +18,7 @@ export const DesignLibraryApp = ({
   page,
   componentMap,
   loadServerImportMap,
+  loadClientImportMap,
```
illiakovalenko, line 21: From my understanding we don't need to add `loadClientImportMap` prop and pass it to `DesignLibrary`, since client import map is provided here in [Providers.tsx](https://github.com/Sitecore/content-sdk/blob/dev/packages/create-content-sdk-app/src/templates/nextjs-app-router%20(beta)/src/Providers.tsx#L13). Also you need to update Providers value since `import-map` file is not present anymore in App Router solution

illiakovalenko (changes requested):
Overall, it looks pretty good 👍
Please see a few general and design-related questions below

packages/core/src/tools/codegen/import-map.ts
art-alexeyenko: I was worried this could be out of scope for this feature and bloat it too much, but I've posted the commit with configurable map templates, worst case it can be reversed based on what @stasmaxymov says

packages/react/src/enhancers/withLoadImportMap.tsx
art-alexeyenko: yeah, you're right, this was a leftover from other approach I tried

packages/react/src/components/DesignLibrary/DesignLibraryApp.tsx
art-alexeyenko: HOC takes import map from context, but I also don't see any place where import map is posted to context. It should be fine to pass it this way, IMO?
@yavorsk what's the right approach based on your recent work?

packages/core/src/graphql-request-client.test.ts
```diff
@@ -17,7 +17,7 @@ use(spies);
 const nodeStatusCode = ['ECONNRESET', 'ETIMEDOUT', 'EPROTO'];
 const statusErrorCodes = [429, 502, 503, 504, 520, 521, 522, 523, 524];
 
-describe('GraphQLRequestClient', () => {
+xdescribe('GraphQLRequestClient', () => {
```
illiakovalenko:
```suggestion
describe('GraphQLRequestClient', () => {
```

packages/react/src/components/DesignLibrary/DesignLibraryApp.tsx
illiakovalenko: @art-alexeyenko As I mentioned in a previous comment it's passed [here](https://github.com/Sitecore/content-sdk/blob/dev/packages/create-content-sdk-app/src/templates/nextjs-app-router%20(beta)/src/Providers.tsx#L13), you need to update it to reference client import map, default import map is not present anymore. So `loadClientImportMap` prop is not needed

packages/core/src/tools/codegen/import-map.ts
illiakovalenko: @art-alexeyenko Yeah, I have the same opinion, I don't want to bloat it more and more. Using templates it's a good option for now. In this case do we want to move a [next.js default map template](https://github.com/Sitecore/content-sdk/blob/d91ff9afcf0fe9dca80e6c7958cc41a8783945f2/packages/core/src/tools/codegen/import-map.ts#L494-L522) as well? Looks like we can do that

illiakovalenko (approved):
Approved in advance 👍
Please see my new comments below. You can resolve them and proceed with the merge, or we can discuss further if anything remains unclear

packages/react/src/components/DesignLibrary/DesignLibraryApp.tsx
art-alexeyenko: I'd leave the loadClientImportMap prop just for simplicity. The current approach is kind of over-complicated with one server map passed directly, then client map applied in a separate place via context + HOC. At least the override should be easier to follow.

packages/core/src/tools/codegen/import-map.ts
art-alexeyenko: I ended up on a half measured approach here. _defaultMapTemplate contains some extra logic, not just the string template, that is applicable for all frameworks, in principle. Ideally we want to only pass the string template, not this extra logic from framework-specific packages. That requires a lot of moving things around.
With this current approach we just pass the framework name. It gets rid of nextjs-specific stuff in core, without moving extra things too much for this PR.

packages/react/src/components/DesignLibrary/DesignLibraryApp.tsx
illiakovalenko: @art-alexeyenko It’s not simplified - it has actually become more complex. Why should we support an additional property if the map is already available through the provider? This is exactly how it works in AppPlaceholder. Why should customers supply a client map here if it’s already provided by the Providers? Could you please simplify this? There is no reason to keep it and it's not logical

PR #295, opened 2025-11-25
Search integration

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
This PR introduces search functionality to the Sitecore Content SDK, including a new `@sitecore-content-sdk/search` package and React hooks (`useSearch` and `useInfiniteSearch`) for performing search queries with pagination, sorting, and request cancellation support.

### _@sitecore-content-sdk/search_

- **`SearchService` class**: Performs search queries
  - Supports pagination (`limit`, `offset`)
  - Supports sorting with nested field paths (dot-notation)
  ```ts
    type Model = {
      id: string;
      skill: {
        name: string;
        level: number;
      };
    };

    const service = new SearchService(...);    

    service.search<Model>({
      searchIndexId: 'your-index-id',
      ...
      sort: {
        name: 'skill.name', // <----- TypeScript autocomplete and validation
        order: 'asc'
      }
    })
  ```
  - Request cancellation via `AbortSignal`
  - Fully type-safe with generic field support

### @sitecore-content-sdk/react

- **`useSearch`**: Paginated search hook
  - Automatic state management (`isLoading`, `isSuccess`, `isError`, `results`, `total`, `totalPages`, ...)
  - Automatic search execution on dependency changes
  - Request cancellation on unmount and when new searches are triggered
  - Race condition prevention

- **`useInfiniteSearch`**: Infinite scroll/search hook
  - All features of `useSearch` plus:
  - `loadMore()` function for pagination
  - `hasNextPage` flag
  - Separate loading states for initial load vs. load more (`isLoadingMore`, `isLoadingMoreError`, ...)
  - Preserves existing results on load more errors

Added package.json "exports" field to ensure correct ESM/CJS module resolution for submodule imports. Previously defaulted to CJS bundles even when ESM was available. This is especially needed in case of submodules defined in _react_ package, we can't reference different bundles when rendering React components, the same React instance should be referenced

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+    });
+  }
+
+  /**
+   * Search for items in the search index.
+   * @param {SearchParameters<T>} params - The search parameters.
+   * @param {RequestInit} [fetchOptions] - The fetch options.
+   * @returns {Promise<SearchResponse<T>>} The search response.
+   */
+  async search<T = GenericFields>(
```
illiakovalenko: @stasmaxymov At the very end I don't extend "GenericFields" anymore. I need to define "GenericFields" as indexed type to handle circular dependencies, so I wouldn't prefer to force developers to define an indexed type as well

stasmaxymov: But without constraint you can actually pass not valid as a document, e.g. this one compiles without errors:
`searchService.search<'I' | 'am' | 'not' | 'a' | 'valid' | 'object'>({
  searchIndexId: '1234567890',
  keyphrase: 'test',
}); `
besides I don't see what problems you have with indexed types, when I add extends GenericFields constaint, this compiles fine:
```
searchService.search<{
  I: string;
  am: string;
  a: string;
  valid: string;
  object: string;
}>({
  searchIndexId: '1234567890',
  keyphrase: 'test',
});
```

packages/search/src/models.ts
```diff
@@ -0,0 +1,25 @@
+export type PrimitiveType = string | number | boolean;
+
+/**
+ * Generic Search fields type.
+ * @public
+ */
+export type GenericFields = {
```
stasmaxymov: I think name is a bit confusing for API users, like what generic, what fields. Lets re-exports this type as something domain specific like `SearchDocument`, `DocumentModel` etc.

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+    this.fetcher = new NativeDataFetcher({
+      debugger: debug.search,
+    });
+  }
+
+  /**
+   * Search for items in the search index.
+   * @param {SearchParameters<T>} params - The search parameters.
+   * @param {RequestInit} [fetchOptions] - The fetch options.
+   * @returns {Promise<SearchResponse<T>>} The search response.
```
stasmaxymov, line 113: Taking lessons from Cloud SDK search codebase we can add `@throws` ts-doc entries for all validation checks would be useful for end user

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+    this.validateParameters<T>({
+      searchIndexId,
+      keyphrase,
+      sort,
+      limit,
+      offset,
+    });
+
+    const { data } = await this.fetcher.post<SearchAPIResponse<T>>(
+      `${this.config.edgeUrl}/v1/search?sitecoreContextId=${this.config.contextId}`,
```
stasmaxymov: I'm proponent of using a managed way of building an url using URL class. From my experience, string interpolation lead to weird behaviors or bugs. E.g. imagine this.config.edgeUrl is not a valid host or right away empty - you will get an unreasonable error on `fetch`

packages/react/src/search/utils.ts
```diff
@@ -0,0 +1,35 @@
...
+  return useMemo(() => {
+    if (!api?.edge?.clientContextId) {
+      return null;
+    }
+
+    return new SearchService({
+      contextId: api.edge.clientContextId,
+      edgeUrl: api.edge.edgeUrl,
+    });
+  }, [api.edge.clientContextId, api.edge.edgeUrl]);
```
stasmaxymov: here you have dependency as `api.edge.clientContextId`
but above you use optional chaining: `api?.edge?.clientContextId`
if api or edge could legitimately be nullish - there might be "undefined exception"

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+  /**
+   * Search for items in the search index.
+   * @param {SearchParameters<T>} params - The search parameters.
+   * @param {RequestInit} [fetchOptions] - The fetch options.
+   * @returns {Promise<SearchResponse<T>>} The search response.
+   */
+  async search<T = GenericFields>(
+    params: SearchParameters<T>,
+    fetchOptions?: RequestInit
+  ): Promise<SearchResponse<T>> {
```
stasmaxymov, line 124: we could use optional abort controller parameter and pass it to fetcher, that's overall a good practice

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+        error: errorMessage,
+      });
+    }
+  }, [searchService, searchIndexId, page, pageSize, sort, query]);
+
+  useEffect(() => {
+    search();
+  }, [search]);
+
+  useEffect(() => {
```
stasmaxymov: what is the point of separating unsubscribe logic into separate effect? It can lead to a faulty behavior as abort controller will be triggered only on component unmount, but effect with search will be executed each time search parameter is changed. So current abort controller doesn't prevent race conditions and "stale state setting".
Move abortion to the main `useEffect`. You also can make search function to accept abort controller - this way you can get rid of `useRef`

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+export const useSearch = <T = GenericFields>(
+  options: UseSearchOptions<T>
+): UseSearchState<T> => {
+  const { query, page = DEFAULT_PAGE, searchIndexId, pageSize = DEFAULT_PAGE_SIZE, sort } = options;
+
+  if (!searchIndexId) {
+    throw new Error('useSearch: searchIndexId is required');
+  }
+
+  const [state, setState] = useState<{
```
stasmaxymov: You probably can use `UseSearchState` type here instead of redeclaring it directly in useState

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+    results: T[];
+    isLoading: boolean;
+    isSuccess: boolean;
+    isError: boolean;
+    total: number;
+    totalPages: number;
+    error: Error | null;
+  }>({
+    results: [],
+    isLoading: true,
```
stasmaxymov: It is technically incorrect  to set isLoading to true before request actually started,

stasmaxymov: Initially hook is in kind of "idle" state

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+  error: Error | null;
+}
+
+/**
+ * React hook for performing search queries with pagination.
+ * @param {UseSearchOptions} options - Configuration options for the search hook.
+ * @returns {UseSearchState} The search state.
+ * @public
+ */
+export const useSearch = <T = GenericFields>(
```
stasmaxymov: Currently hook fires immediately and there is no way to call conditionally, e.g. in button click.
add `enabled` option that controls if search request is executed (default is true)

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+  }, [search]);
+
+  useEffect(() => {
+    return () => {
+      // Abort all requests when component unmounts
+      abortControllerRef.current?.abort();
+    };
+  }, []);
+
+  return useMemo(
```
stasmaxymov, line 220: Currently this useMemo just returns all the same properties you have in useState, that is excessive, as state is referentially stable. Memo might be useful here if you do any transformation

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+
+      const searchParams: SearchParameters<T> = {
+        searchIndexId,
+        keyphrase: query,
+        limit: pageSize,
+        offset,
+        sort,
+      };
+
+      const { results: searchResults, total } = await searchService.search<T>(searchParams, {
```
stasmaxymov, line 174: according useSearchService implementation, it can return null, but there is no null check here.

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+      if (signal.aborted) {
+        return;
+      }
+
+      const errorMessage = err instanceof Error ? err : new Error('Search failed');
+      setState({
+        results: [],
+        total: 0,
+        totalPages: 0,
+        isSuccess: false,
```
stasmaxymov: It is kind of tedious and error-prone to work with 4 "is" flags (4 because you will need `isIdle` too) all the time. 
It's better to have a single variable `state` or 'status' as enum of all states, and infer "is" flags from it, e.g. in a memo down below. You can check Tanstack Query for reference: https://tanstack.com/query/v3/docs/framework/react/reference/useQuery

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+        isLoading: false,
+        error: null,
+      });
+    } catch (err) {
+      // Don't set error if request was aborted
+      if (signal.aborted) {
+        return;
+      }
+
+      const errorMessage = err instanceof Error ? err : new Error('Search failed');
```
stasmaxymov: If `err` is not instance of `Error`, you kind of loose this info, maybe we can json stringify, it's content  
`new Error(`Search failed: ${JSON.stringify(err)}`);`

packages/react/src/search/useInfiniteSearch.ts
```diff
@@ -0,0 +1,263 @@
...
+  error: Error | null;
+}
+
+/**
+ * React hook for performing infinite search queries.
+ * @param {UseInfiniteSearchOptions} options - Configuration options for the infinite search hook.
+ * @returns {UseInfiniteSearchState} The infinite search state.
+ * @public
+ */
+export const useInfiniteSearch = <T = GenericFields>(
```
stasmaxymov: some of the comments from useSearch are also applicable here, please have a look

stasmaxymov (changes requested):
Great work! Please see some comments

packages/search/src/search-service.ts
illiakovalenko: That’s a good point. We typically use string interpolation, but using the URL class is a safer approach

packages/react/src/search/utils.ts
illiakovalenko: In general it should not be possible but let's say component is used outside of SitecoreProvider or may be there is some problem with the initial value or we are in local containers :).
So in any case it makes sense to safely check

packages/search/src/search-service.ts
illiakovalenko: @stasmaxymov It's a part of "fetchOptions", it's accessed via "signal" property. Also we already pass [signal](https://developer.mozilla.org/en-US/docs/Web/API/RequestInit#signal) to "SearchService" from our hooks

packages/react/src/search/useSearch.ts
illiakovalenko: @stasmaxymov No, it can't return null, if search results are empty we will get an empty array

packages/react/src/search/useSearch.ts
illiakovalenko: Let's talk about this one more

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+ * Service that fetches search results from Sitecore.
+ * @public
+ */
+export class SearchService {
+  private fetcher: NativeDataFetcher;
+
+  constructor(private config: SearchServiceConfig) {
+    this.config.edgeUrl = this.config.edgeUrl || constants.SITECORE_EDGE_URL_DEFAULT;
+
+    this.fetcher = new NativeDataFetcher({
```
sc-nikolaoslazaridis, line 104: I would create the NativeDataFetcher instance directly in line 93 since we don't have any dependency injection in the constructor for this case.

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+      {
+        config: {
+          id: searchIndexId,
+        },
+        limit,
+        offset,
+        query: {
+          keyphrase,
+        },
+        facet: {
```
sc-nikolaoslazaridis: Should this object always included in the request? If yes maybe we should externalize this property to the devs so they can configure what to include in the array.

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+
+/**
+ * Service that fetches search results from Sitecore.
+ * @public
+ */
+export class SearchService {
+  private fetcher: NativeDataFetcher;
+
+  constructor(private config: SearchServiceConfig) {
+    this.config.edgeUrl = this.config.edgeUrl || constants.SITECORE_EDGE_URL_DEFAULT;
```
sc-nikolaoslazaridis, line 102: I would try to check if the provided url is not malformed before proceeding the execution. Just to catch it sooner than later.  (maybe with the URL class or a regex, up to you)

packages/search/src/search-service.ts
```diff
@@ -0,0 +1,168 @@
...
+        query: {
+          keyphrase,
+        },
+        facet: {
+          fields: [],
+          all: true,
+        },
+        sort,
+      },
+      fetchOptions
```
sc-nikolaoslazaridis, line 155: Inside the fetcher we spread the fetch options after the body we constructed in line 125. Is this intended? meaning if the dev provides a "custom" body we lose the backend schema. Maybe omit the body prop from the RequestInit or enforce our schema if they provide it.

packages/react/src/search/useSearch.ts
illiakovalenko: yeah, it's reworked now using "status"

packages/react/src/search/useSearch.ts
illiakovalenko: Now fixed by removing "strictNullCheck: false" :)

packages/search/src/search-service.ts
illiakovalenko: There are no differences overall, right? Just depends how you prefer

packages/search/src/search-service.ts
illiakovalenko: No, currently facets are not supported, it's hardcoded on our side

packages/search/src/search-service.ts
illiakovalenko: Done like it was mentioned by Stas above, using URL

packages/search/src/search-service.ts
illiakovalenko: @sc-nikolaoslazaridis Not sure if I understand what you mean. `fetchOptions` are RequestInit parameters that can be passed to our `NativeFetcher` - `post` method

packages/react/src/search/useSearch.ts
illiakovalenko: Added `enabled` property

packages/search/src/search-service.ts
sc-nikolaoslazaridis: Yes, the search backend expects some kind of structure for the body. By accepting the full RequestInit we allow the user to pass any kind of body. That will result in incorrect responses from the backend.

packages/search/src/search-service.ts
sc-nikolaoslazaridis: So if they are not supported why we should add them to the request? Can't we omit it?

packages/react/src/search/useSearch.ts
```diff
@@ -0,0 +1,196 @@
...
+ * @param {UseSearchOptions} options - Configuration options for the search hook.
+ * @returns {UseSearchState} The search state.
+ * @public
+ */
+export const useSearch = <T = GenericFields>(
+  options: UseSearchOptions<T>
+): UseSearchState<T> => {
+  const { query, page = DEFAULT_PAGE, searchIndexId, pageSize = DEFAULT_PAGE_SIZE, sort } = options;
+
+  if (!searchIndexId) {
```
sc-nikolaoslazaridis: Altho is up to the developer to provide the correct configuration, throwing an error here will potentially bring down the whole page. I think we can simply check the value and adapt the initial state of the useState at line 85.

packages/search/src/search-service.ts
illiakovalenko: It doesn't affect, but it's possible

packages/react/src/search/useSearch.ts
illiakovalenko: It's totally expected, application should fail fast if something is configured in a wrong way, in production you should not have some unexpected issues because it was suppressed by sdk. This is how we approach in content-sdk

packages/search/src/search-service.ts
illiakovalenko: I see. I've decided to limit available fetch options, should be better

packages/react/src/search/useSearch.ts
illiakovalenko: Upd: I left "strictNullCheck" it affects a lot of places in our codebase, I will create a separate item to work on it

packages/react/src/search/useSearch.ts
sc-nikolaoslazaridis: If the hook is used outside a error boundary (Placeholder) the issue is still there and you can't wrap it in try catch blocks (ofc you can have another abstraction with error boundaries).

illiakovalenko: it should just fail if you develop component and you didn't pass index, why do we need to catch this error and hide?

illiakovalenko: As we discussed, this value can be changed during runtime, so definitely we need to handle it more gracefully

sc-nikolaoslazaridis (commented):
Besides any other "features" that can be added in the future if required i would still consider a debounce logic for when the search term changes. Example: if the user types into the search bar this solution will potentially do a request per keypress. Usually this is not something that is appealing especially in the case the whole "state" changes back to an empty array for the result. 

If considering to implement this we could accept a debounceMs parameter and if set have the request performed/canceled based on a timeout.

illiakovalenko:
> Besides any other "features" that can be added in the future if required i would still consider a debounce logic for when the search term changes. Example: if the user types into the search bar this solution will potentially do a request per keypress. Usually this is not something that is appealing especially in the case the whole "state" changes back to an empty array for the result.
> 
> If considering to implement this we could accept a debounceMs parameter and if set have the request performed/canceled based on a timeout.

Initially I planned to introduce debounce support, but then it was removed, this is related to "components/ui" stuff and should be handled there if needed. Developers can use their own libs to simplify that

stasmaxymov (approved):

PR #297, opened 2025-11-26
[create-content-sdk][nextjs]: Fixes for middleware errors in local containers development

MenKNas:
- [x] Changelog updated

## Description / Motivation
This PR resolves `"GraphQL client misconfigured"` error in local container development by updating `RedirectsMiddleware` and `PersonalizeMiddleware` to pass both edge and local API configs to the GraphQL client factory, enabling fallback to local API when Edge context IDs aren't configured.

## Testing Details
- [ ] Unit Test Added

## Types of changes


packages/nextjs/src/middleware/personalize-middleware.ts
```diff
@@ -27,6 +27,7 @@ export type PersonalizeGeoData = {
  */
 export type PersonalizeMiddlewareConfig = MiddlewareBaseConfig &
   SitecoreConfig['api']['edge'] &
+  SitecoreConfig['api']['local'] &
```
art-alexeyenko: Personalize cannot function with local deployments - can we instead do a validation of edge config here and set personalizeService to null if edge config is absent?
The way we pass only the edge config into some middleware was intentional - it demostrated when middleware works and when it isn't, so we can add what was missing and just disable some middleware when in local deployment scenario.

art-alexeyenko (approved):
Looks good! thank you

PR #299, opened 2025-11-26
[nextjs][core] QoL follow-up for client import map

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Some QoL follow-ups:
- remove `loadClientImportMap` prop from `DesignLibraryApp` to avoid duplicate configuration between context and component itself
- Remove the client/server import map templates and `separateServerClientMaps` flag props from being configurable in CLI config. This streamlines and makes the import map flow more robust: pages router apps always need one map, app router apps always need separate maps to function fully, and this change prevents app breaking accidentally when modifying config.
- Add a default empty `import-map.client.ts` to App Router sample

## Testing Details
- [ ] Unit Test Added

## Types of changes


yavorsk (approved):
Looks good!

PR #280, opened 2025-11-11
[core] [nextjs] [create-content-sdk-app/nextjs-app-router] DesignLibrary for server side components

yavorsk:
- [x] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)
- [x] Changelog updated
- [ ] Upgrade guide entry added

## Description / Motivation

This PR adds support for Design Library server components for the nextjs app router template.
 - Introduce DesignLibraryServer and DesignLibraryClientEvents components to support design library when editing server components
 - introduce DesignLinbraryApp component to serve as router between server and client component rendering
 - Introduce updateServerComponentAction server function to store the updates received from Design Library
 - introduce globalCache service in core package to store and retrieve data from the globalThis object, used by the server function
 - add POST handler to editing render route handler which proxies the POST request for the server function and triggers rerender of the server component when it needs to be updated
 - includes various refactorings to extract common functionality 
 - extract common DLErrorBoundary component for Design Library error handling
 - move all DesignLibrary components to separate folder in react package
 - add loadImportMap context and introduce withLoadImportMap to be used to load the import map for the client side DesignLibrary component

## Testing Details

- [x] Unit Test Added - unit tests are WIP
- [x] Manual Test/Other (Please elaborate)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [x] Breaking change (fix or feature that would cause existing functionality to change)

PR #301, opened 2025-11-28
[nextjs] Follow up to DesignLibrary for server side components - allow internal POST requests in editing handler

yavorsk:
- [x] Changelog updated

## Description / Motivation

This is a follow up to https://github.com/Sitecore/content-sdk/pull/280 and addresses the issue where same origin and internal xmc editing host requests are blocked by CORS check

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/nextjs/src/route-handler/editing-render-route-handler.ts
+++ b/packages/nextjs/src/route-handler/editing-render-route-handler.ts
@@ -267,9 +267,18 @@ export const createEditingRenderRouteHandlers = (options: EditingHandlerOptions)
    * @param {NextRequest} req - The incoming request
    */
   const POST = async (req: NextRequest) => {
+    const requestOrigin = req.headers.get('origin') || '';
+    const originHost = new URL(requestOrigin).host;
     const expectedCorsHeaders = getCorsHeaders(req);
-    if (!expectedCorsHeaders) {
-      return new Response(getOriginNotAllowedMessage(req.headers.get('origin') || ''), {
+
+    // Allow request if:
+    // we are in local or sitecore environment - requested host is 'localhost:3000'
+    // or the request is same origin (e.g. in vercel netlify environment)
+    const isAllowedRequest =
+      req.nextUrl.host === 'localhost:3000' || req.nextUrl.host === originHost;
+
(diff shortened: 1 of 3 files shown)


packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -267,9 +267,18 @@ export const createEditingRenderRouteHandlers = (options: EditingHandlerOptions)
...
+    const originHost = new URL(requestOrigin).host;
     const expectedCorsHeaders = getCorsHeaders(req);
-    if (!expectedCorsHeaders) {
-      return new Response(getOriginNotAllowedMessage(req.headers.get('origin') || ''), {
+
+    // Allow request if:
+    // we are in local or sitecore environment - requested host is 'localhost:3000'
+    // or the request is same origin (e.g. in vercel netlify environment)
+    const isAllowedRequest =
+      req.nextUrl.host === 'localhost:3000' || req.nextUrl.host === originHost;
```
art-alexeyenko: we should not rely on specific port number, would it make sense to only check for localhost, or `127.0.0.1` here?

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -267,9 +267,18 @@ export const createEditingRenderRouteHandlers = (options: EditingHandlerOptions)
...
+
+    if (!isAllowedRequest && !expectedCorsHeaders) {
```
art-alexeyenko: Should there be an 'OR' operation here? If not, let's rename `isAllowedRequest` to something like `shouldBypassCors` for better readability

art-alexeyenko (approved):
approving in advance - just suggesting some minor changes

packages/nextjs/src/route-handler/editing-render-route-handler.ts
yavorsk: in case of xmc env host  is hardcoded to 'localhost:3000'. but you're right, in case of local env that check might not be valid. i will update to check hostname

PR #304, opened 2025-12-01
[core] [react] Remove "strictNullChecks: false" tsconfig property

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
- Removed the strictNullChecks option from the tsconfig
- Fixed all errors related to the removed option

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/core/src/layout/utils.test.ts
```diff
@@ -68,6 +68,20 @@ describe('core layout utils', () => {
...
+      expect(result).to.be.true;
+    });
+
+    it('should return true if field is null', () => {
+      const field = null;
+      const result = isFieldValueEmpty(field);
+
+      expect(result).to.be.true;
+    });
+
```
illiakovalenko: We already have unit tests covering these scenarios, please, double check

packages/react/src/components/DesignLibrary/DesignLibrary.tsx
```diff
@@ -135,11 +139,13 @@ export const DesignLibrary = withLoadImportMap(({ loadImportMap }: DesignLibrary
         setRenderKey((k) => k + 1);
       });
 
-      const importMapEvent = getDesignLibraryImportMapEvent(rendering.uid, importMap);
+      const importMapEvent = getDesignLibraryImportMapEvent(uid, importMap);
       postToDesignLibrary(importMapEvent);
 
+      if (!propsState.fields || !propsState.params) return;
```
illiakovalenko: Let's don't skip _componentProps_ even when fields or params are undefined. Instead, in [getDesignLibraryComponentPropsEvent](https://github.com/Sitecore/content-sdk/blob/chore/JSS-8592/packages/core/src/editing/codegen/preview.ts#L396-L397) you can make `fields` and `params` optional and set default parameters value as `{}`

packages/react/src/components/DesignLibrary/DesignLibraryClientEvents.tsx
```diff
@@ -70,31 +73,33 @@ export const DesignLibraryVariantGenerationEvents = ({
...
-      const importMapEvent = getDesignLibraryImportMapEvent(component.uid, importMap);
+      const importMapEvent = getDesignLibraryImportMapEvent(component.uid, importMap!);
       postToDesignLibrary(importMapEvent);
 
       const propsEvent = getDesignLibraryComponentPropsEvent(
         component.uid,
-        component.fields,
-        component.params
+        component.fields || {},
+        component.params || {}
```
illiakovalenko: See my comment regarding `getDesignLibraryComponentPropsEvent` function, we can make these parameters optional there and apply default value

packages/react/src/components/ErrorBoundary.tsx
```diff
@@ -69,7 +78,7 @@ class ErrorBoundary extends React.Component<ErrorBoundaryProps> {
         } else {
```
illiakovalenko, line 84: Can we reuse `ErrorComponent` in "if" branch as well?

packages/react/src/components/ErrorBoundary.tsx
```diff
@@ -18,14 +18,23 @@ export type ErrorBoundaryProps = {
...
 
+/**
+ * Simple error component applying basic error styling.
+ * @param {string} message The error message to display.
+ * @returns A JSX element containing the error message in a div.
+ * @internal
+ */
+export const ErrorComponent = ({ message }: { message: string }) => {
+  return <div className="sc-content-sdk-placeholder-error">{message}</div>;
+};
```
illiakovalenko: Let's add a simple unit test to cover that it renders as expected

packages/react/src/components/FEaaS/BYOCWrapper.tsx
```diff
@@ -59,11 +59,9 @@ export class BYOCComponent extends React.Component<BYOCComponentProps> {
...
         errorOverride: 'BYOC: The ComponentName for this rendering is missing',
       };
-      return props.missingComponentComponent ? (
-        <this.props.missingComponentComponent {...noNameProps} />
-      ) : (
-        <MissingComponent {...noNameProps} />
-      );
+      const MissingComp = this.props.missingComponentComponent;
+
+      return MissingComp ? <MissingComp {...noNameProps} /> : <MissingComponent {...noNameProps} />;
```
illiakovalenko, line 65: Do we need to apply these changes? I'm asking because in the same component we have two extra similar cases

packages/react/src/components/FEaaS/FEaaSWrapper.tsx
```diff
@@ -35,10 +35,10 @@ export const FEaaSComponent = (props: FEaaSComponentProps): JSX.Element => {
...
       data={data}
       template={props.template}
-      cdn={props.params?.ComponentHostName}
-      library={props.params?.LibraryId}
-      version={props.params?.ComponentVersion}
-      component={props.params?.ComponentId}
+      cdn={props.params?.ComponentHostName as string}
+      library={props.params?.LibraryId as string}
+      version={props.params?.ComponentVersion as string}
+      component={props.params?.ComponentId as string}
```
illiakovalenko: Type cast is not really needed here, in case of `library` and `component` we can just fallback to empty string. Also update the existing [unit test](https://github.com/Sitecore/content-sdk/blob/45c3ea7cdadb778e1bf283801dc56895670158d5/packages/react/src/components/FEaaS/FEaaSWrapper.test.tsx#L60)
```suggestion
      cdn={props.params?.ComponentHostName}
      library={props.params?.LibraryId ?? ''}
      version={props.params?.ComponentVersion}
      component={props.params?.ComponentId ?? ''}
```

packages/react/src/components/FEaaS/feaas-utils.ts
```diff
@@ -110,12 +122,13 @@ async function fetchData(dataOptions: FEAAS.DataOptions): Promise<FEAAS.DataScop
...
  */
 export const composeComponentEndpoint = (
-  params: FEaaSComponentParams,
+  params: FEaaSComponentParams & RequiredFEaaSParams,
   revisionFallback: RevisionType
 ) => {
   const revision = params.ComponentRevision || revisionFallback;
   const hostname = params.ComponentHostName.startsWith('https://')
     ? params.ComponentHostName
     : `https://${params.ComponentHostName}`;
```
illiakovalenko: For simplicity you can remove `RequiredFEaaSParams`, `isFEaaSComponentParamsComplete`, `ValidatedFEaaSComponentParams` things and just add optional chaining for `ComponentHostName`
```suggestion
  params: FEaaSComponentParams,
  revisionFallback: RevisionType
) => {
  const revision = params.ComponentRevision || revisionFallback;
  const hostname = params.ComponentHostName?.startsWith('https://')
    ? params.ComponentHostName
    : `https://${params.ComponentHostName}`;
```

packages/react/src/components/FEaaS/models.ts
illiakovalenko, line 1: See my comment in `feaas-utils`, all the types can be reverted for simplicity

packages/react/src/components/Placeholder/PlaceholderMetadata.tsx
illiakovalenko, line 1: To simplify all the changes you can just:
* Mark `id` as optional in `getCodeBlockAttributes`
* Mark `id` as optional in `renderComponent`

packages/react/src/components/Placeholder/models.ts
```diff
@@ -21,7 +21,7 @@ export interface AppComponentProps {
     [name: string]: Field | Item | Item[];
   };
   params: {
-    [name: string]: string;
+    [name: string]: string | undefined;
```
illiakovalenko: Let's don't update the interface, [here](https://github.com/Sitecore/content-sdk/blob/dev/packages/react/src/components/Placeholder/placeholder-utils.tsx#L91) we can return `{ styles: '' }`

packages/react/src/enhancers/withAppPlaceholder.test.tsx
```diff
@@ -96,6 +96,32 @@ describe('withAppPlaceholder HOC', () => {
   });
 
   // Basic functionality tests
+  it('should render without placeholders', () => {
+    // Create simple test data with only valid components
```
illiakovalenko: :)
```suggestion
```

illiakovalenko (changes requested):
Good cleanup! 👍
please take a look at my comments below

packages/react/src/components/FEaaS/BYOCWrapper.tsx
sc-nikolaoslazaridis: We get an error here if i wont rearrange the code. This is not a functional change.

packages/react/src/components/ErrorBoundary.tsx
sc-nikolaoslazaridis: I made the ErrorComponent accept either message or children and used it in the "if" also

packages/core/src/editing/codegen/preview.ts
```diff
@@ -393,8 +393,8 @@ export function getDesignLibraryComponentPreviewErrorEvent(
  */
 export function getDesignLibraryComponentPropsEvent(
   uid: string,
-  fields: ComponentFields,
-  parameters: ComponentParams
+  fields: ComponentFields = {},
+  parameters: ComponentParams = {}
```
illiakovalenko, line 397: To don't decrease a coverage let's add a simple unit test to test a scenario with empty "fields", "parameters"

packages/react/src/components/ErrorBoundary.tsx
```diff
@@ -18,14 +18,29 @@ export type ErrorBoundaryProps = {
...
 };
 
+/**
+ * Simple error component applying basic error styling.
+ * @param {object} props - Either with `message` (string) or with `children` (ReactNode), but not both.
+ */
+export const ErrorComponent = (
+  props:
+    | { message: React.ReactNode; children?: never }
+    | { children: React.ReactNode; message?: never }
```
illiakovalenko, line 28:
```suggestion
  props: { children?: React.ReactNode; message?: React.ReactNode; }
```

illiakovalenko (approved):
Approved in advance 👍
Please, look and resolve a few comments below then feel free to merge

packages/react/src/components/ErrorBoundary.tsx
sc-nikolaoslazaridis: I want to enforce it on typescript level, either use one or the other

Issue #307, opened 2025-12-03
App Router:Component variant name (FieldNames) appended to component name during lookup causes "unknown component" error


rikaweb:
## Describe the Bug

When using rendering variants with the `FieldNames` parameter in XM Cloud, the Content SDK (Next.js App Router) incorrectly appends the variant name in parentheses to the component name when looking up the component in the component map.

**Example:**
- Component name in XM Cloud: `HeroST`
- Rendering variant parameter: `FieldNames: "Right"`
- SDK looks for: `HeroST (Right)` ❌
- Should look for: `HeroST` ✅

This causes the error:

<img width="744" height="525" alt="Image" src="https://github.com/user-attachments/assets/3d97b923-44ba-4352-9b53-02d1729422c7" />

The variant value should be passed to the component via `params.FieldNames`, not appended to the component name during lookup.

**Affected variants observed:**
- `HeroST (Right)`
- `HeroST (Stacked)`

## Additional Evidence from React Profiler

The React Profiler confirms the data from XM Cloud is correct:
```
rendering: {
  componentName: "HeroST",  // ✅ Correct - just the component name
  dataSource: "/sitecore/content/sync/sync/Home/Data/HeroST/HeroST 2"
}
params: {
  DynamicPlaceholderId: "3",
  FieldNames: "Stacked"  // ✅ Correct - variant as parameter
}
```
The SDK is internally combining these to look for `HeroST (Stacked)` instead of using `componentName` directly.

## Screenshots

<img width="651" height="506" alt="Image" src="https://github.com/user-attachments/assets/f03e6ebc-e1b2-4bf8-b83b-d34f72a37967" />

## To Reproduce

1. Create a Next.js App Router project: `npx create-content-sdk-app` → select `nextjs-app-router (beta)`
2. Create a component `HeroST.tsx` in `src/components/`
3. In XM Cloud, configure a rendering with `FieldNames` variants (e.g., "Default", "Right", "Stacked")
4. Add the component to a page using a variant other than default
5. Run `npm run dev`
6. Component shows orange placeholder: "Content SDK component is missing React implementation"

## Expected Behavior

The SDK should:
1. Look up component using only `componentName` (e.g., `HeroST`)
2. Pass the variant via `params.FieldNames` to the component

### Possible Fix

**Pull Request:** [#308 - fix(react): fallback to default export when variant named export not found](https://github.com/Sitecore/content-sdk/pull/308)

### Root Cause

[...]

PR #308, opened 2025-12-03
fix(react): fallback to default export when variant named export not …

rikaweb:
# PR: fix(react): fallback to default export when variant named export not found

## Checkboxes to select

- [x] This PR follows the Contribution Guide
- [ ] Changelog updated

---

## Description / Motivation

When using SXA rendering variants in Next.js App Router with XM Cloud, components with `params.FieldNames` set (e.g., `HeroST (Right)`) fail to render with error:

```
Placeholder headless-main contains unknown component HeroST (Right). 
Ensure that a React component exists for it, and that it is registered in your component-map file.
```

### Root Cause

In `packages/react/src/components/Placeholder/placeholder-utils.tsx`, the component resolution logic looks for a **named export** matching `params.FieldNames` (e.g., `component.Right`). If not found, it throws an error instead of falling back to the default export.

**Before:**
```typescript
const renderedComponent =
  exportName && exportName !== DEFAULT_EXPORT_NAME
    ? ((component as ReactModule)[exportName] as ComponentType)  // ❌ No fallback
    : (component as ReactModule).default || ...
```

### Solution

Gracefully fallback to the default export when a named variant export is not found:

```typescript
const variantComponent =
  exportName && exportName !== DEFAULT_EXPORT_NAME
    ? ((component as ReactModule)[exportName] as ComponentType)
    : undefined;

const defaultComponent =
  (component as ReactModule).default ||
  (component as ReactModule).Default ||
  (component as ComponentType);

// Use variant if found, otherwise fallback to default export
const renderedComponent = variantComponent || defaultComponent;
```

This allows components to:
1. Optionally provide named exports for specific variants (e.g., `export const Right = ...`)
2. Handle all variants via `params.FieldNames` prop in the default export

---

## Testing Details

- Tested with `HeroST` component and `HeroST (Right)` variant in XM Cloud
- Component now renders correctly using default export
- Component receives `params.FieldNames` prop to handle variant logic
- Backwards compatible: existing named exports still work as before

### DIFF
--- a/packages/react/src/components/Placeholder/placeholder-utils.tsx
+++ b/packages/react/src/components/Placeholder/placeholder-utils.tsx
@@ -251,15 +251,22 @@ export const getComponentForRendering = (
   // Render SXA Rendering Variant if available
   const exportName = renderingDefinition.params?.FieldNames;
 
-  const renderedComponent =
+  // Try to find a named export matching the variant, otherwise fallback to default export
+  const variantComponent =
     exportName && exportName !== DEFAULT_EXPORT_NAME
       ? ((component as ReactModule)[exportName] as ComponentType)
-      : (component as ReactModule).default ||
-        (component as ReactModule).Default ||
-        (component as ComponentType);
+      : undefined;
+
+  const defaultComponent =
+    (component as ReactModule).default ||
+    (component as ReactModule).Default ||
+    (component as ComponentType);
+
+  // Use variant if found, otherwise fallback to default export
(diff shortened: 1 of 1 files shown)


art-alexeyenko:
As commented in https://github.com/Sitecore/content-sdk/issues/307 , for an out of box default behavior, we'd like to not fallback to Default variant when a custom variant is not found, as this may obscure the issues with an app that should be fixed.
There is an approach to make custom variants fall back to Default when needed in the current implementation of Content SDK.

PR #319, opened 2025-12-12
fix: Fix React Server Components CVE vulnerabilities

malewis5:
Updated dependencies to fix Next.js and React CVE vulnerabilities.

The fix-react2shell-next tool automatically updated the following packages to their secure versions:
- next
- react-server-dom-webpack
- react-server-dom-parcel
- react-server-dom-turbopack

All package.json files have been scanned and vulnerable versions have been patched to the correct fixed versions based on the official React advisory.


art-alexeyenko:
@malewis5 thanks for the contribution! We're in the process of applying this very change (https://github.com/Sitecore/content-sdk/pull/322) and will release an update later today.

PR #325, opened 2025-12-15
[create-content-sdk-app][nextjs] Fix middleware initialization errors when API configuration is missing

MenKNas:
- [x] Changelog updated

## Description / Motivation
The `RedirectsMiddleware` constructor was throwing errors when no Edge or local API configuration was available, causing app startup failures. Updated RedirectsMiddleware to gracefully handle missing API configuration by setting the service to null and skipping processing when configuration is unavailable, similar to how `PersonalizeMiddleware` already handles this scenario. Both Pages Router and App Router templates were updated to always instantiate all middleware, with each middleware responsible for handling missing configuration internally.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/create-content-sdk-app/src/templates/nextjs-app-router/src/middleware.ts
+++ b/packages/create-content-sdk-app/src/templates/nextjs-app-router/src/middleware.ts
@@ -10,75 +10,78 @@ import sites from '.sitecore/sites.json';
 import scConfig from 'sitecore.config';
 import { routing } from './i18n/routing';
 
-const locale = new LocaleMiddleware({
-  /**
-   * List of sites for site resolver to work with
-   */
-  sites,
-  /**
-   * List of all supported locales configured in routing.ts
-   */
-  locales: routing.locales.slice(),
-  // This function determines if the middleware should be turned off on per-request basis.
-  // Certain paths are ignored by default (e.g. files and Next.js API routes), but you may wish to disable more.
-  // This is an important performance consideration since Next.js Edge middleware runs on every request.
-  // in multilanguage scenarios, we need locale middleware to always run first to ensure locale is set and used correctly by the rest of the middlewares
-  skip: () => false,
-});
-
-const multisite = new AppRouterMultisiteMiddleware({
-  /**
-   * List of sites for site resolver to work with
-   */
-  sites,
-  ...scConfig.multisite,
-  // This function determines if the middleware should be turned off on per-request basis.
-  // Certain paths are ignored by default (e.g. files and Next.js API routes), but you may wish to disable more.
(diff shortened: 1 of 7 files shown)


packages/create-content-sdk-app/src/templates/nextjs-app-router/src/middleware.ts
```diff
@@ -10,75 +10,78 @@ import sites from '.sitecore/sites.json';
...
-  // The provided callback will be called on each request to extract geo data.
-  // extractGeoDataCb: () => {
-  //   return {
-  //     city: 'Athens',
-  //     country: 'Greece',
-  //     region: 'Attica',
-  //   };
-  // },
-});
+  const personalize = new PersonalizeMiddleware({
```
illiakovalenko, line 60: What happens if the configuration is not provided when the personalize middleware is executed? Do we have any safeguards?

packages/nextjs/src/middleware/redirects-middleware.ts
```diff
@@ -42,14 +42,30 @@ export type RedirectsMiddlewareConfig = Omit<RedirectsServiceConfig, 'fetch' | '
...
+    const hasEdgeConfig = !!(this.config.contextId || this.config.clientContextId);
+    const hasLocalConfig = !!(this.config.apiHost && this.config.apiKey);
+
+    if (!hasEdgeConfig && !hasLocalConfig) {
+      console.warn(
+        '[RedirectsMiddleware] Redirects middleware requires either Edge configuration (contextId/clientContextId) or local API configuration (apiHost/apiKey). ' +
+          'Redirects features will be disabled. This is expected when API configuration is not available.'
+      );
+      // Set to null to indicate service is disabled
+      this.redirectsService = null;
```
illiakovalenko, line 65: Instead of changing the `redirectsService` type to allow `null`, could we handle this directly in the handler? We could check whether the configuration is present at runtime, and if it isn’t, simply log a debug-level message and return early.
We can extend `disabled` method to don't pollute our "handle" method

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/middleware.ts
MenKNas: @illiakovalenko Yes, we have safeguards. `PersonalizeMiddleware` checks for Edge config in the constructor (L:72-80) and sets the service to null if missing. I also refactored it to use `disabled()` for the runtime check (same pattern as `RedirectsMiddleware`), so both middlewares handle missing config consistently.

illiakovalenko (approved):

PR #334, opened 2026-01-08
Upgrade to NextJS 16 and add new caching mechanism

MenKNas:
- [x] Changelog updated
- [x] Upgrade guide entry added

## Description / Motivation

This PR introduces Next.js v16 across the CSDK and both Next.js templates (Pages Router and App Router), including peer/dev dependencies, as well as `eslint-config-next` alignment. More specifically:

- Migrated middleware to the new proxy model for generated apps
- Enabled Next.js 16 performance features in templates: turn on `cacheComponents` and update image handling to rely solely on `remotePatterns` (no images.domains).
- Aligned with Next.js 16 prerendering rules by removing export `const dynamic = 'force-dynamic'` from API routes now incompatible with `cacheComponents`  and updating handlers to correctly bail out of prerendering and wrap uncached data access in `<Suspense>` where required.
- Replaced `Date.now()` with `performance.now()` in the GraphQL client for safe timing in Server Components, and adjusted robots/sitemap handlers’ timing to occur after request data access.
- Removed unused React imports and cleaned up FEaaS/image logic in templates.

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other -> Tested by scaffolding both Next.js templates (Pages Router and App Router) and confirming the build completes and the apps run without errors.

## Types of changes


packages/create-content-sdk-app/src/templates/nextjs-app-router/package.json
```diff
@@ -33,7 +33,7 @@
     "@sitecore-content-sdk/nextjs": "<%- version %>",
     "@sitecore-feaas/clientside": "^0.6.0",
     "@sitecore/components": "~2.1.0",
-    "next": "^15.5.9",
+    "next": "^16.0.0",
```
art-alexeyenko: let's use 16.1.1 here, latest release atm

packages/create-content-sdk-app/src/templates/nextjs/package.json
```diff
@@ -27,7 +27,7 @@
     "@sitecore-content-sdk/nextjs": "<%- version %>",
     "@sitecore-feaas/clientside": "^0.6.0",
     "@sitecore/components": "~2.1.0",
-    "next": "^15.5.9",
+    "next": "^16.0.0",
```
art-alexeyenko: Same for for the 16.1.1 version of next here

packages/create-content-sdk-app/src/templates/nextjs/package.json
art-alexeyenko: Please adjust next build scripts for webpack in this and app-router samples

packages/nextjs/src/middleware/middleware.ts
```diff
@@ -236,18 +236,21 @@ export const defineMiddleware = (...middlewares: Middleware[]) => {
...
       const response = res || NextResponse.next();
 
       debug.common('middleware start');
 
       const start = Date.now();
 
+      // For Next.js 16+, ev is optional. Pass undefined if not provided.
+      const fetchEvent = ev || ({} as NextFetchEvent);
+
       const middlewareResponse = await middlewares.reduce(
```
art-alexeyenko: Do you think we should rename everything `middleware*` into `proxy*` for consistency with nextJS?

packages/nextjs/src/middleware/middleware.ts
```diff
@@ -236,18 +236,21 @@ export const defineMiddleware = (...middlewares: Middleware[]) => {
...
-    exec: async (req: NextRequest, ev: NextFetchEvent, res?: NextResponse) => {
+    exec: async (req: NextRequest, ev?: NextFetchEvent, res?: NextResponse) => {
       const response = res || NextResponse.next();
 
       debug.common('middleware start');
 
       const start = Date.now();
 
+      // For Next.js 16+, ev is optional. Pass undefined if not provided.
+      const fetchEvent = ev || ({} as NextFetchEvent);
```
art-alexeyenko: maybe we can get rid of 'ev' here altogether? `proxy` in nextjs now only accepts a request

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/layout.tsx
```diff
@@ -1,6 +1,19 @@
 import { draftMode } from 'next/headers';
+import { Suspense } from 'react';
 import Bootstrap from 'src/Bootstrap';
 
+// Component that handles draft mode check (uncached data access)
+async function SiteLayoutContent({ site, children }: { site: string; children: React.ReactNode }) {
```
Comment (art-alexeyenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/layout.tsx:6]: is it neccessary to have it as a separate function? It's not that large to break it apart

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -35,3 +37,19 @@ export default async function NotFound() {
     </div>
   );
 }
+
+export default function NotFound() {
```
Comment (art-alexeyenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/not-found.tsx:41]: also here, is it required to divide the functions?

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/layout.tsx
Comment (MenKNas) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/layout.tsx:6]: Yes, the split is required for Next.js 16 PPR compatibility. Components using uncached APIs (headers(), draftMode(), etc.) must be inside a `<Suspense />`  boundary, or Next.js throws an error. The default export provides the `<Suspense />` boundary, and the inner async component performs the uncached data access. This enables PPR to prerender the static shell while streaming dynamic content. 

You can read more about the use of  `<Suspense />`  with cached components in order to avoid errors on the link below:

https://nextjs.org/docs/messages/blocking-route

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/not-found.tsx
Comment (MenKNas) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/not-found.tsx:41]: See comment above

packages/nextjs/src/middleware/middleware.ts
MenKNas: That's a large-scale change and would cause the customers to make a lot of changes in order to upgrade. I would recommend against it since 'middleware' is still the standard terminology for this pattern, and Next.js only renamed their file/function, not the concept itself. But we can discuss it if you think there are strong benefits.

art-alexeyenko: It does not look like a large commitment for customers: the only changes needed are some renames within `middleware.ts` (now `proxy.ts`) and the /middleware submodule rename. 
The rest, an AI agent or simply the find-and-replace VS code tool can help

packages/nextjs/src/editing/editing-config-proxy.test.ts
```diff
@@ -1,8 +1,8 @@
-﻿﻿/* eslint-disable no-unused-expressions */
+/* eslint-disable no-unused-expressions */
 import { NextApiRequest, NextApiResponse } from 'next';
 import { spy } from 'sinon';
 import { expect } from 'chai';
-import { EditingConfigMiddleware } from './editing-config-middleware';
```
art-alexeyenko, line 5: let's only rename things related to nextjs middleware
EditingConfig and other editing middleware are not a part of middleware (now proxy) - so their names can stay the same

art-alexeyenko (approved):
Thanks for the thorough investigation and for bringing this to the finish.
Please feel free to merge once all the remaining open conversations are addressed.

PR #343, opened 2026-01-20
[create-content-sdk-app] Fixes for cache components related build issues

MenKNas:
- [X] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)
- [X] Changelog updated

## Fix Next.js 16 Cache Components compatibility

Fixes build errors in Next.js App Router template when using Next.js 16 with Cache Components:

- Fixed `EmptyGenerateStaticParamsError` by ensuring `generateStaticParams` returns at least one result
- Fixed TypeScript `JSX.Element` error in `CdpPageView.tsx` 
- Fixed `new Date()` errors by adding `await draftMode()`/`await cookies()` before client method calls in server components

## Testing Details

- [ ] Unit Test Added
- [X] Manual Test/Other (Please elaborate)

## Types of changes

- [X] Bug fix (non-breaking change which fixes an issue)


packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/page.tsx
```diff
@@ -51,6 +51,10 @@ async function PageContent({ site, locale, path, searchParams }: { site: string;
 }
 
 export default async function Page({ params, searchParams }: PageProps) {
+  // Access uncached data first to satisfy Next.js 16 Cache Components requirements
+  // This ensures any time-related operations (like new Date()) can be used safely
+  await draftMode();
```
Comment (art-alexeyenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/page.tsx]: considering the dynamic nature of Sitecore websites, there could be other instances of functionality that breaks cache components, and forces us to add more workarounds like this.
If it's not possible to limit caching to individual components (and it needs to be active from the root down), let's not have it enabled in empty sample by default and instead provide documentation on how users can do it (and enable it in example starters instead)

art-alexeyenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/page.tsx
Comment (MenKNas) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/[locale]/[[...path]]/page.tsx]: This is a good point, and I had similar thoughts myself. I had added these workarounds to support Cache Components being enabled by default, but I agree with the concern raised above, so I removed these workarounds and disabled Cache Components by default. I also added documentation in the README.md file in the app router template that explains the steps for enabling this feature in detail (I wasn't sure where exactly to document this, if you have some other idea feel free to share). So now users can follow the doc if they want to opt-in for this feature.

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/layout.tsx
```diff
@@ -2,8 +2,11 @@ import { draftMode } from 'next/headers';
 import { Suspense } from 'react';
 import Bootstrap from 'src/Bootstrap';
 
-// Component that handles draft mode check (uncached data access)
-async function SiteLayoutContent({ site, children }: { site: string; children: React.ReactNode }) {
+// Dynamic layout component that handles draft mode check.
```
Comment (art-alexeyenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/[site]/layout.tsx]:
```suggestion
// Part of layout component that handles dynamic data, like the `draftMode()` call
```

packages/create-content-sdk-app/src/templates/nextjs/README.md
```diff
@@ -5,3 +5,36 @@
 -->
 
 [Documentation](https://doc.sitecore.com/xmc/en/developers/xm-cloud/sitecore-javascript-rendering-sdk--jss--for-next-js.html)
+
+## Next.js 16 Cache Components
```
art-alexeyenko: let's not put this into README, and have it in official docs instead

packages/create-content-sdk-app/src/templates/nextjs-app-router/next.config.ts
```diff
@@ -2,9 +2,11 @@ import type { NextConfig } from 'next';
 import createNextIntlPlugin from 'next-intl/plugin';
 
 const nextConfig: NextConfig = {
-  // Enable Cache Components for explicit caching with "use cache" directive
+  // Cache Components is disabled by default. To enable it, set cacheComponents: true
```
art-alexeyenko: `cacheComponents` setting is part of nextjs configuration, and can be found there. Let's keep the config light on conent and remove the comment and the commented out settings

art-alexeyenko (approved):
Approved in advance - looks good, just needs a couple light adjustments

PR #345, opened 2026-01-21
[nextjs] [Pages Router] Adjust static path generation when multisite is disabled

yavorsk:
- [x] Changelog updated

## Description / Motivation

When multisite is disabled in `sitecore.config` MultisiteProxy does not run and does not rewrite paths by adding site segment ( `/_site_my-site/...`), while the same setting does not affect the behavior of `getPagePaths` - paths are generated with the site segment. 
This PR adjusts `getPagePaths` to eliminate this discrepancy.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - in Pages Router app explicitly disable multisite; run in prod mode; verify that generated paths are without site prefix, personalize and redirects should work as expected.

## Types of changes

### DIFF
--- a/packages/nextjs/src/client/sitecore-nextjs-client.ts
+++ b/packages/nextjs/src/client/sitecore-nextjs-client.ts
@@ -25,6 +25,13 @@ import {
 } from '@sitecore-content-sdk/core/personalize';
 import { ComponentMap } from '@sitecore-content-sdk/react';
 import { StaticParams } from './models';
+import { SitecoreConfig } from '../config';
+
+/**
+ * Init options for Sitecore Client that allows you to override services too
+ * @public
+ */
+export type SitecoreNextjsClientInit = SitecoreClientInit & Pick<SitecoreConfig, 'multisite'>;
 
 /**
  * The SitecoreNextjsClient class extends the SitecoreClient class to provide additional functionality for Next.js.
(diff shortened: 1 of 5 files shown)


packages/nextjs/src/client/sitecore-nextjs-client.ts
```diff
@@ -143,6 +144,32 @@ export class SitecoreNextjsClient extends SitecoreClient {
...
 
+  /**
+   * Retrieves the static paths for pages based on the given languages.
+   * @param {string[]} sites - An array of site names to fetch routes for.
+   * @param {string[]} [languages] - An optional array of language codes to generate paths for.
+   * @param {FetchOptions} [fetchOptions] - Additional fetch options.
+   * @param {boolean} [multisiteEnabled] - Indicates if multisite support is enabled.
+   * @returns {Promise<StaticPath[]>} A promise that resolves to an array of static paths.
+   */
+  async getPagePaths(
```
illiakovalenko, line 161: You may agree that this new API change doesn’t look ideal in its current form. I’d like to suggest an alternative approach.

Technically, we already have access to the multisite settings in the Client. They were intentionally [omitted](https://github.com/Sitecore/content-sdk/blob/dev/packages/core/src/client/models.ts#L12C13-L12C32) in the core Client by default, but since they are required for the Next.js client, we can safely expose them there.

Moreover, these settings are already part of SitecoreConfig via defineConfig, so this change would not introduce a new concept - only make existing configuration explicitly available where it’s needed

```
import { SitecoreConfig } from '../../config';

export type SitecoreNextjsClientInit = SitecoreClientInit & Pick<SitecoreConfig, 'multisite'>;
...
if (!this.initOptions?.multisite.enabled) {
```

illiakovalenko (approved):
Looks good!
Please, see one suggested improvement below

packages/nextjs/src/client/sitecore-nextjs-client.ts
yavorsk: Makes sense, updated.

PR #344, opened 2026-01-20
[nextjs] Preserve `basePath` when doing redirects in redirects-proxy

yavorsk:
- [x] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)
- [x] Changelog updated
- [ ] Upgrade guide entry added

## Description / Motivation

When `basePath` is configured in NextJs config make sure that it is preserved when doing redirects in redirects-proxy. 
Example:
 - `basePath` is configured in next.config: `basePath: '/base-path',`
 - we have the following redirect map configured in Sitecore AI: `/old-page` -> `/new-page`
 - Navigating 'https://my-site.com/base-path/old-page' should redirect to 'https://my-site.com/base-path/new-page'

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - all redirects with configured NextJs `basePath` should work properly; test with both client side and server side navigation

## Types of changes

- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #352, opened 2026-01-26
[nextjs]  Fix basePath preservation logic in redirects-proxy in case `basePath` is not configured; Fix nextjs specific header names in redirects-proxy

yavorsk:
- [x] Changelog updated

## Description / Motivation

This PR is a follow-up to #344 and addresses two issues:
 - when assigning `basePath` property of NextUrl, if `basePath` is empty string it gets assigned as '/' instead of ''; this causes redirects to not work properly when `basePath` is not configured;
 - fix the names of the next js specific headers that are removed from the response; this was causing redirects to not work properly in case of pages router application and client side navigation.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - test both app router and pages router applications, with and without `basePath`, in both client side and server side navigation;

## Types of changes

### DIFF
--- a/CHANGELOG.md
+++ b/CHANGELOG.md
@@ -47,7 +47,7 @@ Our versioning strategy is as follows:
 
 * `[nextjs]` Sitecore Content SDK does not support X-Forwarded-Host, causing incorrect hostname resolution behind proxies ([#330](https://github.com/Sitecore/content-sdk/pull/330))
 * `[core]` `[search]` `[analytics]` Pass Sitecore Context ID only in headers ([#336](https://github.com/Sitecore/content-sdk/pull/336))
-* `[nextjs]` Fix basePath preservation and header names in redirects-proxy ([#352](https://github.com/Sitecore/content-sdk/pull/352))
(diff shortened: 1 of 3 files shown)


CHANGELOG.md
```diff
@@ -47,6 +47,7 @@ Our versioning strategy is as follows:
 
 * `[nextjs]` Sitecore Content SDK does not support X-Forwarded-Host, causing incorrect hostname resolution behind proxies ([#330](https://github.com/Sitecore/content-sdk/pull/330))
 * `[core]` `[search]` `[analytics]` Pass Sitecore Context ID only in headers ([#336](https://github.com/Sitecore/content-sdk/pull/336))
+* `[nextjs]` Fix basePath preservation and header names in redirects-proxy ([#352](https://github.com/Sitecore/content-sdk/pull/352))
```
illiakovalenko: This entry largery repeats what we already covered in the 1.4.0 release. Perhaps we can make it more specific - for example, by explicitly mentioning the “not defined” case?
The same applies to PR title

illiakovalenko (approved):
Looks good 👍
Please, review one suggestion below

CHANGELOG.md
yavorsk: thanks. yea, I will update

PR #357, opened 2026-01-28
Rework Analytics initialization

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation

Introduce new initialization functionality for the SDK and use it for the analytics packages.
- Add initSitecore to core package to enable initialization
- Adapt analytics-core, events, personalize packages to the new init function
- Create adapters for proxy environments for analytics and personalize packages in the nextjs package
- Use the new init functions in both nextjs templates

## Testing Details

- [ ] Unit Test Added

## Types of changes


CHANGELOG.md
```diff
@@ -30,6 +30,8 @@ Our versioning strategy is as follows:
 
 * `[nextjs]` `[Pages Router]` Adjust static path generation when multisite is disabled ([#345](https://github.com/Sitecore/content-sdk/pull/345))
 
+* `[sdk]` Rework Analytics initialization ([#357](https://github.com/Sitecore/content-sdk/pull/357))
```
illiakovalenko: I would suggest creating a single changelog entry (together with [previous](https://github.com/Sitecore/content-sdk/pull/357/files#diff-06572a96a58dc510037d5efa622f9bec8519bc1beab13c9f251e97e657a9d4edR29) one) where we can collect all Cloud SDK migration - related changes, it should be highligthed, for example:
```
* Introduce Analytics packages:
  * feature 1 (pr1)
  * feature 2 (pr2)
  ...
```

packages/analytics-core/package.json
```diff
@@ -29,26 +30,24 @@
     "ts-jest": "^29.1.0",
     "ts-node": "^10.9.2"
   },
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "types/src/index.d.ts",
   "exports": {
```
illiakovalenko, line 36: We need to update `generate-docs` npm command, there are references to "server", "browser" submodules

packages/events/src/debug.ts
```diff
@@ -0,0 +1,11 @@
+import { debugNamespace, debugModule } from '@sitecore-content-sdk/core';
+
+export const EVENTS_NAMESPACE = 'events';
+/**
+ * Debug module for content package
```
illiakovalenko:
```suggestion
 * Debug module for events package
```

packages/core/src/initialization/consts.ts
```diff
@@ -0,0 +1,7 @@
+export const ERROR_MESSAGES = {
+  IE_0002: '[IE-0002] - You must first initialize the SDK using "initSitecore"',
```
illiakovalenko:
```suggestion
  IE_0002: '[IE-0002] You must first initialize the SDK using "initSitecore"',
```

packages/core/src/index.ts
```diff
@@ -20,3 +20,6 @@ export {
 } from './native-fetcher';
 export { RetryStrategy, GenericGraphQLClientError, FetchOptions } from './models';
 export { constants };
+export { initSitecore, getCoreSettings } from './initialization/init-sitecore';
+export { CoreSettings } from './initialization/types';
+export { Plugin, PluginDependency, PluginEnvironment } from './initialization/types';
```
illiakovalenko: Could you please create an index file under the `initialization` folder? It would make it easier to track and manage all exports.

packages/core/src/initialization/types.ts
```diff
@@ -0,0 +1,72 @@
...
+    contextId: string;
+    sitecoreEdgeUrl?: string;
+    siteName: string;
+  };
+  /** Array of plugins to initialize */
+  plugins: Plugin[];
+}
+
+/**
+ * Internal settings of the SDK initialization
```
illiakovalenko, line 17: For the new types, could you please go through them and add the necessary API Extractor tags (internal, public, etc.)? This particular type is described as internal, but it is currently exported - presumably intended for use only by our packages - so we should add the appropriate tag.

packages/core/src/initialization/helpers.ts
```diff
@@ -0,0 +1,71 @@
...
+ * @returns A promise that resolves when all plugins have been initialized.
+ * @internal
+ */
+export async function initPlugins(plugins: Map<string, Plugin>): Promise<void> {
+  debugInit(`Initializing ${plugins.size} plugins`);
+
+  for (const plugin of plugins.values()) {
+    checkPluginDependencies(plugin, plugins);
+
+    await plugin.init?.();
```
illiakovalenko, line 38: Should we add proper error handling in case any uncaught errors are thrown? What happens if one of the plugins fails?

packages/events/package.json
```diff
@@ -31,16 +32,24 @@
     "ts-jest": "^29.4.6",
     "ts-node": "^10.9.2"
   },
+  "main": "dist/cjs/index.js",
+  "module": "dist/esm/index.js",
+  "types": "types/src/index.d.ts",
   "exports": {
```
illiakovalenko, line 38: Do not forget to update "generate-docs" command here as well

packages/core/src/initialization/types.ts
```diff
@@ -0,0 +1,72 @@
...
+ * Used to declare that a plugin requires another plugin to be present.
+ * @internal
+ */
+export type PluginDependency = string;
+
+/**
+ * Plugin interface for extending SDK functionality.
+ * Plugins are the standard way to add capabilities to the SDK.
+ * @template Settings - Plugin-specific settings type
+ * @internal
```
illiakovalenko: This type should be marked as public, as it is directly used in `InitSitecoreOptions`.

packages/core/src/initialization/types.ts
```diff
@@ -0,0 +1,72 @@
...
+  /**
+   * Promise that resolves when initialization is complete
+   */
+  readyPromise: Promise<void> | null;
+}
+
+/**
+ * Dependency definition for a plugin.
+ * Used to declare that a plugin requires another plugin to be present.
+ * @internal
```
illiakovalenko: The same as in case of `Plugin` type, it's a part of our public API. Let's mark it as public

packages/analytics-core/src/debug/debug.ts
```diff
@@ -1,19 +1,18 @@
-import debug from 'debug';
+import { enabled } from 'debug';
```
illiakovalenko: We should probably reference `debugModule` from the `core` package, as we do in other packages. This would ensure that all packages consistently rely on the same debug module.

packages/analytics-core/src/initialization/types.ts
illiakovalenko, line 1: Let’s double-check all new and exported types in all the packages, add the appropriate API Extractor tags, and provide descriptions for every property

packages/personalize/src/debug.ts
```diff
@@ -0,0 +1,11 @@
+import { debugNamespace, debugModule } from '@sitecore-content-sdk/core';
+
+export const PERSONALIZE_NAMESPACE = 'personalize';
+/**
+ * Debug module for content package
```
illiakovalenko:
```suggestion
 * Debug module for personalize package
```

illiakovalenko (changes requested):
This is a solid solution - great job!
Please see some review comments below

packages/analytics-core/package.json
sc-nikolaoslazaridis: yes, it will be done in the next task.

packages/core/src/initialization/consts.ts
sc-nikolaoslazaridis: We have the full list waiting in the next ticket.

packages/core/src/initialization/types.ts
sc-nikolaoslazaridis: Ill add the tags in this one and finalize the modules in the next ticket. (possibly create an "internal" module for the core package)

packages/analytics-core/src/debug/debug.ts
sc-nikolaoslazaridis: ill add a "isNamespaceEnabled" function

packages/analytics-core/src/initialization/types.ts
sc-nikolaoslazaridis: will be done in the next ticket

packages/core/src/initialization/helpers.ts
sc-nikolaoslazaridis: From my point of view:
- If a plugin can fail, it should be handled by the plugin itself. For example if events fail we shouldn't beak the app with an error, so the events plugin should handle that gracefully internally. If another mission critical plugin fails and the app can't work we should let the error bubble up and handled by the developer.
- If a plugin throws an error it should mean that the some miss-configuration happened or other external dependencies are down hence we need to throw that during dev time and let the devs handle it accordingly. That means try catching here would mean we just need to re-throw the error.

Your thoughts? We can include product for this one also.

illiakovalenko: Agree with that

illiakovalenko (approved):
👍

PR #359, opened 2026-02-02
Add custom hostname support

MenKNas:
- [x] Changelog updated

## Support custom Edge hostnames

This PR adds support for custom Sitecore Edge hostnames by resolving the Edge base URL from `SITECORE_EDGE_PLATFORM_HOSTNAME` (and, for Next.js, `NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME`) and ensuring SDK requests and layout/editing data can use the custom hostname instead of the default Edge Platform host. It also adds opt-in media URL rewriting so that URLs in layout data (e.g. image fields, rich text) can be rewritten to the custom hostname when desired.

## How it works

- **Edge URL resolution:** The Edge base URL is resolved once at config level (e.g. in `define-config`): explicit config value → `SITECORE_EDGE_PLATFORM_HOSTNAME` / Next.js public equivalent → default Edge Platform URL. Helpers receive the already-resolved URL and only normalize it; they do not read env or resolve again.
- **Custom hostname in responses:** When a custom hostname is configured, layout/editing/component responses can be post-processed to replace embedded default Edge hostnames with the resolved custom hostname (default rewriter in `rewrite-edge-host.ts`).
- **Opt-in media rewrite:** Media URL rewriting is off by default. When `rewriteMediaUrls: true` is set in Sitecore config, the client applies the default Edge host rewriter to layout data before returning it (rewriting media URLs and rich text to the custom hostname). When `rewriteMediaUrls` is a function, that function is used to transform each string in the layout.
- **Static files:** `resolveEdgeUrlForStaticFiles()` always returns the default Edge Platform URL so that static assets (e.g. stylesheets under `/v1/files/...`) continue to load from the platform when the custom host does not serve them.
- **Custom rewriter:** Apps can set `rewriteMediaUrls` to a custom `(value: string) => string` to implement their own URL transformation over layout strings.

## Key files

- `packages/core/src/tools/resolve-edge-url.ts` – Edge base URL resolution (used at config time); `resolveEdgeUrlForStaticFiles()` for static assets.
- `packages/content/src/layout/rewrite-edge-host.ts` – Rewrites default Edge hostnames in layout/editing responses to the resolved custom hostname; used as the default when `rewriteMediaUrls: true`.
- `packages/content/src/config/define-config.ts` – Resolves edge URL at config level; default `rewriteMediaUrls: false`.
- `packages/content/src/client/sitecore-client.ts` – Applies the `rewriteMediaUrls` rewriter when enabled.

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other => 

Tested manually in sample app using custom hostname provided in https://portal-staging.sitecore-staging.cloud/admin/custom-hostnames?organization=org_Jk9RJbQPa0m9Psul (test3-staging.envoy-test.cloud)

## Types of changes


packages/content/src/config/define-config.ts
```diff
@@ -55,6 +55,8 @@ export const getFallbackConfig = (): SitecoreConfig => ({
       timeout: 60,
     },
   },
+  rewriteContentUrls: false,
+  contentRewrite: rewriteEdgeHostInResponse,
```
illiakovalenko: Consider consolidating into a single option (it's a common practice): `rewriteMediaUrls: boolean | ((value: string) => string)`. When true, use the default rewriter; when a function, use that instead. One property is easier to manage and avoids the case where a custom contentRewrite is set but rewriteContentUrls is left false.
This method should allow to transform a string value and returns a new value.
We don't need to require devs to traverse all the layout by themselves

packages/content/src/config/define-config.ts
```diff
@@ -13,7 +13,7 @@ export const getFallbackConfig = (): SitecoreConfig => ({
     edge: {
       contextId: process.env.SITECORE_EDGE_CONTEXT_ID || '',
       clientContextId: '',
-      edgeUrl: process.env.SITECORE_EDGE_URL || SITECORE_EDGE_URL_DEFAULT,
+      edgeUrl: resolveEdgeUrl(),
```
illiakovalenko, line 15: There is no need to introduce complex logic to resolve the Edge URL, especially considering that this package is framework-agnostic.

Let’s rename it to `SITECORE_EDGE_PLATFORM_HOSTNAME` - this naming has already been accepted by Product.

At this level, we should reference either this environment variable or fall back to the default Edge URL. The core package must not depend on or reference any Next.js-specific APIs or concepts.

The Next.js-specific handling logic should remain within the nextjs package, where we already have dedicated implementation in place (see my related comment).

packages/content/src/layout/content-rewrite.ts
illiakovalenko, line 1: I suppose that content-rewrite is unused

packages/content/src/client/edge-proxy.ts
```diff
@@ -1,30 +1,27 @@
...
 export const getEdgeProxyFormsUrl = (
   sitecoreEdgeContextId: string,
   formId: string,
-  sitecoreEdgeUrl = SITECORE_EDGE_URL_DEFAULT
+  sitecoreEdgeUrl?: string
 ) =>
-  `${normalizeUrl(
-    sitecoreEdgeUrl
-  )}/v1/forms/publisher/${formId}?sitecoreContextId=${sitecoreEdgeContextId}`;
+  `${resolveEdgeUrl(sitecoreEdgeUrl)}/v1/forms/publisher/${formId}?sitecoreContextId=${sitecoreEdgeContextId}`;
```
illiakovalenko: This **comment** applies to **ALL** similar occurrences where `SitecoreConfig` is involved or where `edgeUrl` is passed:

There is no need to call `resolveEdgeUrl` in each case, as the resolved Edge URL is already available within `SitecoreConfig`.
You can verify this by debugging and confirming that the value is properly resolved at the configuration level.

packages/content/src/client/sitecore-client.ts
```diff
@@ -357,25 +359,25 @@ export class SitecoreClient implements BaseSitecoreClient {
...
-        personalizeLayout(
-          layout,
-          pageOptions.personalize.variantId,
-          pageOptions.personalize.componentVariantIds
-        );
-      }
-
-      return {
+    }
+    layout = this.applyContentRewrite(layout);
```
illiakovalenko: First, we need to apply personalization. Only after that should we proceed with rewriting, since there is no need to process or rewrite variants that are not used

packages/content/src/layout/rewrite-edge-host.ts
```diff
@@ -0,0 +1,126 @@
+import { hasCustomEdgeHostname, resolveEdgeUrl } from '@sitecore-content-sdk/core/tools';
+
+/**
+ * The default Edge Platform hostnames that may appear in responses.
+ * These will be replaced with the custom hostname when configured.
+ * Includes production and staging so layout content (e.g. media URLs) can be rewritten.
+ * @internal
+ */
+const DEFAULT_EDGE_HOSTNAMES = [
+  'edge-platform.sitecorecloud.io',
```
illiakovalenko: Is there a scenario where this URL could appear in media URLs? Do we have a concrete use case for that?
Overall, it seems that we expect only the Experience Edge URL to be used as the value

packages/content/src/layout/rewrite-edge-host.ts
```diff
@@ -0,0 +1,126 @@
...
+  'edge.sitecorecloud.io', // Legacy hostname included for safety replacement
```
illiakovalenko: This is not a legacy URL - it is an Experience Edge URL.

We can leave default production url, but we cannot hardcode or enumerate custom environments here (staging, dev etc.). Customers and some internal teams may use additional environments, and with the current implementation they would not be able to configure custom hostnames.

To address this properly, we should introduce an **internal** environment variable: SITECORE_EXPERIENCE_EDGE_HOSTNAME. This variable would need to be populated by the Deploy team, allowing custom environment hostnames to be specified without constraining the implementation.

illiakovalenko: This value should be defined as a constant within the core package

packages/nextjs/src/config/define-config.ts
```diff
@@ -19,7 +20,7 @@ export const getNextFallbackConfig = (config?: SitecoreConfigInput): SitecoreCon
         contextId: config?.api?.edge?.contextId || '',
         clientContextId:
           config?.api?.edge?.clientContextId || process.env.NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID,
-        edgeUrl: config?.api?.edge?.edgeUrl || process.env.NEXT_PUBLIC_SITECORE_EDGE_URL,
+        edgeUrl: resolveEdgeUrl(config?.api?.edge?.edgeUrl),
```
illiakovalenko: This is the location where you should fall back to `NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME`, you don't really need to change it much here

illiakovalenko (changes requested):
Please see my feedback and suggestions for improvement

packages/create-content-sdk-app/src/templates/nextjs/sitecore.config.ts
```diff
@@ -4,4 +4,7 @@ import { defineConfig } from '@sitecore-content-sdk/nextjs/config';
  * See the documentation for `defineConfig`:
  * https://doc.sitecore.com/xmc/en/developers/content-sdk/the-sitecore-configuration-file.html
  */
-export default defineConfig({});
+export default defineConfig({
+  // Enable to use custom hostname for content/media URLs (default: false)
+  // rewriteContentUrls: true,
```
illiakovalenko: I'm not entirely sure this specific setting needs to be documented here. It’s probably better to keep the configuration clean, as it was before

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
```diff
@@ -19,6 +19,12 @@ SITECORE_EDGE_CONTEXT_ID=
 # Will be used as a fallback if separate SITECORE_EDGE_CONTEXT_ID value is not provided
 NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 
+# Optional: custom Experience Edge hostname override (XM Cloud/Edge; hostname or full URL).
+SITECORE_EDGE_HOSTNAME=
+
+# Optional: custom Experience Edge hostname override for client-side use (XM Cloud/Edge).
+NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME=
```
illiakovalenko: It's not required to define two environment variables, since `NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME` is available on both server and client side

packages/create-content-sdk-app/src/templates/nextjs/.env.container.example
illiakovalenko, line 1: I'm not sure this is applicable to local development with containers

packages/content/src/layout/rewrite-edge-host.ts
MenKNas: Yes, most likely. Experience Edge returns Layout Service output where component fields contain full URLs. Image fields have `value.src` (e.g. `https://edge-platform.sitecorecloud.io/-/media/...`), and Rich Text fields embed `` and `<a href="...">`. So when publishing to Edge, these often include the Edge hostname.

packages/content/src/client/sitecore-client.ts
```diff
@@ -391,18 +394,20 @@ export class SitecoreClient implements BaseSitecoreClient {
     options: { enableStyles?: boolean; enableThemes?: boolean } = {}
   ): HTMLLink[] {
     const { enableStyles = true, enableThemes = true } = options;
-    const { contextId: serverContextId, clientContextId, edgeUrl } = this.initOptions.api.edge;
+    const { contextId: serverContextId, clientContextId } = this.initOptions.api.edge;
     const headLinks: HTMLLink[] = [];
 
     const contextId = serverContextId || clientContextId;
```
sc-nikolaoslazaridis, line 403: @illiakovalenko Since the client can be used in server and browser environments, is the usage of serverContextId as the first option the way to go here? Im just thinking that the server context id can be leaked by mistake since it is a required settings.

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -0,0 +1,151 @@
...
+function normalizeMaybeEnvValue(value: string | undefined): string | undefined {
+  if (!value) return undefined;
+
+  const trimmed = value.trim();
+  if (!trimmed) return undefined;
+
+  const lowered = trimmed.toLowerCase();
+  if (lowered === 'undefined' || lowered === 'null') return undefined;
+
+  return trimmed;
```
sc-nikolaoslazaridis: just personal preference

```suggestion
   const normalizedValue = value?.trim().toLocaleLowerCase();

  return !normalizedValue || normalizedValue === 'undefined' || normalizedValue === 'null'
    ? undefined
    : normalizedValue;
```

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Resolves the Edge URL for static files (e.g. stylesheets) by ignoring the custom hostname.
+ * Use this when the custom host does not serve static file paths (e.g. /v1/files/...).
+ * Priority: SITECORE_EDGE_URL / NEXT_PUBLIC_SITECORE_EDGE_URL env, then default.
+ * @returns {string} The Edge Platform base URL for static files (no trailing slash)
+ * @public
+ */
+export function resolveEdgeUrlForStaticFiles(): string {
+  const isBrowser = typeof window !== 'undefined';
```
sc-nikolaoslazaridis: We have a isServer function, can we use it here?

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -0,0 +1,151 @@
...
+
+/**
+ * Normalizes values that may come from environment variables.
+ * In Node, setting `process.env.FOO = undefined` results in the string 'undefined',
+ * which should be treated as if the variable is not set.
+ * @param {string | undefined} value - Possibly undefined env-like value
+ * @returns {string | undefined} A usable string value, or undefined if not meaningful
+ * @internal
+ */
+function normalizeMaybeEnvValue(value: string | undefined): string | undefined {
```
sc-nikolaoslazaridis: Is it possible to remove the "Maybe" from the name and make it a standalone function not in the resolve-edge-url file since it looks like a generic one?

packages/content/src/client/edge-proxy.ts
```diff
@@ -1,30 +1,35 @@
-import { constants } from '@sitecore-content-sdk/core';
-import { normalizeUrl } from '@sitecore-content-sdk/core/tools';
+import { normalizeUrl, resolveEdgeUrl } from '@sitecore-content-sdk/core/tools';
 
-const { SITECORE_EDGE_URL_DEFAULT } = constants;
+/**
```
sc-nikolaoslazaridis, line 4: Let't try to keep the codebase warning-free please

packages/content/src/config/define-config.ts
```diff
@@ -13,7 +17,11 @@ export const getFallbackConfig = (): SitecoreConfig => ({
     edge: {
       contextId: process.env.SITECORE_EDGE_CONTEXT_ID || '',
       clientContextId: '',
-      edgeUrl: process.env.SITECORE_EDGE_URL || SITECORE_EDGE_URL_DEFAULT,
+      edgeUrl: resolveEdgeUrl(
+        process.env[SITECORE_EDGE_HOSTNAME_PUBLIC_ENV] ??
+          process.env[SITECORE_EDGE_URL_PUBLIC_ENV] ??
+          process.env[SITECORE_EDGE_URL_ENV]
```
sc-nikolaoslazaridis: Should we add as the last fallback option here the default edge url?

packages/content/src/config/define-config.ts
```diff
@@ -110,6 +119,13 @@ const resolveConfig = (base: SitecoreConfig, override: SitecoreConfigInput): Sit
   if (Number.isNaN(result.personalize.edgeTimeout) || !result.personalize.edgeTimeout) {
     result.personalize.edgeTimeout = base.personalize.edgeTimeout;
   }
+  // Resolve edge URL at config level so consumers use the resolved value directly
+  result.api.edge.edgeUrl = resolveEdgeUrl(
+    result.api.edge.edgeUrl ??
+      process.env[SITECORE_EDGE_HOSTNAME_PUBLIC_ENV] ??
+      process.env[SITECORE_EDGE_URL_PUBLIC_ENV] ??
+      process.env[SITECORE_EDGE_URL_ENV]
```
sc-nikolaoslazaridis: Same as the above comment

packages/content/src/config/define-config.ts
```diff
@@ -13,7 +17,11 @@ export const getFallbackConfig = (): SitecoreConfig => ({
     edge: {
       contextId: process.env.SITECORE_EDGE_CONTEXT_ID || '',
       clientContextId: '',
-      edgeUrl: process.env.SITECORE_EDGE_URL || SITECORE_EDGE_URL_DEFAULT,
+      edgeUrl: resolveEdgeUrl(
```
sc-nikolaoslazaridis: Internally the resolveEdgeUrl does partially the same checks, do we need to pass the array explicitly here? Maybe we could all that internally?

packages/content/src/editing/editing-service.ts
```diff
@@ -100,14 +100,16 @@ export class EditingService {
...
+    const layoutData = editingData?.item?.rendered || {
+      sitecore: {
+        context: { pageEditing: true, language },
+        route: null,
       },
     };
+
+    return {
+      layoutData,
+    };
```
sc-nikolaoslazaridis, line 112: personal preference, don't pay attention if don't like it
```suggestion
    return {
      layoutData: editingData?.item?.rendered || {
        sitecore: {
          context: { pageEditing: true, language },
          route: null,
        },
      },
    };
```

packages/content/src/layout/rewrite-edge-host.ts
```diff
@@ -0,0 +1,166 @@
...
+ * const layout = await layoutService.fetchLayoutData(path, options);
+ * const rewritten = rewriteEdgeHostInResponse(layout);
+ */
+export function rewriteEdgeHostInResponse<T>(response: T): T {
+  // Skip if no custom hostname is configured
+  if (!hasCustomEdgeHostname()) {
+    return response;
+  }
+
+  const customEdgeUrl = resolveEdgeUrl();
```
sc-nikolaoslazaridis: What happens if the dev provided an URL in the defineConfig and not in the env variables? The resolveEdgeUrl function doesn't check the config object internally.

packages/content/src/client/sitecore-client.ts
illiakovalenko: @sc-nikolaoslazaridis It will not be leaked since env variables defined without NEXT_PUBLIC don't expose values to the browser, unless it's set explicitly by developer, but it's properly documented

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -0,0 +1,131 @@
...
+import isServer from './is-server';
+import { normalizeEnvValue } from './normalize-env-value';
+import { normalizeUrl } from './normalize-url';
+
+/**
+ * Environment variable name for the custom Edge hostname.
+ * Available on both server and client (e.g. NEXT_PUBLIC_* in Next.js).
+ * @public
+ */
+export const SITECORE_EDGE_HOSTNAME_PUBLIC_ENV = 'NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME';
```
illiakovalenko: Please refer to my previous [comment](https://github.com/Sitecore/content-sdk/pull/359#discussion_r2798429197) where I pointed out that the core package should not reference Next.js–specific environment variables. This should instead be handled at the nextjs package level, for example via a defineConfig override. Also it's not renamed to "SITECORE_EDGE_PLATFORM_HOSTNAME"

packages/content/src/editing/design-library.ts
```diff
@@ -223,12 +220,13 @@ export function getDesignLibraryStatusEvent(
...
- * @param {string} [sitecoreEdgeUrl] Sitecore Edge Platform URL. Default is https://edge-platform.sitecorecloud.io
+ * @param {string} [sitecoreEdgeUrl] Sitecore Edge Platform URL. If not provided,
+ *   resolves from NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME or SITECORE_EDGE_URL env vars, falling back to default.
  * @returns The full URL to the design library script.
  * @internal
  */
-export function getDesignLibraryScriptLink(sitecoreEdgeUrl = SITECORE_EDGE_URL_DEFAULT): string {
-  return `${normalizeUrl(sitecoreEdgeUrl)}/v1/files/designlibrary/lib/rh-lib-script.js`;
+export function getDesignLibraryScriptLink(sitecoreEdgeUrl?: string): string {
+  return `${(sitecoreEdgeUrl ? normalizeUrl(sitecoreEdgeUrl) : resolveEdgeUrl())}/v1/files/designlibrary/lib/rh-lib-script.js`;
```
illiakovalenko: Please refer to my earlier [comment](https://github.com/Sitecore/content-sdk/pull/359#discussion_r2798442665). The current approach still introduces unnecessary complexity without a clear justification.

As mentioned before, the edge URL should be resolved and normalized in a single place - at the configuration level. There is no need to repeat this logic every time the edge URL is passed as a parameter.

This applies to **ALL** occurrences

packages/content/src/layout/rewrite-edge-host.ts
illiakovalenko: @MenKNas This comment is not resolved, as no changes have been applied, except of moving to "core"

packages/core/src/constants.ts
```diff
@@ -4,6 +4,23 @@
  */
 export const SITECORE_EDGE_URL_DEFAULT = 'https://edge-platform.sitecorecloud.io';
```
illiakovalenko:
```suggestion
export const SITECORE_EDGE_PLATFORM_URL_DEFAULT = 'https://edge-platform.sitecorecloud.io';
```

packages/core/src/constants.ts
```diff
@@ -4,6 +4,23 @@
...
+ * - Link field href values
+ * - Other URL fields in component data
+ * @public
+ */
+export const DEFAULT_EDGE_HOSTNAMES = [
+  'edge-platform.sitecorecloud.io',
+  'edge.sitecorecloud.io',
+  'edge-staging.sitecore-staging.cloud',
+  'edge-platform-staging.sitecore-staging.cloud',
+] as const;
```
illiakovalenko:
```suggestion
export const SITECORE_EXPERIENCE_EDGE_URL_DEFAULT = 'https://edge.sitecorecloud.io'
```

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
```diff
@@ -19,6 +19,10 @@ SITECORE_EDGE_CONTEXT_ID=
 # Will be used as a fallback if separate SITECORE_EDGE_CONTEXT_ID value is not provided
 NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 
+# Optional: custom Experience Edge hostname override (XM Cloud/Edge; hostname or full URL).
+# Available on both server and client.
+NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME=
```
illiakovalenko:
```suggestion
NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME=
```

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
```diff
@@ -19,6 +19,10 @@ SITECORE_EDGE_CONTEXT_ID=
 # Will be used as a fallback if separate SITECORE_EDGE_CONTEXT_ID value is not provided
 NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 
+# Optional: custom Experience Edge hostname override (XM Cloud/Edge; hostname or full URL).
```
illiakovalenko:
```suggestion
# Optional: custom Sitecore Edge Platform hostname (hostname or full URL).
```

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
```diff
@@ -19,6 +19,10 @@ SITECORE_EDGE_CONTEXT_ID=
...
+# Available on both server and client.
```
illiakovalenko: This can be removed, as it is already expected behavior for Next.js developers when using NEXT_PUBLIC environment variables.
```suggestion
```

packages/create-content-sdk-app/src/templates/nextjs-app-router/.env.remote.example
```diff
@@ -19,6 +19,10 @@ SITECORE_EDGE_CONTEXT_ID=
 # Will be used as a fallback if separate SITECORE_EDGE_CONTEXT_ID value is not provided
 NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 
+# Optional: custom Experience Edge hostname override (XM Cloud/Edge; hostname or full URL).
+# Available on both server and client.
+NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME=
```
illiakovalenko: Please, see my comments for "nextjs" template as a references

packages/core/src/tools/normalize-env-value.ts
illiakovalenko, line 1: Please, add unit tests

illiakovalenko (changes requested):

CHANGELOG.md
```diff
@@ -15,9 +15,9 @@ Our versioning strategy is as follows:
 ### 🎉 New Features & Improvements
 
 * `[nextjs]` `[create-content-sdk-app]` Enable Next.js 16 Cache Components and Turbopack File System Caching ([#334](https://github.com/Sitecore/content-sdk/pull/334))
-  - Enabled `cacheComponents: true` for explicit caching with "use cache" directive
-  - Enabled `experimental.turbopackFileSystemCacheForDev: true` for faster dev startup (beta)
-  - Available in both Pages Router and App Router templates
+
+* `[core]` `[content]` `[nextjs]` Support custom Edge hostnames via `NEXT_PUBLIC_SITECORE_EDGE_HOSTNAME` ([#359](https://github.com/Sitecore/content-sdk/pull/359))
+  - Consolidated `rewriteContentUrls` and `contentRewrite` into `rewriteMediaUrls: boolean | ((value: string) => string)`. When `true`, uses default Edge host rewriter; when a function, transforms each string (SDK traverses layout). Migration: `rewriteContentUrls: true` → `rewriteMediaUrls: true`; custom rewriter → `rewriteMediaUrls: (value: string) => string`
```
illiakovalenko: Could we make sure to double check the AI-generated notes? It should be removed as it was renamed and consolidated during the PR review, initially we didn't have this feature.
We can simply mention the new property that was introduced

packages/content/src/client/edge-proxy.ts
```diff
@@ -1,30 +1,36 @@
...
- * @param {string} [sitecoreEdgeUrl] - The base endpoint URL for the Edge Platform. Default is https://edge-platform.sitecorecloud.io
+ * @param {string} [sitecoreEdgeUrl] - The base endpoint URL for the Edge Platform (resolved at config level). Defaults to platform URL.
  * @returns {string} The complete URL for accessing content through the Edge Platform.
  * @public
  */
-export const getEdgeProxyContentUrl = (sitecoreEdgeUrl = SITECORE_EDGE_URL_DEFAULT) =>
-  `${normalizeUrl(sitecoreEdgeUrl)}/v1/content/api/graphql/v1`;
+export const getEdgeProxyContentUrl = (
+  sitecoreEdgeUrl: string = constants.SITECORE_EXPERIENCE_EDGE_URL_DEFAULT
+) => `${getBaseEdgeUrl(sitecoreEdgeUrl)}/v1/content/api/graphql/v1`;
```
illiakovalenko, line 21: Let me clarify again the difference between Experience Edge and Edge Platform URLs, as there appears to be a misunderstanding.

Experience Edge (https://edge.sitecorecloud.io)
– Used for media URLs.

Edge Platform (https://edge-platform.sitecorecloud.io)
– Used by our services to resolve endpoints.

Based on the changes you introduced, Experience Edge is now being referenced when resolving service endpoints. This is not correct. We should continue using the default Edge Platform URL for service-related requests.

The Experience Edge URL should be referenced only when replacing media URLs.
This is the only place where the Experience Edge URL environment variable should be used, in order to allow overriding it when necessary.

packages/content/src/client/sitecore-client.ts
```diff
@@ -391,18 +395,20 @@ export class SitecoreClient implements BaseSitecoreClient {
...
+    // Use default Edge URL for styles (ignore custom hostname) so stylesheets load from platform
+    const edgeUrlForStyles = resolveEdgeUrlForStaticFiles();
```
illiakovalenko, line 405: Are stylesheets not supported with a custom hostname? Is this an existing limitation?

packages/create-content-sdk-app/src/templates/nextjs-app-router/.env.remote.example
```diff
@@ -19,6 +19,10 @@ SITECORE_EDGE_CONTEXT_ID=
 # Will be used as a fallback if separate SITECORE_EDGE_CONTEXT_ID value is not provided
 NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 
+# Optional: custom Sitecore Edge Platform hostname (hostname or full URL).
+# Next.js: use NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME for client; SITECORE_EDGE_PLATFORM_HOSTNAME for server-only.
```
illiakovalenko:
```suggestion
```

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
```diff
@@ -19,6 +19,10 @@ SITECORE_EDGE_CONTEXT_ID=
 # Will be used as a fallback if separate SITECORE_EDGE_CONTEXT_ID value is not provided
 NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 
+# Optional: custom Sitecore Edge Platform hostname (hostname or full URL).
+# Next.js: use NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME for client; SITECORE_EDGE_PLATFORM_HOSTNAME for server-only.
```
illiakovalenko:
```suggestion
```

packages/content/src/layout/rewrite-edge-host.ts
```diff
@@ -0,0 +1,190 @@
...
+ * Default Edge hostnames that may appear in layout/editing responses.
+ * Used when rewriting URLs to a custom hostname. Includes production and staging.
+ * @internal
+ */
+const DEFAULT_EDGE_HOSTNAMES = [
+  'edge-platform.sitecorecloud.io',
+  'edge.sitecorecloud.io',
+  'edge-staging.sitecore-staging.cloud',
+  'edge-platform-staging.sitecore-staging.cloud',
+] as const;
```
illiakovalenko: Please refer to my previous [comment](https://github.com/Sitecore/content-sdk/pull/359#discussion_r2798500820) - the issue has not been resolved yet

illiakovalenko (changes requested):

packages/content/src/client/sitecore-client.ts
MenKNas: When using a custom hostname and attempting to serve styles, it resulted in a 404 error, and the styles on the page appeared broken. So I kept it this way to make sure styles don't break in this case.

packages/content/src/client/sitecore-client.ts
```diff
@@ -623,12 +634,15 @@ export class SitecoreClient implements BaseSitecoreClient {
     // regular sitemap
     if (sitemapPath) {
       try {
+        const edgeUrl =
+          this.initOptions.api.edge.edgeUrl ?? constants.SITECORE_EDGE_PLATFORM_URL_DEFAULT;
```
illiakovalenko: Expected to be resolved on config level, right? As I mentioned in previous comments
```suggestion
          this.initOptions.api.edge.edgeUrl;
```

packages/content/src/client/sitecore-client.ts
```diff
@@ -701,6 +715,25 @@ export class SitecoreClient implements BaseSitecoreClient {
...
+   * @returns {LayoutServiceData} Rewritten layout (or same reference if rewrite disabled)
+   * @internal
+   */
+  protected applyContentRewrite(layout: LayoutServiceData): LayoutServiceData {
+    const opt = this.initOptions.rewriteMediaUrls;
+    if (!opt) {
+      return layout;
+    }
+    const edgeUrl =
+      this.initOptions.api.edge.edgeUrl ?? constants.SITECORE_EDGE_PLATFORM_URL_DEFAULT;
```
illiakovalenko: https://github.com/Sitecore/content-sdk/pull/359#discussion_r2820813635

packages/react/src/components/SitecoreProvider.tsx
```diff
@@ -86,7 +86,7 @@ export class SitecoreProvider extends React.Component<
         ...props.api,
         edge: {
           ...props.api.edge,
-          edgeUrl: constants.SITECORE_EDGE_URL_DEFAULT,
+          edgeUrl: resolveEdgeUrl(),
```
illiakovalenko: Can it be removed since it's expected to be resolved on config level now?

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -0,0 +1,97 @@
...
+ * @returns {string | undefined} The custom Edge URL if configured, undefined otherwise
+ * @public
+ */
+export function getCustomEdgeUrl(): string | undefined {
+  if (!hasCustomEdgeHostname()) {
+    return undefined;
+  }
+
+  return resolveEdgeUrl();
+}
```
illiakovalenko: Unused function

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -0,0 +1,97 @@
...
+  // Otherwise, treat as hostname and add https://
+  return normalizeUrl(`https://${trimmed}`);
+}
+
+/**
+ * Checks if a custom Edge hostname is configured via environment variables.
+ * @returns {boolean} True if a custom hostname is configured
+ * @public
+ */
+export function hasCustomEdgeHostname(): boolean {
```
illiakovalenko: Unused function

packages/nextjs/src/config/define-config.ts
```diff
@@ -19,7 +29,11 @@ export const getNextFallbackConfig = (config?: SitecoreConfigInput): SitecoreCon
         contextId: config?.api?.edge?.contextId || '',
         clientContextId:
           config?.api?.edge?.clientContextId || process.env.NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID,
-        edgeUrl: config?.api?.edge?.edgeUrl || process.env.NEXT_PUBLIC_SITECORE_EDGE_URL,
+        edgeUrl: resolveEdgeUrl(
+          config?.api?.edge?.edgeUrl ??
+            process.env[NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME_ENV] ??
+            process.env[SITECORE_EDGE_PLATFORM_HOSTNAME_ENV]
```
illiakovalenko: I suppose it's redundant since it's resolved on "core" level
```suggestion
            process.env[NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME_ENV]
```

packages/content/src/layout/rewrite-edge-host.ts
```diff
@@ -0,0 +1,176 @@
...
+ * @param {string} input - The string to escape
+ * @returns {string} The escaped string safe for RegExp construction
+ * @internal
+ */
+function escapeRegExp(input: string): string {
+  return input.replace(/[.*+?^${}()|[\]\\]/g, '\\$&');
+}
+
+/**
+ * Rewrites Edge Platform hostnames in a response object to use the custom hostname.
```
illiakovalenko:
```suggestion
 * Rewrites Experience Edge hostnames in a response object to use the custom hostname.
```

packages/content/src/layout/rewrite-edge-host.ts
```diff
@@ -0,0 +1,176 @@
...
+ * This function performs a deep traversal of the object and replaces any string values
+ * containing the default Edge hostnames with the custom hostname.
+ * Caller must pass the resolved Edge URL from config (no env resolution here).
```
illiakovalenko: Could you please clarify how this approach is expected to work for custom (non-production) environments?

At the moment, we don’t have a mechanism to define a dedicated Sitecore Experience Edge hostname via internal environment variable. Please refer to the related [comment](https://github.com/Sitecore/content-sdk/pull/359#discussion_r2798500820) for additional context.
Also [here](https://github.com/Sitecore/content-sdk/blob/f22711f1bd59e7c2199ec72338960bd8ca99653e/packages/content/src/client/sitecore-client.ts#L730) you are providing Edge Platform url that is not right based on discussion

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
```diff
@@ -22,6 +22,9 @@ NEXT_PUBLIC_SITECORE_EDGE_CONTEXT_ID=
 # Optional: custom Sitecore Edge Platform hostname (hostname or full URL).
 NEXT_PUBLIC_SITECORE_EDGE_PLATFORM_HOSTNAME=
 
+# Optional: custom Experience Edge hostname for media URL rewriting (e.g. staging).
+SITECORE_EXPERIENCE_EDGE_HOSTNAME=
```
illiakovalenko, line 26: This environment variable, as I mentioned previously, is **internal**, so we should assume that it will be populated by Deploy if custom hostname is configured
Can be removed from both templates

illiakovalenko (approved):
Looks good! 👍
Please, see one minor comment, then feel free to merge

packages/create-content-sdk-app/src/templates/nextjs/.env.remote.example
illiakovalenko: @MenKNas I’ve just revisited this and have a follow-up question.
Is it possible to keep this internal-only, or does it require explicit configuration from the customer side? In other words, can we encapsulate it within our implementation, or is customer-level access and setup unavoidable?

MenKNas: If customers need to use their own custom hostname for media, then they would need to populate it somewhere, so as our common practice, this usually happens through an env var. In my opinion, it's better to keep it in the example at least as optional, so people know it exists and it can be used. Otherwise, they would need to discover it from the code or the docs and add it/populate it manually. To me it's better DX to have it in the examples so people can understand easily how they can do this. But I am up for discussion if we think we don't want this.

PR #360, opened 2026-02-03
[Chore] Remove deprecated api's for v2.0 release

yavorsk:
- [x] Changelog updated
- [x] Upgrade guide entry added

## Description / Motivation

This PR removes deprecations and TODO's ahead of Content SDK v2.0 release. 
#### `core` package
- `enforceCors` function have been removed, `getEnforcedCorsHeaders` should be used instead;

#### `content` package
- the `config` property of `SitecoreCliConfigInput` type is now required
- the build time functions passed in the `build` array of `SitecoreCliConfigInput` now require passing `scConfig` (if `args` argument is passed at all)
- in accordance with the above `generateSites`, `writeImportMap` and `extractFiles` build functions have been updated; unit tests have been updated and fixed

#### `react` package
- `renderEmptyPlaceholder` method of `PlaceholderComponent` has been removed; `renderEmptyPlaceholder` from `placeholder-utils` should be used instead which is exported
- `DesignLibrary` component now does not accept any props, `DesignLibraryProps` type has been removed
- `SitecoreProvider`'s `loadImportMap` prop is now required

#### `next` package
- usage of `enforceCors` has been replaced with `getEnforcedCorsHeaders` in `feaas-render-middleware`; some comments have been updated

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate) - smoke test

## Types of changes


packages/react/src/components/DesignLibrary/models.ts
```diff
@@ -20,11 +20,9 @@ export type DynamicComponent = React.ComponentType<{
   params?: ComponentParams;
 }>;
 
-// @MAJOR-RELEASE-TODO - Make loadImportMap required in next major version
 export type DesignLibraryProps = {
   /**
    * The dynamic import for import map to be used in variant generation mode.
-   * Currently it's optional but it will be required in the next major version.
```
yavorsk, line 27: `loadImportMap` is still required to be optional, in case we use DesignLibrary for client side component in app router. In that case, `loadImportMap` is passed via SitecoreProvider and not as a prop.

packages/nextjs/src/index.ts
art-alexeyenko, line 1: Please make sure you remove FEaaSWrapper and BYOCWrapper exports from here.

CHANGELOG.md
```diff
@@ -43,6 +43,13 @@ Our versioning strategy is as follows:
     - Previously, `/da/source -> /target` rule would redirect to `/da/target` path when default locale is not `da`
     - Now, `/da/source -> /target` rule would redirect to `/target` path, using default locale, unless the `Shall language be preserved upon redirect?` checkbox is enabled in Redirect Map.
 * Upgrade to Node.js 24.x ([#332](https://github.com/Sitecore/content-sdk/pull/332))
+* Remove deprecated api's ([#360](https://github.com/Sitecore/content-sdk/pull/360]))
+    - `enforceCors` function have been removed, `getEnforcedCorsHeaders` should be used instead
+    - usage of `enforceCors` has been replaced with `getEnforcedCorsHeaders` in `feaas-render-middleware`
+    - the `config` property of `SitecoreCliConfigInput` type is now required
+    - the build time functions passed in the `build` array of `SitecoreCliConfigInput` now require passing `scConfig` (if `args` argument is passed at all)
```
illiakovalenko: It’s actually the opposite: the constructor no longer accepts the `scConfig` argument, as it was removed, but the current function implementation requires `scConfig` to be provided.

packages/content/src/tools/codegen/extract-files.ts
```diff
@@ -51,10 +47,8 @@ function _extractFiles(args: ExtractFilesConfig = {}) {
   };
   const renderingHost = process.env.SITECORE_RENDERINGHOST_NAME;
 
-  return async ({ scConfig }: { scConfig?: SitecoreConfig } = {}) => {
-    const config = args.scConfig ?? scConfig;
-
-    if (!config) {
+  return async ({ scConfig }: { scConfig: SitecoreConfig }) => {
+    if (!scConfig) {
```
illiakovalenko, line 51: As we discussed earlier, could we remove these additional checks? The config will always be provided, and the same applies to our other tools. Instead, we can check if "scConfig" is provided in "build" cli command, only in one place

packages/react/src/components/DesignLibrary/models.ts
illiakovalenko: Technically, the DesignLibrary component doesn’t accept any props - it is wrapped by an HOC, which provides the `loadImportMap` function. Therefore, the `DesignLibraryProps` type can be removed entirely.
However in [SitecoreProvider](https://github.com/Sitecore/content-sdk/blob/chore/JSS-8876/packages/react/src/components/SitecoreProvider.tsx#L25-L27) we have to mark `loadImportMap` as required

packages/content/src/tools/codegen/test-data/extract-components/named-imports/package.json
```diff
@@ -0,0 +1 @@
+{}
```
illiakovalenko:
```suggestion
{}

```

packages/content/src/tools/codegen/test-data/extract-components/with-client-map/package.json
```diff
@@ -0,0 +1 @@
+{}
```
illiakovalenko:
```suggestion
{}

```

CHANGELOG.md
```diff
@@ -43,6 +43,13 @@ Our versioning strategy is as follows:
     - Previously, `/da/source -> /target` rule would redirect to `/da/target` path when default locale is not `da`
     - Now, `/da/source -> /target` rule would redirect to `/target` path, using default locale, unless the `Shall language be preserved upon redirect?` checkbox is enabled in Redirect Map.
 * Upgrade to Node.js 24.x ([#332](https://github.com/Sitecore/content-sdk/pull/332))
+* Remove deprecated api's ([#360](https://github.com/Sitecore/content-sdk/pull/360]))
```
illiakovalenko, line 46: What about the personalize scope - will it be removed? If not, let’s remove this [note](https://github.com/Sitecore/content-sdk/blob/chore/JSS-8876/packages/content/src/personalize/personalize-service.ts#L22)

CHANGELOG.md
```diff
@@ -43,6 +43,13 @@ Our versioning strategy is as follows:
...
+    - `enforceCors` function have been removed, `getEnforcedCorsHeaders` should be used instead
+    - usage of `enforceCors` has been replaced with `getEnforcedCorsHeaders` in `feaas-render-middleware`
```
illiakovalenko: Can be removed, since this is our internal implementation

CHANGELOG.md
illiakovalenko: Suggestion to simplify the description:
```
* `sitecore.cli.config`: 
    - The `config` property is now required; a reference to `sitecore.config` must be provided.
    - Build-time functions defined in the build array no longer accept the scConfig argument in their constructor. However, the function implementation must receive scConfig, which is passed internally by the CLI command.
```

illiakovalenko (changes requested):
Looks good! Please see my suggestions and comments below

CHANGELOG.md
yavorsk: yep, we'll leave it and will remove the note

packages/content/src/tools/codegen/extract-files.ts
yavorsk: I decided leaving the scConfig checks to the responsible functions seems more reasonable. and keep the build function clean, just to execute them.

packages/nextjs/src/index.ts
yavorsk: as discussed separately - we'll leave the exports to support pages router apps

illiakovalenko (approved):
👍

PR #363, opened 2026-02-04
Refine other analytics packages

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation

- Update error messages
- Change all reference of browserId to clientId
- Make all analytics packages public
- Update generate-docs scripts
- Update the object that is injected to window
- Rename environments to adapters
- Add api-extractor to the analytics packages
- Update JSDocs
- Rename internal variables

## Testing Details

- [ ] Unit Test Added

## Types of changes


packages/analytics-core/src/client-id/construct-get-client-id-url.spec.ts
```diff
@@ -0,0 +1,13 @@
+import { SITECORE_EDGE_URL } from '../consts';
+import { constructGetClientIdUrl } from './construct-get-client-id-url';
+import { expect } from '@jest/globals';
+
+describe('constructGetClientIdUrl', () => {
```
illiakovalenko: I think it was proposed to use `resolve` instead of `construct`, what do you think about renaming?

packages/analytics-core/src/initialization/browser-adapter.ts
```diff
@@ -0,0 +1,73 @@
...
+ */
+export interface AnalyticsBrowserAdapter extends AnalyticsAdapter {
+  /**
+   * The type of the adapter.
+   */
+  type: 'browser';
+}
+
+/**
+ * Enables analytics in the browser.
```
illiakovalenko: It does not enable analytics in the browser, rather, it provides an entity (adapter). Please expand the description and purpose to reflect this more accurately

packages/analytics-core/src/initialization/const.ts
```diff
@@ -1,2 +1,5 @@
+/**
+ * The name of the Analytics plugin.
+ * @public
```
illiakovalenko, line 3:
```suggestion
 * @internal
```

packages/analytics-core/src/initialization/types.ts
```diff
@@ -1,33 +1,94 @@
...
   };
 }
 
-export interface AnalyticsSettings {
-  cookieSettings: {
-    name: {
-      browserId: string;
-    };
+/**
+ * Defines the structure of the resolved analytics plugin options, including cookie settings and proxy values.
```
illiakovalenko: To me, this description looks overengineered - let’s simplify it a bit
For example
"Defines options for the Analytics plugin."

packages/analytics-core/src/initialization/types.ts
```diff
@@ -1,33 +1,94 @@
...
 
+/**
+ * Defines the structure of the analytics plugin, including its initialization method, name, options, and adapter.
+ * @public
+ */
 export interface AnalyticsPlugin extends Plugin {
-  settings: AnalyticsSettings;
-  environment: AnalyticsEnvironment;
+  /**
+   * The resolved options for the analytics plugin.
```
illiakovalenko: They are not resolved - they are simply passed through by the developer.
```suggestion
   * Options for the analytics plugin.
```

packages/analytics-core/src/interfaces.ts
```diff
@@ -9,13 +10,24 @@ export interface EPResponse {
...
 }
 
-export interface ProxySettings {
-  browserId: string;
-  guestId: string;
+/**
+ * The resolved visitor IDs returned from the Edge Proxy.
+ * @public
+ */
+export interface ResolvedVisitorIds {
```
illiakovalenko: Suggestion to follow common practices: it is usually unnecessary to explicitly mention that something is “resolved,” as all values are resolved at some point
```suggestion
export interface VisitorIds {
```

packages/analytics-core/src/interfaces.ts
```diff
@@ -9,13 +10,24 @@ export interface EPResponse {
   customer_ref: string;
 }
 
-export interface ProxySettings {
-  browserId: string;
-  guestId: string;
+/**
+ * The resolved visitor IDs returned from the Edge Proxy.
```
illiakovalenko:
```suggestion
 * The visitor IDs returned from the Edge Proxy.
```

packages/analytics-core/src/interfaces.ts
```diff
@@ -1,5 +1,6 @@
 /**
- * The response object that Sitecore EP returns.
+ * The response object that Sitecore Edge Proxy returns.
+ * @internal
```
illiakovalenko: This is not truly an internal type, as it is used as a return type in the public API. We should rename it accordingly. The new name needs to reflect the specific API or product being referenced, as “Edge Proxy” is too generic.

packages/core/src/initialization/init-content-sdk.ts
```diff
@@ -0,0 +1,50 @@
...
+  for (const plugin of plugins) {
+    coreContext.plugins.set(plugin.name, plugin);
+  }
+
+  debug.init(`Registered ${coreContext.plugins.size} plugins`);
+  coreContext.readyPromise = initPlugins(coreContext.plugins);
+
+  await coreContext.readyPromise;
+
+  debug.init('SDK initialization complete');
```
illiakovalenko:
```suggestion
  debug.init('Content SDK initialization complete');
```

packages/events/src/events/custom-event/custom-event.ts
```diff
@@ -55,29 +58,31 @@ export class CustomEvent extends BaseEvent {
...
     const fetchBody = Object.assign({}, this.customEventPayload, baseAttr);
 
-    return await this.sendEvent(fetchBody, this.settings);
+    return await this.sendEvent(fetchBody, this.config);
   }
 }
 
 /**
  * Interface of the unified arguments object for custom event
+ * @internal
```
illiakovalenko, line 75: If developer can directly use "CustomEvent" it means that this interface should be "public".
Also, please, add tag for "CustomEvent" class

packages/events/src/events/custom-event/event.ts
```diff
@@ -3,27 +3,28 @@ import { getAnalyticsPlugin } from '@sitecore-content-sdk/analytics-core/interna
...
+import { getCoreContext } from '@sitecore-content-sdk/core';
 import { getEventsPlugin } from '../../initialization/plugin';
 
 /**
- * A function that sends an event to SitecoreCloud API with the specified type
- * @param {EventData} eventData - The required/optional attributes in order to be send to SitecoreCloud API
- * @returns The response object that Sitecore EP returns
+ * A function that sends an event to the SitecoreCloud API with the specified type
+ * @param {EventData} eventData - The required/optional attributes to be sent to the SitecoreCloud API
+ * @returns The response object that Sitecore Edge Proxy returns
```
illiakovalenko, line 12: It doesn’t clearly indicate what is actually returned. Can we describe it using more precise terms? Please review all similar occurrences.

packages/events/src/initialization/const.ts
```diff
@@ -1,2 +1,5 @@
+/**
+ * The name of the Events plugin.
+ * @public
```
illiakovalenko, line 3:
```suggestion
 * @internal
```

illiakovalenko (changes requested):
Good work! See my comments and suggestions below

packages/analytics-core/src/client-id/construct-get-client-id-url.spec.ts
sc-nikolaoslazaridis: For this case i think it doesn't apply.

packages/analytics-core/src/initialization/const.ts
sc-nikolaoslazaridis: Since the analytics plugin interface is public itself we have to have this one also as public.

packages/analytics-core/src/initialization/types.ts
sc-nikolaoslazaridis: For this case it is not exactly how the dev passed them. For example we add the cookie name.

packages/analytics-core/src/interfaces.ts
sc-nikolaoslazaridis: This is a suggestion from Akos. If you insist ill change it.

packages/analytics-core/src/client-id/construct-get-client-id-url.spec.ts
illiakovalenko: This is also a good candidate, just to maintain consistency

packages/events/src/events/custom-event/custom-event.ts
sc-nikolaoslazaridis: The developer **can't** use this.

packages/events/src/initialization/const.ts
sc-nikolaoslazaridis: Same as the previous comment

packages/analytics-core/src/initialization/const.ts
illiakovalenko: In this case you need to export it from "index" but it's exported only from "internal" submodule. So technically it means that it's still internal. What do you think?

packages/analytics-core/src/initialization/types.ts
illiakovalenko: But it still describes initial options passed by dev, isn't it?

packages/analytics-core/src/interfaces.ts
illiakovalenko: Yeah, as we discussed - let's update it, to follow common practices

packages/events/src/events/custom-event/custom-event.ts
illiakovalenko: I see, in this case it's expected to keep it "internal"

packages/analytics-core/src/initialization/const.ts
sc-nikolaoslazaridis: From my perspective we need to keep the index clean for the normal users which wont ever need the plugin names. For people that want to develop custom implementations of the adapters i think it is ok to keep it in the internal submodule. Thoughts?

illiakovalenko (approved):

PR #364, opened 2026-02-08
Fix behaviour of renderEach in page editing mode

liamhubers:
- [ ] Changelog updated

## Description / Motivation
The `renderEach` attribute on the `Placeholder` and `AppPlaceholder` show weird behaviour during editing mode in the Page Builder. This is because all of the children of the placeholder are wrapped in the `PlaceholderMetadata` component before being returned from the `getRenderedComponents` function. The result is only a single wrapper is being added when there are more than one children in the placeholder.

The fix is to call the `renderEach` inside the `getRenderedComponents` function instead of on the result. As a side effect, you can now also combine `render` and `renderEach` where before `renderEach` was ignored when you pass a `render` function to the `Placeholder`.

## Testing Details

- [x] Unit Test Added

### Before:
image

### After:
image

## Types of changes

### DIFF
--- a/packages/react/src/components/Placeholder/AppPlaceholder.test.tsx
+++ b/packages/react/src/components/Placeholder/AppPlaceholder.test.tsx
@@ -1080,6 +1080,52 @@ describe('App Placeholder logic', () => {
     });
   });
 
+  it('should use renderEach for each child in the placeholder when page editing is enabled', () => {
+    const page = getPage();
+    const components = new Map<string, React.FC>();
+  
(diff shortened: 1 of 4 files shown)


illiakovalenko:
Closed as it was already included in PR #367. Thank you for the contribution!

art-alexeyenko:
And to add to this, thank you @liamhubers for pointing out this issue and proposing a solution.
It's a great coincidence, but it came at the perfect time as we were revisiting the `react` package as a whole.

PR #367, opened 2026-02-16
[react] Refactor components and HOCs

art-alexeyenko:
- [ ] Changelog updated - TBD

## Description / Motivation
Refactors react package with simplicity and best React practices in mind.
- Convert all built in components to functions
- Make them use React hooks where appropriate
- Use JSX logic instead of React.createElement
- Major Placeholder, AppPlaceholder rewrite: 
  - align logic between the two
  - convert Placeholder to function component
  - rework the logic of passing props to components, only pass down what's neccessary
  - introduce `passThroughProps` prop to add flexibility in passing props to child components
  - fix `renderEach` (https://github.com/Sitecore/content-sdk/pull/364) Thanks @liamhubers
  - general refactoring
- Remove `withPlaceholder` and replace it with `withClientPlaceholder` - a clientside/SSR analogue to `withAppPlaceholder`
- Add a `useComponentMap` hook to be used instead of `withComponentMap` when needed.
- FEAAS/BYOC logic remains unchanged for compatibility with `components` logic

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/react/src/components/Placeholder/AppPlaceholder.tsx
```diff
@@ -21,124 +18,67 @@ import { rsc } from '#rsc-env';
  * @public
  */
 export const AppPlaceholder = (props: AppPlaceholderProps) => {
```
stasmaxymov: `AppPlaceholder` was designed to work universally in App router and Pages router, with only new requirement - have `componentMap` and `page` provided as required props.
Can we make `Placeholder` just as wrapper over `AppPlaceholder`? Smth like:
```
const Placeholder = (props)=>{
  const { componentMap, page } = useSitecore();
  const appProps = {
     ...props,
     componentMap
     page
  }
  return <AppPlaceholder {...appProps} />
}
```
If not, what behavior we expect `Placeholder` to have that is not present in `AppPlaceholder`?

packages/react/src/components/Placeholder/placeholder-utils.tsx
```diff
@@ -282,3 +256,95 @@ export const getComponentForRendering = (
     isEmpty: false,
   };
 };
+
+export const drawPlaceholderComponents = (
```
stasmaxymov: This function accepts some props and returns back JSX, why won't you make it as React component?
It also accepts `drawPlaceholderChildComponent` function as parameter, which is also accepts some props  and returns JSX, so it is could also be a component passed as parameter. But I think it is possible to make it as regular childern prop

sc-nikolaoslazaridis (changes requested):
Lets try to keep the code base error and warning free

Errors: 
  - packages/react/src/enhancers/withAppPlaceholder.tsx

Warnings:
  - packages/react/src/components/SitecoreProvider.tsx
  - packages/react/src/components/DesignLibrary/DesignLibrary.tsx

packages/react/src/components/Placeholder/AppPlaceholder.tsx
```diff
@@ -21,124 +18,67 @@ import { rsc } from '#rsc-env';
...
+    ) : (
+      <componentForRendering.component
+        key={key}
+        {...renderedProps}
+        {...props.passThroughComponentProps}
+        page={props.page}
+        componentMap={props.componentMap}
+      />
+    );
+  };
```
sc-nikolaoslazaridis: Since the props of the AppPlaceholder can cover both client and server cases i would suggest to consolidate to one component instead of having both line 39 and 46. With that we would can pass the props directly to the component and remove the internal usage of hooks (making the destinction for the client components).

packages/react/src/components/Placeholder/AppPlaceholder.tsx
sc-nikolaoslazaridis: If all changes applied this can become something like a "UniversalPlaceholder" and the reexport it from the Nextjs package as "AppPlaceholder" or another better name.

packages/react/src/components/SitecoreProvider.tsx
```diff
@@ -65,68 +65,68 @@ export const ImportMapReactContext = React.createContext<
```
sc-nikolaoslazaridis, line 83: Based on the usage i believe we need to remove the extra contexts by moving the logic to the SitecoreProviderReactContext and SitecoreProvider. That way we wont have the extra overhead of creating those and improve code readability.

packages/react/src/enhancers/withSitecore.tsx
sc-nikolaoslazaridis: Since we have the provider and the useSitecore hook this function is redundant.

sc-nikolaoslazaridis: Usually the context hooks live within the same file the Providers. Since we have the SitecoreProvider, can you pls move it there?

sc-nikolaoslazaridis: Based on my previous comment on the provider now the context should provide all the extra thing that are provided in the contexts to be deleted (ex. ImportMapReactContext).

packages/react/src/enhancers/withComponentMap.tsx
```diff
@@ -18,9 +19,9 @@ export function withComponentMap<T extends WithComponentMapProps>(
    * @returns {JSX.Element} - the rendered component
    */
   function WithComponentMap(props: T): JSX.Element {
```
sc-nikolaoslazaridis: Following the same pattern with the other comments, we need to delete this one.

packages/react/src/enhancers/withComponentMap.tsx
```diff
@@ -29,3 +30,13 @@ export function withComponentMap<T extends WithComponentMapProps>(
 
   return WithComponentMap;
 }
+
+/**
+ * Hook to access the component map in client context.
+ * @returns {ComponentMap} The component map from the SitecoreProvider
+ * @public
+ */
+export function useComponentMap(): ComponentMap {
```
sc-nikolaoslazaridis: Following the same pattern, we can move the functionality to the sitecore context.

packages/react/src/enhancers/withLoadImportMap.tsx
sc-nikolaoslazaridis: Following the same pattern, we can move the functionality to the sitecore context.

sc-nikolaoslazaridis: Following the same pattern with the other comments, we need to delete this one.

packages/react/src/enhancers/withClientPlaceholder.tsx
```diff
@@ -0,0 +1,41 @@
...
+  page: Page;
+  componentMap: ComponentMap;
+};
+
+/**
+ * Provides a slot-like functionality by wrapping a component and rendering placeholders defined in the layout data.
+ * @param {ComponentType<T>} Component - The component to be wrapped around placeholders.
+ * @returns {React.ReactNode} A new component that renders the original component with placeholders.
+ */
+export const withClientPlaceholder = <T extends ComponentProps, W extends T & WrapperProps>(
```
sc-nikolaoslazaridis: The current naming of "Client" and "App" placeholders are confusing. The "App" sounds like it a whole app wrapper, but the case here is different. We need to either keep the old name "withPlaceholder" or rename to "withAppPlaceholder" to something different.

sc-nikolaoslazaridis (changes requested):
We went through my comments with @illiakovalenko and thought that these are some quick wins we can do to improve the codebase. If you find any other places that can have the same refactoring please do apply them. 

Also there is this file packages/react/src/enhancers/withEditorChromes.tsx that still returns a class component. Can you rewrite it also?

 Feel free to ping me if have any questions.

packages/react/src/enhancers/withSitecore.tsx
art-alexeyenko: I agree about the other HOCs, but I'd like to keep this one for several reasons:
- Some things in React still cannot be achieved with function components only (like error boundary)
- our ErrorBoundary depends on it
- Leaving this will make customers' life easier as they can still easily inject sitecore context into their class components (if they have any). React recommends moving to functions, of course, but the migration is a decision that should be coming from customers, not be forced by us.

sc-nikolaoslazaridis: Based on what i see, the places that we use the ErrorBoundary we already can have the sitecore data. That means we can directly pass that as props to the component. Example of that is the AppPlaceholder, it already has whatever we need to be passed down to the ErrorBoundary.

art-alexeyenko: It would be safer to always retrieve `page` from context in ErrorBoundary and other client components. Most of the fields in `page` are serializable, but we also have `clientData` in `LayoutServiceContext`:
```
clientData?: Record<string, Record<string, unknown>>;
```
That could throw a wrench into things.

sc-nikolaoslazaridis: withSitecore is only for client side usage same as the ErrorBoundary. That means no matter what the places you will use it, you can simply call useSitecore in the parent component and pass down the props or already have the data needed like in our cases.

art-alexeyenko: Yeah, which is more work during upgrade for people already using it, which is why I listed it as one of the reasons to leave it:
> Leaving this will make customers' life easier

I do agree it needs to go, eventually, so the middle ground I'd propose is mark it `deprecated` now and remove come 3.0.

packages/react/src/components/Placeholder/placeholder-utils.tsx
art-alexeyenko: We return not so much a component as a collection of components. Some logic then relies on this collection being, well, a collection. I renamed this method instead, for better description of intent.

packages/react/src/components/Placeholder/AppPlaceholder.tsx
art-alexeyenko: Please check the current approach

packages/react/src/components/Placeholder/AppPlaceholder.tsx
art-alexeyenko: not something we can do here, I think. props that can be passed to client components (the wrapper) and server components (componentForRendering.component) differ in RSC - we err on the side of caution here.

packages/react/src/components/Placeholder/AppPlaceholder.tsx
art-alexeyenko: We still need separate placeholder exports for client and server contexts - but this latest refactor should streamline the code. Please take a look

packages/react/src/enhancers/withClientPlaceholder.tsx
art-alexeyenko: I reverted to the `withPlaceholder` name, agree it was confusing

packages/react/src/components/RichText.tsx
sc-nikolaoslazaridis: Can we remove the usage of forwardRef based on this doc https://react.dev/reference/react/forwardRef? Same comment applies all the places we use it.

packages/react/src/components/Date.tsx
sc-nikolaoslazaridis, line 51: Double nested enhancers is hard to follow, can we simplify this? Maybe create a component instead of an enhancer so it can be used directly in JSX to wrap the current component or even a hook to get the end compoent?

packages/react/src/components/Placeholder/Placeholder.tsx
```diff
@@ -1,172 +1,35 @@
...
+    props.name,
+    page.mode.isEditing
+  );
+  const isEmpty = !placeholderRenderings.length;
+  useEffect(() => {
+    if (isEmpty && PagesEditor.isActive()) {
       PagesEditor.resetChromes();
     }
-  }
+    // eslint-disable-next-line react-hooks/exhaustive-deps
```
sc-nikolaoslazaridis, line 28: Are we sure we want to disable the lint rule? We could refactor it by introducing a ref for the isEmpty variable.

art-alexeyenko: We need to run this effect only once. Both ref and the empty dependency array can acheive that, it all comes down to style preference I suppose

packages/react/src/components/Date.tsx
art-alexeyenko: done, this and the one above

art-alexeyenko:
Moved to https://github.com/Sitecore/content-sdk/pull/371

PR #380, opened 2026-02-26
[Angular CSDK] Scaffold package, sample

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Basic angular package and sample as base for Angular Content SDK implementation.
- Draft package structure:
   - scaffolded as angular library
   - relevant re-exports from core, content
   - set up unit testing
   - draft field components and directives that can be replaced or refined later
- draft angular sample
  - configured for SSR
  - placeholder 404, 500 pages
  - draft components
  - has general folder structure to be used in future
- `core/tools` exports that are safe for client-side are decoupled from the client-unsafe exports to ensure angular app bundles okay (note: this will be adjusted based on the outcome of https://github.com/Sitecore/content-sdk/pull/383).

## Testing Details
- [~] Unit Test Added

## Types of changes


packages/core/src/tools/index.ts
illiakovalenko, line 1: This does not look like the best approach - simply moving all utilities into core feels too broad
Are we certain that this logic actually needs to be decoupled? What is the exact scenario that reproduces the issue?
If decoupling is really required, a cleaner option would be to introduce a /node-tools submodule and place all CLI- and server-related code there. That would keep responsibilities clear and avoid overloading index

illiakovalenko (changes requested):
Can you please simplify the pull request first?

As agreed earlier, we only need to add an empty package and an application template at this stage. It should not include any components, services, or types for now - those will be introduced as part of a separate work item.

You can review the related tasks for additional context.

packages/angular/src/lib/components/placeholder/placeholder.component.ts
```diff
@@ -0,0 +1,81 @@
...
+ * <sc-placeholder name="jss-main" [rendering]="route" />
+ * @public
+ */
+@Component({
+  selector: 'sc-placeholder',
+  standalone: true,
+  imports: [NgComponentOutlet, FieldMetadataComponent],
+  template: `
+    @for (r of renderings(); track r.uid || $index) { @if (isEditing()) {
+    <sc-field-metadata [rendering]="r">
```
stasmaxymov: such setup creates an intermediate `sc-field-metadata` html element around each rendering. I don't know if it is problem, but it is a difference to a React implementation

packages/angular/src/lib/components/sitecore-provider/sitecore-provider.component.ts
```diff
@@ -0,0 +1,47 @@
...
+ *   <router-outlet />
+ * </sc-sitecore-provider>
+ * @public
+ */
+@Component({
+  selector: 'sc-sitecore-provider',
+  standalone: true,
+  template: '<ng-content />',
+})
+export class SitecoreProviderComponent {
```
stasmaxymov: What is the practical benefit of it over just calling `this.contextService.setPage(this.page());` in that top component?
Providers concept is react thing not angular thing

packages/angular/src/lib/components/placeholder/placeholder.directive.ts
```diff
@@ -0,0 +1,126 @@
...
+ * when there are no renderings and the page is in editing mode.
+ * @example
+ * <ng-container
+ *   [scPlaceholder]="route"
+ *   scPlaceholderName="jss-main"
+ * ></ng-container>
+ * @public
+ */
+@Directive({
+  selector: '[scPlaceholder]',
```
stasmaxymov: you can add multiple selectors e.g one as attribute and another as component (and also convert it to component to use template. Cause it is typically not expected thing for an attribute directive to change inner html of an eelement it is applied to

packages/create-content-sdk-app/src/templates/angular/src/app/lib/sitecore-context.service.ts
```diff
@@ -0,0 +1,79 @@
...
+import { Page, RouteData } from '@sitecore-content-sdk/angular';
+/**
+ * Service that provides access to the current Sitecore page context.
+ * Use this service to set the current page data and access it throughout the application.
+ * @public
+ */
+@Injectable({
+  providedIn: 'root',
+})
+export class SitecoreContextService {
```
stasmaxymov: looks like a duplicate to a similarly named one in the package

packages/angular/src/lib/components/placeholder/placeholder.component.ts
stasmaxymov: We can use our old approach where a metadata component renders a single node, opened or closed depending on input parameters

packages/core/src/tools/index.ts
illiakovalenko: Also keep in mind that we can't just provide these changes in "minor" release

packages/angular/src/lib/components/placeholder/placeholder.component.ts
art-alexeyenko: as per discussions with Illia - the components and services are reverted for now and will be included later, with your feedback applied

sc-nikolaoslazaridis (changes requested):
Can you please do all the changes regarding the node-tools in the other PR. Also i noticed that in the next js package the name here is server-tools, i think we need to follow the same naming to avoid any possible mistake to use those in an invalid runtime and introduce bugs.

packages/angular/package.json
```diff
@@ -0,0 +1,83 @@
+{
+  "name": "@sitecore-content-sdk/angular",
+  "version": "2.0.0-canary.6",
+  "description": "Angular components and utilities for Sitecore Content SDK",
+  "sideEffects": false,
+  "publishConfig": {
```
sc-nikolaoslazaridis, line 7: Can we keep it private to be safe and avoid publishing it by mistake?

packages/angular/README.md
```diff
@@ -0,0 +1,59 @@
+# MyWorkspace
```
sc-nikolaoslazaridis: Do we need to keep the info in here as is?

packages/angular/.vscode/mcp.json
```diff
@@ -0,0 +1,9 @@
+{
```
sc-nikolaoslazaridis: Let's remove this for now and add it if we need to in the future (i think this is still experimental)

packages/angular/.vscode/tasks.json
```diff
@@ -0,0 +1,42 @@
+{
```
sc-nikolaoslazaridis: Is this useful for our workflow?

packages/angular/src/public-api.ts
```diff
@@ -0,0 +1,54 @@
+// ─── Types ─────────────────────────────────────────────────────────────────
```
sc-nikolaoslazaridis: I would say lets keep this file clean and add features per task and only if needed.

packages/create-content-sdk-app/src/templates/angular/README.md
```diff
@@ -0,0 +1,59 @@
+# AngularSample
```
sc-nikolaoslazaridis: Lets clean up this file and provide a proper readme when the time comes

packages/create-content-sdk-app/src/templates/angular/src/sitecore.config.ts
```diff
@@ -0,0 +1,10 @@
+// This file is server-side only.
```
sc-nikolaoslazaridis: Out of scope, lets keep this PR as clean as possible

sc-nikolaoslazaridis (changes requested):
- The lib folder in the angular package has some out of scope things, please remove those
- Please create a helloWorld function to the package and use it in the template to ensure proper integration

packages/angular/.vscode/tasks.json
art-alexeyenko: not really - removed

packages/angular/README.md
art-alexeyenko: removed readme in package and adjusted the one in sample

sc-nikolaoslazaridis (approved):

PR #386, opened 2026-03-01
[react][nextjs]: Add Content-Type header for editing/render api in Pages Router

MenKNas:
- [x] Changelog updated

## Description / Motivation
This PR sets an explicit `Content-Type: text/html; charset=utf-8` header on the Pages Router editing render route response.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other

## Types of changes

### DIFF
--- a/packages/nextjs/src/editing/editing-render-middleware.ts
+++ b/packages/nextjs/src/editing/editing-render-middleware.ts
@@ -222,6 +222,7 @@ export class EditingRenderMiddleware extends RenderMiddlewareBase {
         route,
       });
 
+      res.setHeader('X-Content-Type-Options', 'nosniff');
       res.status(200).send(html);
     } catch (err) {
       const error = err as Record<string, unknown>;
--- a/packages/nextjs/src/editing/editing-render-middleware.ts
+++ b/packages/nextjs/src/editing/editing-render-middleware.ts
@@ -222,7 +222,7 @@ export class EditingRenderMiddleware extends RenderMiddlewareBase {
         route,
       });
 
-      res.setHeader('X-Content-Type-Options', 'nosniff');
+      res.setHeader('Content-Type', 'text/html; charset=utf-8');
       res.status(200).send(html);
     } catch (err) {
       const error = err as Record<string, unknown>;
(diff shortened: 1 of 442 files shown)


packages/react/src/components/DesignLibrary/DesignLibraryClientEvents.test.tsx
```diff
@@ -140,7 +140,6 @@ describe('<DesignLibraryClientEvents />', () => {
             message: {
               status: DesignLibraryStatus.READY,
               uid: testEditedComponent.uid,
-              isRenderingServerComponent: true,
```
illiakovalenko, line 143: Could you clarify why updates are required in the DesignLibrary unit test?

packages/react/src/tests/jsdom-setup.ts
```diff
@@ -36,3 +36,20 @@ global.jsdom = jsdom;
 
 global.HTMLElement = jsDomWindow.HTMLElement; // makes chai "happy" https://github.com/chaijs/chai/issues/1029
 copyProps(jsDomWindow, global);
+
+// Suppress React error-boundary test noise (expected errors that React logs to console.error)
+const originalConsoleError = global.console.error;
```
illiakovalenko: I assume this was added to suppress additional console errors. However, the errors thrown in unit tests are expected, so we shouldn’t be suppressing them intentionally

packages/nextjs/src/route-handler/editing-render-route-handler.ts
```diff
@@ -255,6 +255,7 @@ export const createEditingRenderRouteHandlers = (options: EditingHandlerOptions)
       });
 
       responseHeaders['Content-Type'] = 'text/html; charset=utf-8';
+      responseHeaders['X-Content-Type-Options'] = 'nosniff';
```
illiakovalenko: Your task is about adding a header to the render endpoint in the Pages Router, whereas in the App Router this behavior already works as expected

illiakovalenko (changes requested):

packages/nextjs/src/editing/editing-render-middleware.ts
```diff
@@ -222,6 +222,7 @@ export class EditingRenderMiddleware extends RenderMiddlewareBase {
         route,
       });
 
+      res.setHeader('X-Content-Type-Options', 'nosniff');
```
illiakovalenko: The X-Content-Type-Options: nosniff header may be configured by the customer. To guarantee proper response handling, we must explicitly include the correct Content-Type header, we don't need to set "X-Content-Type-Options"

illiakovalenko (changes requested):

illiakovalenko (approved):

PR #387, opened 2026-03-02
[content] [react] [nextjs] Improve import map generation and refactor Design Library error handling

yavorsk:
- [X] Changelog updated

## Description / Motivation

Improve import map generation and refactor Design Library error handling:
- fix issue, where Import Map generation incorrectly identifies a client component as server component in case its 'use client' directive is preceded by comments or eslint rules
- Refactor error handling in Design Studio by introducing more error types and additional more general event name
   - send `component:generation:component-preview-error` event only when we want to trigger retry component generation from the AI.
   - send the more general `component:generation:error` when the error can not be fixed by the AI to prevent unnecessary retries.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/content/src/tools/codegen/import-map.ts
```diff
@@ -357,7 +357,8 @@ const prepImportMaps = async (paths: string[], separateMaps?: boolean): Promise<
     // read the start of the file that may be 'use client'
     const firstLine = await new Promise<string>((resolve) => {
       let readBuffer = '';
-      const stream = fs.createReadStream(fullPath, { end: 12 });
+      // Read first 1KB - enough for any reasonable comments/rules before 'use client'
+      const stream = fs.createReadStream(fullPath, { end: 1024 });
```
stasmaxymov: do we really need to 1k buffer heuristics? Are the any concerns to read file fully?

packages/content/src/tools/codegen/import-map.ts
```diff
@@ -357,7 +357,8 @@ const prepImportMaps = async (paths: string[], separateMaps?: boolean): Promise<
     // read the start of the file that may be 'use client'
     const firstLine = await new Promise<string>((resolve) => {
```
stasmaxymov: this is not a firstLine anymore as it contains multiple of them

packages/content/src/tools/codegen/import-map.ts
yavorsk: i had none other than performance reasons - to me seams to be a reasonable middle ground between 'read only first line of each file' and 'read everything of each file' and regex match the whole thing. what do you think?

yavorsk: as discussed separately - will update this part to read the whole file, to be on the safe side

packages/content/src/tools/codegen/import-map.test.ts
```diff
@@ -602,6 +602,55 @@ ${defaultMapTemplate(indexedImportMap)}`;
       expect(clientCall.args[1]).to.not.match(/test-exports/);
     });
 
+    it('should correctly determine client import maps when _use client_ directive is not on first line', async () => {
```
art-alexeyenko, line 605: Could you please add a couple more unit tests to verify possible false positives? For example, it should not return true when:
- use client is within a comment (// 'use client';)
- use client string is a part of a function call (console.log('use client');
- some other cases perhaps?

packages/content/src/editing/codegen/preview.ts
```diff
@@ -166,13 +171,29 @@ export interface DesignLibraryComponentPropsEvent extends DesignLibraryEvent {
...
     error: unknown;
-    type: DesignLibraryPreviewError;
+    type: DesignLibraryPreviewError.Render | DesignLibraryPreviewError.RenderInit;
+  };
+}
+
+/**
+ * Represents an event indicating a general error to be sent to design library.
+ */
+export interface DesignLibraryErrorEvent extends DesignLibraryEvent {
```
art-alexeyenko, line 192: Please mark this and related exports as `@internal` so that api coverage can start tracking them.

art-alexeyenko (approved):
Great job, thank you. Leaving a couple comments - but they are straightforward to implement. Please take a look and merge once addressed.

PR #406, opened 2026-03-12
[Angular CSDK] Loaders system

art-alexeyenko:
## Description / Motivation
Adds a loader system for Angular SSR setup in Content SDK. For context, see POC documentation.
- Loader system is designed to allow for the app to retrieve logic from XMCloud/other sources in both client and server contexts
- Loaders are provided via loader registry in DI
- Loaders can return data, redirects or throw errors
- Loader error handling will redirect to either `notFound` or `error` route depending on the error thrown
  - loader resolver will return the redirect command directly on browser or let handleNavigationError to process it on server (navigation error is not produced on browser)
  - notFound and error routes provided via CSDK configuration
- LoaderResolver + LoaderDataService ensure loader data is available on both server and browser via TransferState and cache.
- LoaderDataServiceMiddleware allows loader data processing in browser - by executing loaders and sharing results via an expressJS endpoint
- loader prefetch logic moved to a separate task to get proper attention and implementation that doesn't suck

## Testing Details
- [x] Unit Test Added


packages/angular/src/loaders/loader-data.service.ts
```diff
@@ -0,0 +1,130 @@
...
+
+  /**
+   * Get data for the given request, using cache or fetching if needed.
+   * If a request is already pending for this URL/loader combination,
+   * waits for it to complete instead of making a duplicate request.
+   * Consumes (removes) cached data after retrieval.
+   * @param {LoaderDataRequest} request - The loader data request
+   * @returns {Promise<LoaderApiResponse>} Promise resolving to the API response
+   */
+  async getData(request: LoaderDataRequest): Promise<LoaderApiResponse> {
```
stasmaxymov, line 72: I might be missing it, but I can't find a service that enables parallel loaders fetching.

packages/angular/src/loaders/loader-data.service.ts
```diff
@@ -0,0 +1,130 @@
...
+    // Only fetch in browser
+    if (!isPlatformBrowser(this.platformId)) {
+      return { kind: 'error', status: 500, message: 'LoaderDataService only works in browser' };
+    }
+
+    const key = cacheKey(request.loaderId, request.url);
+
+    // Return cached response if available (consume on use); supports data and redirect
+    const cached = this.cache.get(key);
```
stasmaxymov, line 81: If there is no parallel fetching, like it is right now as it seems, this cache is going to return a stale data for each second client side navigation to the same request

packages/angular/src/loaders/loader-data.service.ts
```diff
@@ -0,0 +1,130 @@
...
+      query: request.query ?? {},
+    };
+    console.log('DEBUG: LoaderDataService fetchData', endpoint, reqBody);
+
+    const fetchPromise = firstValueFrom(
+      this.http.post<LoaderApiResponse>(endpoint, reqBody, {
+        cache: 'no-store',
+      })
+    )
+      .then((resp) => {
```
stasmaxymov: this `fetchData` function is already marked with `async`, but having multiple .then here makes it harder to read

packages/angular/src/server/loader-data-service-middleware.ts
```diff
@@ -0,0 +1,176 @@
...
+ */
+async function executeLoader(
+  request: LoaderApiRequest,
+  loaders: LoaderRegistry,
+  requestContext?: RequestContext
+): Promise<LoaderApiResponse> {
+  const { loaderId, url, params, query } = request;
+
+  const loader = loaders[loaderId];
+  if (!loader) {
```
stasmaxymov, line 36: If we can't find a loader by ID, it feels like misconfiguration, or asynchronization between client and server, so it more a 500 error with some additional logging

packages/angular/src/loaders/loader-resolver.ts
```diff
@@ -0,0 +1,141 @@
...
+import { redirectOnNavigationError } from './router-error-handling';
+import { ERROR_ROUTE_TOKEN, NOT_FOUND_ROUTE_TOKEN } from '../lib/tokens';
+
+/**
+ * Create a state key for the loader
+ * @param {string} loaderId - The loader ID
+ * @param {string} url - The URL
+ * @returns {StateKey} The state key
+ */
+function stateKey(loaderId: string, url: string) {
```
stasmaxymov, line 30: I remember I had some issues with url in keys when tried to request a non existing page. It was like on client it we requesting a wrong key, miss a cache and request it on a client again. Please check it doesn't happen to you

packages/angular/src/loaders/loader-data.service.ts
art-alexeyenko: I removed the prefetch service as the prefetch part grew into its own separate task. That led to inadvertently removing the parallel preload part too.
Brought it back, with another name and a logic change. Instead of preloading all loaders on all routes, we preload current route + ancestors.

packages/angular/src/loaders/loader-data.service.ts
art-alexeyenko: good catch, you're right. fixed

packages/angular/src/loaders/router-error-handling.ts
```diff
@@ -0,0 +1,86 @@
...
+ * when the failed navigation was already to the not-found route or the error route.
+ * Must be called from an injection context (uses NOT_FOUND_ROUTE_TOKEN, ERROR_ROUTE_TOKEN, Router).
+ *
+ * **HTTP status codes (SSR):** RedirectCommand only triggers navigation to the not-found or error
+ * route; it does not set the HTTP response status. To return 404 or 500 when those pages are
+ * rendered on the server, configure your app so the server sends the correct status. For example,
+ * in `app.routes.server.ts` add ServerRoute entries for your not-found and error paths with
+ * `status: 404` and `status: 500` (see Angular "Setting headers and status codes" in the SSR guide).
+ * Alternatively, inject `RESPONSE_INIT` in your NotFoundComponent and ErrorComponent and set the
+ * status when running on the server.
```
stasmaxymov, line 28: I'm not necessarily agree with this statement. We know when we need to return 404 status - it server page rendering and we even know how to do it (via Response_init)

packages/angular/src/loaders/router-error-handling.ts
```diff
@@ -0,0 +1,86 @@
...
+  const notFoundPath = normalizePath(notFoundRoute);
+  const errorPath = normalizePath(errorRoute);
+
+  if (kind === 'notFound') {
+    if (failedPath === notFoundPath) {
+      console.log('RouteErrorHandler: Not found route was not found. Avoiding redirect loop.');
+      return;
+    }
+    const urlTree = router.parseUrl(notFoundRoute);
+    return new RedirectCommand(urlTree);
```
stasmaxymov, line 58: don't we need a `skipLocationChange`. I would assume that the default behavior of angular is to return 302 on `RedirectCommand` when page is rendered on server.

PR #408, opened 2026-03-18
[content] [react] [atoms] Introduce Atom Registry utils

yavorsk:
- [ ] Changelog updated

## Description / Motivation

Introduce Atom Registry utilities `DesignLibraryAtoms` and `AtomRenderer`
- introduce atom registry utilities to unfold atom metadata array and serialize atom metadata
- add atom:registry event to `content`
- add basic implementation of DesignLibraryAtoms which acts as AtomRender wrapper in Design Library mode and handles communication with hosting application
- basic AtomRenderer component which would render a component comprised of atoms based on component layout data received by Design Studio and available atoms
- include `atoms` property in SitecoreProvider
- add `isAtomsMode` to PageMode
- for app template - add atoms folder and index file and export empty atoms array; pass it to SitecoreProvider

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/content/src/editing/atoms-builder/atoms-builder.ts
```diff
@@ -0,0 +1,88 @@
...
+/**
+ * Event to send import map to design library
+ */
+const DESIGN_LIBRARY_ATOM_REGISTRY_EVENT_NAME = 'atom:registry';
+
+/**
+ * Represents the type of atom, which can be either a top-level atom or an atom child
+ * @internal
+ */
+export enum AtomType {
```
stasmaxymov: My opinion that we should use string unions instead of enums everywhere. No runtime baggage, easier parsing from external input data (like from string). No problems with forgetting to export them from core packages

packages/react/src/atoms/atom-registry-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+      type: atom.type,
+      description: atom.description,
+      props: z.toJSONSchema(atom.props, { target: 'draft-7' }),
+      children,
+      defaultChildren,
+      htmlEvents: atom.htmlEvents,
+      customEvents,
+    });
+
+    for (const ch of atom.allowedChildren ?? []) {
```
stasmaxymov, line 61: There is no need to do recursive walk, as `unfoldAtomsRegistry` already did that.

packages/react/src/atoms/atom-registry-utils.ts
```diff
@@ -0,0 +1,95 @@
...
+    let customEvents: Record<string, unknown> | undefined = undefined;
+    if (atom.customEvents && Object.keys(atom.customEvents).length > 0) {
+      const eventsShape: Record<string, z.ZodTuple> = {};
+      for (const [name, args] of Object.entries(atom.customEvents)) {
+        eventsShape[name] = z.tuple(args as [z.ZodType, ...z.ZodType[]]);
+      }
+      customEvents = z.toJSONSchema(z.object(eventsShape), { target: 'draft-7' });
+    }
+
+    atomInfos.push({
```
stasmaxymov, line 49: I don't see allowedCHildren, this info will be needed for DS to control rules for adding atoms. Makes sure to have it in a serializable format

packages/react/src/components/DesignLibrary/DesignLibraryAtoms.tsx
```diff
@@ -0,0 +1,37 @@
...
+ * - Fetches Component model data, and passes it to the `AtomRenderer` which is responsible for rendering the low code component
+ * based on component model data and the available atoms.
+ * @internal
+ */
+export const DesignLibraryAtoms = () => {
+  const { atoms } = useSitecore();
+
+  useEffect(() => {
+    if (atoms) {
+      const flattenedAtoms = unfoldAtomsRegistry(atoms);
```
stasmaxymov: we don't necessarily need a flattening as a separate function, it can be called inside serialized atoms.
Or at least we need to change how it works, as now it still keeps all the references via allowedChildren, and defaultChildren. To be the real unfold function it should change those properties to return just component names not full objects

stasmaxymov:
I think we also need some new mode on /api/editing/render endpoint. Like library-atoms or smth like that, so DS can enable it

stasmaxymov (changes requested):
Overall looks good, see some comments

packages/react/src/atoms/atom-registry-utils.ts
yavorsk: 'children'  contains allowed children. i will rename it to be clearer

packages/react/src/components/DesignLibrary/DesignLibraryAtoms.tsx
yavorsk: ill remove it

packages/react/src/atoms/atom-registry-utils.ts
yavorsk: i removed `unfoldAtomRegistry`

PR #409, opened 2026-03-18
[Chore] v2.0 release prep

art-alexeyenko:
- Graduate all canary package versions
- Add corresponding git tags
- Adjust changelog
- Sets the ground for applying the changeset release flow after release
- Prepares the repo for npm release publish


packages/nextjs/package.json
yavorsk: update the peer deps as well

yavorsk (approved):
just update peer deps versions

PR #412, opened 2026-03-23
[chore] Reverse changeset release flow

art-alexeyenko:
- Reverts changeset flow
  - lerna does not update peerDependencies for canary or regular release. Was not a problem, until monorepo packages are added as peers. Needs another approach.
  - old lerna canary flow is tied to canary versions. Updating a single package creates a canary version for that package - but the rest of monorepo is not not using that canary, affecting testing. An alternative, snapshot canary approach is needed.

- applies canary versions for old flow


yavorsk (approved):
Looks good.

PR #414, opened 2026-03-24
Lightweight Tracking

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Implements usage tracking aligned with a standardized Visit metric for Sitecore AI billing. The SDK captures customer interactions where possible, with explicit handling for bots so they are measured and billed correctly (not as normal human visits).

#### Bot detection & behavior
* Detection: Evaluate the User-Agent per request using known bot patterns via the `isbot` library (e.g. Googlebot, GPTBot, and other crawlers).
* Prefetch: Prefetch requests are excluded from this counting path so they are not treated as visits.
* Client-side page views: Do not fire normal client pageView for bots that execute JavaScript; bot traffic is handled on the server so analytics stay consistent with the Visit definition.
* Bot tracking proxy runs early in the chain, sets a bot cookie for downstream logic, and triggers server-side bot page view where appropriate.
* Local / dev: Bot tracking is skipped in development and on loopback hosts so local testing does not pollute usage data.
* Host parsing: Host / x-forwarded-host values are normalized (including IPv6 with ports) so loopback and site resolution behave correctly.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/events/src/events/page-view/bot-detection.ts
```diff
@@ -0,0 +1,45 @@
...
+};
+
+/**
+ * A function that gets the bot cookie.
+ * @returns {string | undefined} The value of the bot cookie, or undefined if the cookie is not found.
+ * Only available on the client-side.
+ * @internal
+ */
+export function getBotCookie(): string | undefined {
+  if (typeof document === 'undefined') return undefined;
```
sc-nikolaoslazaridis: Can you use the function you created in line 21?

packages/events/src/index.ts
```diff
@@ -4,6 +4,7 @@ export type { EventAttributesInput, ExtensionData } from './events/common-interf
 export { form } from './events/custom-event/form';
 export { event } from './events/custom-event/event';
 export { identity } from './events/identity/identity';
-export { pageView } from './events/page-view/page-view';
+export { pageView, botPageView } from './events/page-view/page-view';
+export { isBot, BOT_DETECTION_COOKIE } from './events/page-view/bot-detection';
```
sc-nikolaoslazaridis: Maybe move those to the internal module since the main purpose is to use tho only in our packages?

packages/nextjs/src/proxy/bot-tracking-proxy.ts
```diff
@@ -0,0 +1,150 @@
...
+  };
+
+  /**
+   * @param {NextRequest} req - Incoming request
+   * @returns True when bot tracking should be skipped for a local / dev environment.
+   * @internal
+   */
+  protected shouldSkipForLocalEnvironment(req: NextRequest): boolean {
+    // Allow bot tracking in local environment for development purposes
+    if (process.env.SITECORE_ENABLE_BOT_TRACKING === 'true') {
```
sc-nikolaoslazaridis, line 127: Since we need bot tracking to be true by default, i'm thinking to reverse the name and make it "SITECORE_DISABLE_BOT_TRACKING". Thoughts?

packages/events/src/events/page-view/bot-detection.ts
```diff
@@ -0,0 +1,45 @@
...
+  return typeof document !== 'undefined';
+}
+
+/**
+ * A function that checks if visitor is a bot.
+ * @param {string} userAgent - The user agent of the visitor
+ * @returns {boolean} True if the visitor is a bot, false otherwise
+ * @public
+ */
+export const isBot = (userAgent?: string | null): boolean => {
```
sc-nikolaoslazaridis, line 31: Since we need to use the bot detection/cookie also inn the personalize proxy maybe it would be better if we moved the bot detection to the analytics core package.

packages/events/src/events/page-view/page-view.ts
```diff
@@ -5,15 +5,63 @@ import type { PageViewData } from './page-view-event';
...
+};
+
+/**
+ * Sends a VIEW event for server-side bot tracking (e.g. Next.js proxy / Edge).
+ * Uses a synthetic per-invocation client id and defaults `channel` to `bot`.
+ * Returns `null` in browser environments.
+ * @returns The response from Sitecore Edge Proxy, or `null` if skipped (browser).
+ * @public
+ */
+export async function botPageView(): Promise<EPResponse | null> {
```
sc-nikolaoslazaridis: Please move to a separate file

packages/events/src/events/page-view/bot-detection.spec.ts
```diff
@@ -0,0 +1,78 @@
...
+    });
+  });
+
+  describe('isBrowserEnvironment', () => {
+    it('is true in jsdom where document is defined', () => {
+      expect(isBrowserEnvironment()).toBe(true);
+    });
+  });
+
+  describe('getBotCookie', () => {
```
art-alexeyenko, line 61: could you add the test for "returns undefined when in server context"? A small thing, but a useful sanity check

packages/nextjs/src/proxy/bot-tracking-proxy.ts
```diff
@@ -0,0 +1,150 @@
...
+              adapter: analyticsProxyAdapter(req, res),
+            }),
+            eventsPlugin(),
+          ],
+        });
+
+        await botPageView();
+      };
+
+      res.cookies.set(BOT_DETECTION_COOKIE, '1', {
```
art-alexeyenko, line 96: did you consider making this cookie HttpOnly to ensure client-side js doesn't tamper with it?

packages/nextjs/src/proxy/bot-tracking-proxy.ts
```diff
@@ -0,0 +1,150 @@
...
+     * If not provided, the bot tracking will run synchronously.
+     * Read more about `fetchEvent` in the [Next.js documentation](https://nextjs.org/docs/app/api-reference/file-conventions/proxy#waituntil-and-nextfetchevent)
+     * @param {NextFetchEvent} fetchEvent - Fetch event to run the bot tracking in the background.
+     */
+    fetchEvent?: NextFetchEvent;
+    /**
+     * When `false`, bot tracking is disabled for every request.
+     * Default `true`. Local runs (`next dev` or loopback host) still skip automatically.
+     */
+    enabled?: boolean;
```
art-alexeyenko: we already have the `skip` option in the ProxyBaseConfig - what do you think about leveraging it for the `enabled/disabled` switch and the special cases like forcing execution via SITECORE_ENABLE_BOT_TRACKING, streamlining the current approach of multiple ways bot tracking proxy can be turned on or off ?

art-alexeyenko (changes requested):

packages/nextjs/src/proxy/bot-tracking-proxy.ts
illiakovalenko: This one was added just for us for development purpose in order to avoid the case when bot tracking is disabled locally, this env var is not expected to be used by customers and will not be exposed anywhere

packages/events/src/events/page-view/bot-detection.ts
illiakovalenko: I don’t think we need to move it to analytics-core, since it’s still specific and relates to the events domain.
It’s currently used in the personalize proxy to check cookies, and it could also be reused in other proxies if needed for optimization.

packages/nextjs/src/proxy/bot-tracking-proxy.ts
illiakovalenko: The `skip` flag can’t fully replace the local development check. Even if `skip` returns `false`, we should still prevent this functionality from running in local or preview environments.
I updated the implementation to use a `skip` flag instead of an `enabled` flag for greater flexibility, especially to cover potential edge cases customers might encounter.
The environment variable was introduced for internal local development purposes only (only for us, since we need to test it locally as well) - it won’t be exposed to customers or documented.

packages/nextjs/src/proxy/bot-tracking-proxy.ts
illiakovalenko: It should be exposed by design and remain accessible in the browser, since some bots are able to execute JavaScript.

The client-side code in @sitecore-content-sdk/events reads this cookie via getBotCookie() using document.cookie (see page-view.ts / bot-detection.ts). If the cookie is set as HttpOnly, it becomes inaccessible to JavaScript, which would break this flow - the client-side pageView() would no longer detect the bot flag set.

As a result, we would end up sending page views from both the server and the client. However, for bots, the intended behavior is to send the page view only from the server (via the proxy).

art-alexeyenko (approved):

PR #417, opened 2026-03-28
[Angular] Refine loaders logic, add env, config support + default loaders

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
- Adjust sitecore.config flow for Angular
- Add process.env handling logic for Angular (work around process.env not being available in angular apps)
- Logically separate angular native logic (/app) and content-sdk integrations (/content-sdk) in the sample. This demonstrates the areas that fully support angular conventions and the areas that may not
- Add default loaders

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/create-content-sdk-app/src/templates/angular/scripts/generate-environment.ts
```diff
@@ -0,0 +1,82 @@
...
+  for (const key of keys) {
+    if (key === 'NODE_ENV') {
+      continue;
+    }
+    if (key === 'production') {
+      continue;
+    }
+    const value = process.env[key] ?? '';
+    lines.push(`  ${JSON.stringify(key)}: ${JSON.stringify(value)},`);
+  }
```
MenKNas, line 39: Is this serialization approach safe? Won't this be shipped to Angular's client bundle, possibly exposing internal variables?

packages/core/src/tools/resolve-edge-url.ts
```diff
@@ -47,8 +47,10 @@ export function resolveEdgeUrl(edgeUrl?: string): string {
     return normalizeUrl(explicit);
   }
 
+  const runtimeEnv = process ? process.env : {};
```
MenKNas, line 50: It's possible that `process ? process.env : {}` can still throw a `ReferenceError` in runtimes where process is undefined.

I think it would be safe to type it like `typeof process !== 'undefined'` before accessing process.env

packages/create-content-sdk-app/src/templates/angular/scripts/generate-environment.ts
```diff
@@ -0,0 +1,82 @@
...
+
+const projectRoot = process.cwd();
+const angularEnvDir = resolve(projectRoot, 'src', 'environments');
+
+function loadEnvFiles(mode: 'dev' | 'prod'): void {
+  const contextEnvFile = mode === 'dev' ? '.env.dev' : '.env.prod';
+
+  // Base .env first — does NOT override existing process.env values
+  const baseEnvPath = resolve(projectRoot, '.env');
+  const localEnvPath = resolve(projectRoot, '.env.local');
```
MenKNas, line 24: Should we add `env.local` to `.gitignore` like the rest env files?

packages/angular/src/lib/providers.ts
```diff
@@ -27,6 +41,14 @@ export interface SitecoreAngularConfig {
 export function provideSitecoreAngular(config: SitecoreAngularConfig): EnvironmentProviders {
   const providers = [];
 
+  if (config.sitecoreConfig !== undefined) {
+    providers.push({ provide: SITECORE_CONFIG_TOKEN, useValue: config.sitecoreConfig });
+
+    providers.push({
+      provide: SITECORE_CLIENT_TOKEN,
+      useFactory: () => new SitecoreClient({ ...config.sitecoreConfig! }),
```
MenKNas: I might be missing context, but it looks like we end up with two `SitecoreClient` instances: one from `provideSitecoreAngular (SITECORE_CLIENT_TOKEN)` and one from template `getClient()`. Is that intentional? Isn't there a way to unify the two different approaches?

MenKNas (approved):

MenKNas:
Great work @art-alexeyenko! Thank you

PR #421, opened 2026-04-01
Add redirect handled callback

mjarosch:
Add a redirect handled callback so that analytics can be collected for redirects. This is a POC implementation of https://github.com/Sitecore/content-sdk/discussions/420.

- [ ] Changelog updated

## Description / Motivation

1. Added an optional redirectHandled parameter to RedirectsProxyConfig
2. Added a HandledRedirectInfo type to represent the handled redirect that is passed to the callback
3. Added a check to the dispatchRedirect method to call the redirectHandled callback if provided. It uses the Vercel [after](https://nextjs.org/docs/app/api-reference/functions/after) function so that it doesn't block the proxy 

## Testing Details

Added a unit test to validate that callback is called when supplied.

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/nextjs/src/proxy/redirects-proxy.test.ts
+++ b/packages/nextjs/src/proxy/redirects-proxy.test.ts
@@ -12,7 +12,7 @@ import {
 } from '@sitecore-content-sdk/content/site';
 import chai, { use } from 'chai';
 import chaiString from 'chai-string';
-import { NextRequest, NextResponse } from 'next/server';
+import nextjs, { NextRequest, NextResponse } from 'next/server';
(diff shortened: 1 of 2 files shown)


sc-nikolaoslazaridis:
Hello @mjarosch, thank you for putting together this PoC, it really helped us validate the direction we want to take.
We created an internal ticket for this new feature and the result of that is this PR: https://github.com/Sitecore/content-sdk/pull/490. The implementation is different but it should serve your needs. You can find an example of how to use it in the PR description. I will close this PR and if you have any other feedback please use the following discussion https://github.com/Sitecore/content-sdk/discussions/420.

PR #432, opened 2026-04-07
[Angular CSDK] Basic rendering

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
Basic angular site rendering.
- Placeholder + utils
- field components and directives
- static component-map
- basic skate-park setup for easier testing

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/angular/src/field-directives/sc-link.directive.ts
```diff
@@ -0,0 +1,106 @@
...
+        }
+      } else {
+        this.renderer.removeAttribute(element, 'target');
+      }
+
+      const hasChildren = element.childNodes.length > 0 && element.textContent?.trim();
+      if (this.showLinkTextWithChildrenPresent() || !hasChildren) {
+        const text = link.text || link.href || '';
+        if (!hasChildren) {
+          this.renderer.setProperty(element, 'textContent', text);
```
MenKNas: Why do we need this block if we already allow `!hasChildren` from the initial condition?

Maybe we could add something like this to make it clearer in general?

```
const hasChildren = element.childNodes.length > 0 && element.textContent?.trim();
const text = link.text || link.href || '';

// Handle the case where children are missing first (the fallback)
if (!hasChildren) {
  this.renderer.setProperty(element, 'textContent', text);
  return; // Exit early since we've handled the empty state
}

// Handle the case where children are present and the flag is true
if (this.showLinkTextWithChildrenPresent()) {
  //  Logic for when children exist 
}
```

packages/angular/src/field-directives/sc-text.directive.ts
```diff
@@ -0,0 +1,50 @@
...
+ *
+ * @public
+ */
+@Directive({
+  selector: '[scText]',
+  standalone: true,
+})
+export class ScTextDirective {
+  /** The Sitecore text field. */
+  readonly scText = input.required<Field<string | number> | undefined>();
```
MenKNas: How can this be required and possibly `undefined()` at the same time?

packages/angular/src/components/sc-form.component.ts
```diff
@@ -0,0 +1,112 @@
...
+    ></div>
+  `,
+})
+export class ScFormComponent {
+  readonly rendering = input<ComponentRendering>();
+  readonly params = input<{ [key: string]: string }>({});
+  readonly fields = input<{ [key: string]: unknown }>({});
+
+  @ViewChild('formContainer', { static: true })
+  private formContainerRef!: ElementRef<HTMLDivElement>;
```
MenKNas, line 40: Can't we also write this as a signal? I think we have done it in other places

`private readonly formContainerRef = viewChild<ElementRef<HTMLDivElement>>('formContainer');`

packages/create-content-sdk-app/src/templates/angular/scripts/generate-environment.ts
```diff
@@ -27,7 +27,7 @@ function loadEnvFiles(mode: 'dev' | 'prod'): void {
 }
 
 function serialize(production: boolean): string {
-  const lines: string[] = [`  production: ${production},`];
+  const lines: string[] = [`  production: ${JSON.stringify(String(production))},`];
```
MenKNas: Cant this just be?

```
const lines: string[] = [`  production: "${production}",`] 
````

MenKNas:
@art-alexeyenko Really great work! See some minor suggestions for improvements.

MenKNas (approved):

packages/angular/src/components/sc-form.component.ts
art-alexeyenko: For this one, `viewChild` signal doesn't have a static option - and since we access the Ref from an event handler we need to ensure the ref is present. 
So I'd leave the old syntax here

PR #433, opened 2026-04-07
[nextjs] Support redirect items redirects

art-alexeyenko:
- [x] Changelog updated

## Description / Motivation
Content SDK supports rendering maps from SXA out of the box.
This PR adds support for language-versioned redirect items on top of it.

**Important**: This should be merged after Edge adjusts the item redirect results. Do not merge until then

## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other - redirects work correctly locally, both from maps and items

## Types of changes
- [~] Bug fix (non-breaking change which fixes an issue)


packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -301,15 +301,47 @@ export class RedirectsProxy extends ProxyBase {
...
+  protected matchFromRedirectMapRedirect(
+    redirects: RedirectResult[],
+    urlLocale: string,
+    incomingURL: string,
+    incomingQS: string
+  ): RedirectResult | undefined {
+    const nonLocaleRedirects = redirects.filter((redirect: RedirectResult) => !redirect.locale);
     let matchedQueryString: string | undefined;
-    const localePath = `/${locale.toLowerCase()}${normalizedPath}`;
+    const normalizedPath = incomingURL.replace(/\/*$/gi, '').toLowerCase();
```
sc-nikolaoslazaridis, line 340: Do we need to add the case insensitive option to the regex replace? We are not checking any alphabetic characters.

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -362,15 +394,40 @@ export class RedirectsProxy extends ProxyBase {
...
+  ): RedirectResult | undefined {
+    // locale patterns will include a redirect item suffix which we need to remove
+    const localeRedirects = redirects.filter((redirect) => redirect.locale === locale);
+    // locale rules are easy and nice
+    return localeRedirects.length
+      ? localeRedirects.find((redirect: RedirectResult) => {
+          const patternPath = redirect.pattern.replace(/\/*$/gi, '').toLowerCase();
+          return patternPath === currentPath;
+        })
+      : undefined;
```
sc-nikolaoslazaridis: We iterating over the array 2 times, maybe we can do something like this to avoid that (just an example, you can beautify it or even potentially totally skip the matchRedirectItemRedirect method and do it in place where it was called):

```suggestion
   return redirects.find((redirect) => redirect.locale === locale && redirect.pattern.replace(/\/*$/g, '').toLowerCase() === currentPath} );
```

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -301,15 +301,47 @@ export class RedirectsProxy extends ProxyBase {
     const { pathname: incomingURL, search: incomingQS = '' } = this.normalizeUrl(
       req.nextUrl.clone()
     );
-    const locale = this.getLanguage(req);
+    // locale of current request
+    const requestLocale = this.getLanguage(req);
+    // locale of the url - if present in the url
+    const urlLocale = req.nextUrl.locale;
```
sc-nikolaoslazaridis: Do we need both of these? It seems like the `getLanguage` has multiple fallbacks and one of those is the nextUrl.locale

art-alexeyenko: We do - requestLocale is for versioned redirect items , urlLocale is for redirect maps. It could be demonstrated better though - so I adjusted the code

sc-nikolaoslazaridis (approved):
Looks good, good job!

PR #434, opened 2026-04-08
[Chore] Better changesets release flow

art-alexeyenko:
## Description / Motivation
- Applies multi-version release support for monorepo via changesets
- Introduces a new canary release flow via snapshots
- All version updates and publishes are now handled by changesets, instead of lerna CLI

There are certain complications of the current canary flow makes multi-version support harder:
- Pushing canary version updates into both npm and updating package.json versions in GitHub
   - full releases need to have caret (^x.x.x) dependencies, canary releases need to use exact canaries
   - internal packages are present in peerDependencies - and lerna does not bump those on canary updates, causing errors
- managing separate canary versions within GitHub is hard to track

As an answer to these challenges, new, simplified canary flow:
- Does not update package versions to canary in GitHub
- Only publishes snapshot canary releases to npm
- Only publishes snapshot canary releases for the packages that are pending full release (i.e. packages that have changesets + create-content-sdk so the canary can be tested)
- NPM canary release is tagged with `canary` (dev branch, regular release flow) or `hotfix-canary` (release/ branches, hotfix release flow)
- Additionally, canary release of create-content-sdk is tagged with `canary.commit-{SHA}` and `canary.pr-{origin_pr_id}` for easier identification

This approach preserves pre-release publishes for QA and preview, and simplifies the overall flow.
Note: should be merged after next full release for all packages in monorepo.

## Testing Details
- [ ] Unit Test Added

## Types of changes
- [x] Chore


.github/workflows/publish.yml
```diff
@@ -70,94 +71,111 @@ jobs:
...
-          if [ -n "$(git status --porcelain)" ]; then
-            git commit -a -m "Bump canary version [skip ci]"
-            git push origin HEAD:$MAIN_BRANCH
-          else
-            echo "No version changes to commit"
-          fi
+      - name: Build packages
+        run: npm run build
+
+      - name: Creating .npmrc
```
sc-nikolaoslazaridis, line 125: (Optional) Maybe we can create a job that will create the file so we don't have the same block 2 times

.github/workflows/hotfix_release_prep.yml
```diff
@@ -0,0 +1,104 @@
...
+      package_name:
+        description: 'Full package name'
+        required: true
+        type: choice
+        options:
+          - '@sitecore-content-sdk/nextjs'
+          - '@sitecore-content-sdk/cli'
+          - '@sitecore-content-sdk/react'
+          - '@sitecore-content-sdk/search'
+          - 'create-content-sdk'
```
sc-nikolaoslazaridis: Is this the correct name to be displayed?

sc-nikolaoslazaridis (approved):
Looks good, we can revaluate after the first run

.github/workflows/hotfix_release_prep.yml
art-alexeyenko: good catch, as it was not

illiakovalenko (approved):

PR #438, opened 2026-04-15
Lightweight Tracking: Add user agent for lightweight tracking

MenKNas:
- [x] Changelog updated

## Description / Motivation
 This PR extends Bot VIEW data by including user agent value as `sourceUserAgent` in the event for improved lightweight tracking.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/packages/nextjs/src/proxy/bot-tracking-proxy.ts
+++ b/packages/nextjs/src/proxy/bot-tracking-proxy.ts
@@ -91,6 +91,7 @@ export class BotTrackingProxy extends ProxyBase {
         await botPageView({
           page: req.nextUrl.pathname,
           language: language,
+          userAgent: req.headers.get('user-agent') ?? '',
         });
       };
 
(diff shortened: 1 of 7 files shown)


packages/nextjs/src/proxy/bot-tracking-proxy.ts
```diff
@@ -91,6 +91,7 @@ export class BotTrackingProxy extends ProxyBase {
         await botPageView({
           page: req.nextUrl.pathname,
           language: language,
+          userAgent: req.headers.get('user-agent') ?? '',
```
sc-nikolaoslazaridis: Do we actually need to fallback to empty string? The isBot function will check for empty strings and will return false immediately so there is no way we end up here without a value. Maybe we can use some other default ("unknown") or assume everything is ok and we don't need the fallback. Same applies to the implementation of the botPageView. (if we need a fallback we can have it in one place, in the function itself)

sc-nikolaoslazaridis (approved):
Good job!

PR #445, opened 2026-04-22
Search Component: pass CDP session id in the queries to the search service

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
- Used the current `getClientId` function from analytics core package as the new `sessionId` for the search request

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/core/src/tools/try-catch.ts
```diff
@@ -0,0 +1,33 @@
+/**
+ * Executes a sync or async function and returns a tuple of [result, error].
+ * If the function returns a Promise, the result is a Promise resolving to the tuple.
+ * On success: returns [result, null]. On error: returns [fallback, Error].
+ * @param {() => CallbackResult} fn - The function to execute.
+ * @param {Fallback} [fallback] - Optional fallback value returned as the result when an error occurs.
+ * @returns {Result} A tuple of [result, error].
+ * @public
+ */
+export function tryCatch<
```
illiakovalenko: Good idea, however not sure we should introduce `tryCatch` as a new shared utility and especially 'public' one.
* We save a couple of lines, at the cost of an import and a tuple convention readers have to remember ([value, error] here).
* The signature is () => T, so any future caller with arguments has to wrap in a lambda: tryCatch(() => doThing(x, y), fallback) - which is visually on par with an inline try/catch.
* We use standard try/catch consistently across the codebase, and it's clean and readable as-is. We're not going to rewrite the existing statements, so this utility will likely stay almost unused. I'd prefer not to add something that doesn't pay for itself and forces a specific convention

illiakovalenko (changes requested):
Looks good 👍
Please see the one conceptual comment below

packages/core/src/tools/try-catch.ts
sc-nikolaoslazaridis: I will convert it to a private one. I find that this improves the readability and avoids the creation of variables without values and then set the value inside the try catch. Besides that, we don't have to go and change the whole codebase. If the team likes it, we can change the code as we work on it and only in the cases that makes sense.

illiakovalenko (approved):

PR #461, opened 2026-05-03
[Angular CSDK] Basic rendering finalize

art-alexeyenko:
- Adjust Tailwind integration a little
- Fix styles and style application via component props
- Add dedicated router link directive
- Miscellaneous fixes for Angular template

* Most of template logic will be moved out repo and transformed into separate sample app closer to 1.0


packages/angular/src/components/sc-form.component.ts
```diff
@@ -44,16 +43,20 @@ export class ScFormComponent {
   private readonly destroyRef = inject(DestroyRef);
 
   readonly styles = () => {
-    const s = this.params()?.['styles'];
+    const p = { ...this.rendering()?.params, ...this.params() };
```
illiakovalenko: Looks like this path is only partially covered by tests.
Current specs validate `rendering.params` usage, but I don’t see a case where `params` input is provided directly and merged with `rendering.params`.
Could we add a unit test for direct params input (and ideally precedence when both are set)?

packages/angular/src/components/sc-form.component.ts
```diff
@@ -25,15 +25,14 @@ const { executeScriptElements, loadForm, subscribeToFormSubmitEvent } = form;
...
   selector: 'sc-form',
-  standalone: true,
   template: `
     <div #formContainer [class]="styles()" [id]="renderingId()"></div>
   `,
 })
 export class ScFormComponent {
   readonly rendering = input<ComponentRendering>();
-  readonly params = input<{ [key: string]: string }>({});
   readonly fields = input<{ [key: string]: unknown }>({});
```
illiakovalenko: `fields` input is defined but not used

packages/angular/src/field-directives/link-field-binding.ts
```diff
@@ -0,0 +1,94 @@
...
+  } else {
+    renderer.removeAttribute(element, 'title');
+    if (options.originalTitle) {
+      renderer.setAttribute(element, 'title', options.originalTitle);
+    }
+  }
+  if (link.target) {
+    renderer.setAttribute(element, 'target', link.target);
+    if (link.target === '_blank' && !element.getAttribute('rel')) {
+      renderer.setAttribute(element, 'rel', 'noopener noreferrer');
```
illiakovalenko: I think `rel="noopener noreferrer"` can become stale after target changes away from `_blank`.
There is no symmetric removal when target is later cleared or changed.

packages/angular/src/field-directives/link-field-binding.ts
```diff
@@ -0,0 +1,94 @@
...
+    if (options.originalTarget) {
+      renderer.setAttribute(element, 'target', options.originalTarget);
+    }
+  }
+
+  const hasChildren = element.childNodes.length > 0 && element.textContent?.trim();
+  if (!hasChildren) {
+    const text = link.text || link.href || '';
+    renderer.setProperty(element, 'textContent', text);
+  } else if (options.preferTextFromField && link.text) {
```
illiakovalenko: `preferTextFromField` is not covered by unit tests

packages/angular/src/field-directives/link-field-binding.ts
illiakovalenko, line 1: Some functions are not covered by unit tests:
* applyLinkFieldToAnchor
* clearLinkHrefOnAnchor

packages/angular/src/field-directives/link-field-binding.ts
```diff
@@ -0,0 +1,94 @@
...
+    renderer.setProperty(element, 'textContent', link.text || '');
+  }
+}
+
+/**
+ * Clears link-driven attributes when the field is empty (matches ScLink behavior: drop `href` only).
+ */
+export function clearLinkHrefOnAnchor(renderer: Renderer2, element: HTMLAnchorElement): void {
+  renderer.removeAttribute(element, 'href');
```
illiakovalenko: Do we need to cleanup field-empty transitions of other link-driven attrs (target, title, classes, maybe rel)?

packages/angular/src/field-directives/sc-router-link.directive.ts
```diff
@@ -0,0 +1,87 @@
...
+import { LinkFieldValue, LinkField } from '@sitecore-content-sdk/content/layout';
+import {
+  applyLinkFieldToAnchor,
+  clearLinkHrefOnAnchor,
+  resolveLinkFromField,
+} from './link-field-binding';
+
+/**
+ * Renders a Sitecore link field onto a host `<a>` and calls `Router.navigateByUrl` on click,
+ * following the Sitecore JSS Angular `RouterLinkDirective` pattern (sitecore-jss-angular).
```
illiakovalenko: Let’s avoid mentioning JSS, as it may not provide meaningful context for new consumers. I think the name of this component is self explaining

packages/angular/src/field-directives/sc-router-link.directive.ts
```diff
@@ -0,0 +1,87 @@
...
+    if (hrefAttr == null || hrefAttr === '') {
+      return;
+    }
+
+    // Early return in editing mode
+    // if (this.sitecoreContext.isEditing()) {
+    //   return;
+    // }
+
+    void this.router.navigateByUrl(hrefAttr);
```
illiakovalenko, line 43: Should we think about guarding router link directive in case content authors pass external url? Most likely 'onClick' behavior will not be functional

packages/angular/src/field-directives/sc-router-link.directive.spec.ts
```diff
@@ -0,0 +1,80 @@
...
+    a.dispatchEvent(ev);
+
+    expect(spy).toHaveBeenCalledTimes(1);
+    expect(spy).toHaveBeenCalledWith('/about');
+    expect(preventSpy).toHaveBeenCalled();
+
+    spy.mockRestore();
+  });
+
+  it('calls navigateByUrl for href with hash and does not preventDefault (JSS parity)', async () => {
```
illiakovalenko: In general all the functionality should fully match JSS functional requirements, so no need to mention it
```suggestion
  it('calls navigateByUrl for href with hash and does not preventDefault', async () => {
```

packages/angular/src/field-directives/sc-router-link.directive.spec.ts
illiakovalenko, line 1: We are missing some 'navigation' scenarios:
* target="_blank" behavior
* missing/empty href path (no navigation)
* external/protocol URLs (should not router-navigate, if you adopt guardrails)

illiakovalenko (changes requested):
Looks good overall  thanks for the work here 👍 
I’ve left a few suggestions below, and also have a couple of follow-ups:
* Could you enable a code coverage report for the Angular SDK?
* Once coverage is enabled, please review the results and add unit tests where needed. I’ve already highlighted some missing scenarios in the comments.
* Could you also address the lint errors and warnings? There are quite a few at the Angular SDK level.
* Regarding the components implemented at the template level: should we treat this as a draft for the SCB team? In particular, there are some utilities included that likely shouldn’t remain there long term.

Let me know your thoughts.

illiakovalenko (approved):
👍

PR #468, opened 2026-05-06
[nextjs][nextjs-app-router-osr] OSR App Router Template with Tag Revalidation Support

MenKNas:
- [x] Changelog updated

### Summary

This PR introduces a new Next.js App Router template with OSR and tag revalidation out of the box.
It adds deterministic tag-based revalidation so published Sitecore changes refresh cached content reliably.
It exposes a single revalidation endpoint at `POST /api/revalidate` that accepts both explicit tags (for manual / operational invalidation and custom workflows) and Sitecore publish-event payloads (which the app translates into cache tags before invalidation).
The endpoint is protected by an optional shared secret (`SITECORE_REVALIDATE_SECRET`) so authorized callers can trigger invalidations.
It also adds dictionary fallback invalidation to reduce stale dictionary-content cases.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other 

Verified end-to-end with local app + ngrok webhook URL: published Sitecore content, confirmed webhook calls hit `/api/revalidate`, and confirmed updated content appears after refresh for both route and dictionary changes.

## Types of changes


packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/api/revalidate/route.ts
illiakovalenko, line 1: I think the manual revalidation route handler is a good addition, but in our case it’s purely optional. Most customers will likely rely on webhooks, or implement their own lightweight manual triggering mechanism tailored to their needs

scripts/samples.json
```diff
@@ -5,5 +5,12 @@
       "prerender": "SSG",
       "silent": false
     }
+  },
+  {
+    "template": "nextjs-app-router-osr",
```
illiakovalenko: The current template name might not be very clear for customers, since the term "osr" is used purely internally by us. What do you think about making it more explicit that the template uses the Cache Components approach? Something like `nextjs-app-router-cache-components`. It makes sense to ask Product about this as well

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/docs/tag-based-revalidation.md
illiakovalenko, line 1: I'm not sure we should start adding documentation directly into our starters, since it’s easy for it to become outdated and it introduces additional maintenance overhead. However, we definitely should document this properly on the Sitecore documentation side

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -0,0 +1,47 @@
...
+      <p>This page does not exist.</p>
+      <Link href="/">Go to the Home page</Link>
+    </div>
+  );
+}
+
+async function SitecoreLocaleNotFound() {
+  const { site, locale } = getCachedPageParams();
+
+  const page = await client.getErrorPage(ErrorPage.NotFound, {
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: What about error pages? In our Cache Components guide, we also recommend caching error pages requests, but I see this part remains unchanged. How do error pages behave in terms of webhooks and revalidation?

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -0,0 +1,47 @@
...
+  return (
+    <div style={{ padding: 10 }}>
+      <h1>Page not found</h1>
+      <p>This page does not exist.</p>
+      <Link href="/">Go to the Home page</Link>
+    </div>
+  );
+}
+
+async function SitecoreLocaleNotFound() {
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: Could you please explain why it’s called `SitecoreLocaleNotFound`?

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/layout.tsx
```diff
@@ -0,0 +1,9 @@
+import './globals.css';
+
+export default function RootLayout({ children }: { children: React.ReactNode }) {
+  return (
+    <html lang="en">
+      <body>{children}</body>
+    </html>
+  );
+}
```
illiakovalenko:
```suggestion
}

```

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/components/content-sdk/SitecoreStyles.tsx
```diff
@@ -0,0 +1,32 @@
...
+  return (
+    <>
+      {headLinks.map(({ rel, href }: HTMLLink) => (
+        <link rel={rel} key={href} href={href} precedence="high" />
+      ))}
+    </>
+  );
+};
+
+export default SitecoreStyles;
```
illiakovalenko:
```suggestion
export default SitecoreStyles;

```

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/lib/image-remote-patterns.ts
illiakovalenko, line 1: This looks a bit different from our other templates. Usually, we define the patterns in next.config. Could we follow the same approach as in the other templates, or was this done intentionally? Components typically don’t import remote patterns directly.

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/.env.remote.example
```diff
@@ -0,0 +1,55 @@
+# This file should be copied to .env.local and modified with your environment variables to connect to your remote Sitecore instance.
+
+# To secure the Sitecore editor endpoint exposed by your Next.js app
+# (`/api/editing/render` by default), a secret token is used. This (client-side)
+SITECORE_EDITING_SECRET=
+
+# Secret token for protected cache tag revalidation endpoint (`/api/revalidate`).
+# Send this value in the `x-revalidate-secret` header.
```
illiakovalenko: Do customers need to be aware of these details and specific headers? It looks like the webhook always uses the same format and passes the same header, customers will not do that manually

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/package.json
```diff
@@ -0,0 +1,56 @@
+{
+  "name": "content-sdk-nextjs-app-router-osr",
+  "description": "Application utilizing Content SDK and Next.js",
```
illiakovalenko: Suggestion to make the description more specific:
```suggestion
  "description": "Content SDK Next.js App Router sample with Cache Components and On-demand revalidation",
```

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/package.json
```diff
@@ -0,0 +1,56 @@
...
+  "version": "0.1.0",
+  "private": true,
+  "author": {
+    "name": "Sitecore Corporation",
+    "url": "https://doc.sitecore.com/xmc/en/developers/content-sdk/index.html"
```
illiakovalenko:
```suggestion
    "url": "https://doc.sitecore.com/sai/en/developers/content-sdk/sitecore-content-sdk-for-sitecoreai.html"
```

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "type": "git",
+    "url": "git+https://github.com/sitecore/content-sdk.git"
+  },
+  "bugs": {
+    "url": "https://github.com/sitecore/content-sdk/issues"
+  },
+  "license": "Apache-2.0",
+  "scripts": {
+    "build": "cross-env NODE_ENV=production run-s sitecore-tools:generate-map sitecore-tools:build next:build",
+    "build:corp-tls": "cross-env NODE_TLS_REJECT_UNAUTHORIZED=0 NODE_ENV=production run-s sitecore-tools:generate-map sitecore-tools:build next:build",
```
illiakovalenko: I’m not sure this is something we want to include by default

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "next:dev": "cross-env NODE_OPTIONS='--inspect' next dev",
+    "next:start": "next start",
+    "sitecore-tools:generate-map": "sitecore-tools project component generate-map",
+    "sitecore-tools:generate-map:watch": "sitecore-tools project component generate-map --watch",
+    "sitecore-tools:build": "sitecore-tools project build",
+    "dev": "cross-env NODE_ENV=development run-s sitecore-tools:generate-map sitecore-tools:build && run-p next:dev sitecore-tools:generate-map:watch",
+    "start": "cross-env NODE_ENV=production run-s build next:start"
+  },
+  "dependencies": {
+    "@sitecore-content-sdk/nextjs": "<%- version %>",
```
illiakovalenko: Could you please check the latest approach for referencing package versions in templates? This one doesn’t work anymore.

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/package.json
```diff
@@ -0,0 +1,56 @@
...
+    "@types/react": "^19.2.7",
+    "@types/react-dom": "^19.2.3",
+    "cross-env": "^10.0.0",
+    "eslint": "^9.33.0",
+    "eslint-config-next": "16.2.2",
+    "npm-run-all2": "^8.0.4",
+    "tailwindcss": "^4",
+    "typescript": "~5.8.3"
+  }
+}
```
illiakovalenko:
```suggestion
}

```

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/sitecore.config.ts
```diff
@@ -0,0 +1,14 @@
...
+/**
+ * @type {import('@sitecore-content-sdk/nextjs/config').SitecoreConfig}
+ * See the documentation for `defineConfig`:
+ * https://doc.sitecore.com/xmc/en/developers/content-sdk/the-sitecore-configuration-file.html
+ */
+export default defineConfig({
+  dictionary: {
+    caching: {
+      enabled: true,
+      timeout: 60,
```
illiakovalenko: I don't think we need to have it enabled for the OSR / Cache Components story to work. I believe we discussed it - 'use cache' + `cacheTag` is already your primary dictionary cache and the layer `revalidateTag` actually clears.
dictionary.caching is a second, separate cache inside DictionaryService (in-process memory). revalidateTag does not clear it.

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/lib/cache/get-sitecore-page.ts
```diff
@@ -0,0 +1,33 @@
...
+/**
+ * Gets page data using Next.js Cache Components and deterministic Sitecore cache tags.
+ */
+export async function getSitecorePage(params: GetSitecorePageParams): Promise<Page | null> {
+  'use cache';
+
+  const { site, locale, path } = params;
+  const page = await client.getPage(path, { site, locale });
+
+  const personalizedPathname = path.length ? `/${path.join('/')}` : '/';
```
illiakovalenko: If this part is needed, can it be handled internally at the `collectSitecorePageCacheTags` function level like in `getPage`?

packages/nextjs/src/index.ts
```diff
@@ -178,3 +178,33 @@ export {
...
+  type CollectSitecorePageCacheTagsParams,
+} from './cache/sitecore-page-cache-tags';
+
+export {
+  extractSitecoreEdgeContentId,
+  collectSitecoreTagsFromEdgeRevalidateRequestBody,
+  type SitecoreEdgeRevalidateUpdate,
+  type SitecoreEdgeRevalidateRequestBody,
+  type CollectSitecoreTagsFromEdgeBodyOptions,
+} from './cache/sitecore-edge-webhook-revalidation';
```
illiakovalenko: I’d like to ask you to revisit which APIs need to be exposed to customers and which are intended to remain internal. At the moment, we expose quite a lot, and I’m not convinced customers need to be aware of these internals or work with them directly.

packages/nextjs/src/route-handler/revalidate-route-handler.ts
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Environment variable name for resolving the shared secret.
+   * Default is SITECORE_REVALIDATE_SECRET.
+   */
+  secretEnvVarName?: string;
+  /**
+   * Request header name used for passing the secret.
+   * Default is x-revalidate-secret.
+   */
+  secretHeaderName?: string;
```
illiakovalenko: I’m not sure we really need to expose this extra flexibility. It could be exposed, but it doesn’t seem necessary since customers typically rely on our environment variables, and we can also rely on those internally.

Could you elaborate on the header name? Why does it need to be customizable - does it relate to the “authorization item” configuration?

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/api/revalidate/webhook/route.ts
```diff
@@ -0,0 +1,31 @@
...
+          ? [
+              buildSitecoreDictionaryCacheTag({
+                site: scConfig.defaultSite,
+                locale: scConfig.defaultLanguage,
+              }),
+            ]
+          : []
+      )
+  )
+);
```
illiakovalenko: Can we potentially pass `sites` to the route handler and handle dictionary cache resolution internally? Is there a real use case for supporting an `additionalTags` property?

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
```diff
@@ -0,0 +1,71 @@
...
+  const personalize = getPersonalizedRewriteData(pathname);
+  const pathSegments = routeSegmentsFromPersonalizedPathname(pathname);
+
+  return dedupeSitecoreCacheTags([
+    buildSitecoreRouteCacheTag({
+      site: params.site,
+      locale: params.locale,
+      pathSegments,
+    }),
+    buildSitecoreDictionaryCacheTag({ site: params.site, locale: params.locale }),
```
illiakovalenko: Since we add dictionary cache tags via "getDictionary" call, it looks like we don't need to set it here, since "getPage" method is not aware of dictionary and dictionary is optional to be used by customers

packages/nextjs/src/route-handler/edge-webhook-revalidate-route-handler.ts
illiakovalenko, line 1: It might make sense to add debug logs, similar to other route handlers, to make debugging easier

packages/nextjs/src/route-handler/revalidate-route-handler.ts
```diff
@@ -0,0 +1,101 @@
...
+  /**
+   * Next.js cache profile used by revalidateTag.
+   * Default is max.
+   */
+  cacheProfile?: 'max';
```
illiakovalenko: Next.js `cacheProfile` also supports object shape and you can provide a different string value, `max` profile is recommended but not required to be used.
May be it could be extended to:
```ts
cacheProfile: string | Parameters<typeof revalidateTag>[1]
```

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
```diff
@@ -0,0 +1,71 @@
...
+    buildSitecorePersonalizedPageVariantCacheTag({
+      variantId: personalize.variantId,
+      componentVariantIds: personalize.componentVariantIds,
+    }),
+    buildSitecoreItemCacheTagFromRouteData(params.route, params.locale) ?? '',
+  ]).filter(Boolean);
```
illiakovalenko, line 115: Based on our discussion, please revisit the cache tags you are building. It looks like for our use case it’s enough to generate a single tag representing the actual route.

The tag could combine the site, locale, and variant ID (using default when the variant is not present), since all of these are part of the actual page item in Sitecore.

illiakovalenko (changes requested):
Looks solid! Could you please take a look at my suggestions below, also please check linting warnings

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/api/revalidate/webhook/route.ts
illiakovalenko, line 1: I'm thinking if we can just move the route path under /api/revalidate.

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/api/revalidate/webhook/route.ts
```diff
@@ -0,0 +1,31 @@
...
+
+export const { POST } = createEdgeWebhookRevalidateRouteHandler({
```
illiakovalenko: One suggestion: maybe we could rename the function to `createWebhookRevalidateRouteHandler`. Ideally, we wouldn’t mention Edge directly, since it can be somewhat ambiguous. We also probably don’t need to include Sitecore, as it’s omitted in the other route handlers as well. What do you think?

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/package.json
MenKNas: We don't, I added this as a convenience because I have had issues with Node TLS but forgot to remove it.

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/sitecore.config.ts
MenKNas: Agreed, we can drop explicit dictionary memory caching in the OSR template so dictionary invalidation is driven by `use cache` + `revalidateTag` only. Probably was missed on my end.

packages/nextjs/src/index.ts
MenKNas: @illiakovalenko I am not entirely sure about which ones we should offer and which ones we should treat as internal, any suggestions?

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
MenKNas: @illiakovalenko Right now `collectSitecorePageCacheTags` attaches several tags per page (route, personalization variant, route/item id) so cache entries can line up with different revalidation signals. It means that we can cache/revalidate specific items instead of whole route. I am not entirely sure what the better approach is. This was also based on Christian's POC where he does the same per item and not just route. 

I guess we need to answer why is the single route approach better for our OSR flow than the current multi-tag setup? From my understanding webhooks and publishing pipelines return a GUID so it makes more sense to cache/revalidate per item. But if we think handling the route caching is enough I have no objection to adjust this. 
What do you think?

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/docs/tag-based-revalidation.md
MenKNas: I will keep this as seperate notes to provide for doc input and remove from the repo at this point.

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/api/revalidate/route.ts
MenKNas: Everything will be consolidated in a single route `/api/revalidate` since indeed customers can have their own custom solution.

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
illiakovalenko: @MenKNas I think this makes sense to me. One thing I’d like to ask: could you please prepare a mapping between webhook-triggered updates and the corresponding cache tags?
It would be helpful to understand in which scenarios each tag is actually beneficial to use, so we avoid introducing extra tags that are never referenced. This mapping will also be useful for testing purposes.

packages/nextjs/src/index.ts
illiakovalenko: @MenKNas We can check it together once you finalize a first round of changes

packages/nextjs/src/route-handler/revalidate-route-handler.ts
MenKNas: I initially added this for more flexibility but it doesn't make sense the way we wire webhooks anyway so I dropped `secretEnvVarName` and `secretHeaderName`. The idea is we always use `SITECORE_REVALIDATE_SECRET` + the `x-revalidate-secret` header now.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/proxy.ts
```diff
@@ -0,0 +1,110 @@
...
+  RedirectsProxy,
+  LocaleProxy,
+  BotTrackingProxy,
+} from '@sitecore-content-sdk/nextjs/proxy';
+import sites from '.sitecore/sites.json';
+import scConfig from 'sitecore.config';
+import { routing } from './i18n/routing';
+
+export default function proxy(req: NextRequest, event: NextFetchEvent) {
+  // BotTrackingProxy will detect and track bots before any other proxies run
```
illiakovalenko, line 23: Let's add a PreviewProxy to start supporting preview protection as well

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/next.config.ts
```diff
@@ -0,0 +1,56 @@
+﻿import type { NextConfig } from 'next';
+import createNextIntlPlugin from 'next-intl/plugin';
+
+const allowedDevOrigins = (process.env.NEXT_ALLOWED_DEV_ORIGINS ?? '')
```
illiakovalenko: Do we need ot keep it? I believe it's not documented

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/postcss.config.mjs
illiakovalenko, line 1: Do we need to have this file? It's not present in basic nextjs template

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/README.md
```diff
@@ -0,0 +1,7 @@
+# Sitecore Content SDK — Next.js App Router + Cache Components
+
+This starter matches the default **nextjs-app-router** template, with **Next.js Cache Components** (`cacheComponents`), **`getSitecorePage`**, **`getSitecoreDictionary`**, and **`getSitecoreErrorPage`** (Sitecore 404) helpers that apply Sitecore cache tags for tag-based on-demand revalidation, plus a single **`POST /api/revalidate`** route that accepts explicit cache tags or Sitecore webhook-style payloads. From the app root you can call that URL with standard HTTP tooling (see `docs/tag-based-revalidation.md` and the monorepo **`docs/tag-based-revalidation.md`** for the full walkthrough).
```
illiakovalenko:
```suggestion
This starter matches the default **nextjs-app-router** template, with **Next.js Cache Components** (`cacheComponents`), **`getSitecorePage`**, **`getSitecoreDictionary`**, and **`getSitecoreErrorPage`** (Not found / Server Error) helpers that apply Sitecore cache tags for tag-based on-demand revalidation, plus a single **`POST /api/revalidate`** route that accepts explicit cache tags or Sitecore webhook-style payloads. From the app root you can call that URL with standard HTTP tooling (see `docs/tag-based-revalidation.md` and the monorepo **`docs/tag-based-revalidation.md`** for the full walkthrough).
```

illiakovalenko: > From the app root you can call that URL with standard HTTP tooling (see `docs/tag-based-revalidation.md` and the monorepo **`docs/tag-based-revalidation.md`** for the full walkthrough)
These docs will not be present and already removed

packages/nextjs/src/route-handler/revalidate-route-handler.ts
illiakovalenko, line 1: Currently we have:
* revalidate-route-handler
* sitecore-revalidate-route-handler
* webhook-revalidate-route-handler

Could you, please, explain the purpose of each of them? It looks like there are duplicates

scripts/samples.json
```diff
@@ -5,5 +5,12 @@
       "prerender": "SSG",
       "silent": false
     }
+  },
+  {
+    "template": "nextjs-app-router-cache-components",
+    "args": {
+      "prerender": "SSG",
+      "revalidate": true
```
illiakovalenko: Do we support such parameter? I wasn't able to find it

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/.env.remote.example
illiakovalenko: Do we still need to mention `x-revalidate-secret` here? It looks more like an internal implementation detail.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/gitignore
```diff
@@ -0,0 +1,31 @@
...
+.vercel
+
+# sitecore temp files
+.sitecore/*
+# except for component-map
+!.sitecore/component-map.ts
+!.sitecore/component-map.client.ts
+!.sitecore/import-map.ts
+!.sitecore/import-map.server.ts
+!.sitecore/import-map.client.ts
```
illiakovalenko: Let's exclude temp files (Yes, I know it exists in other templates, but we plan to exclude them)

packages/create-content-sdk-app/src/templates/nextjs-app-router-osr/src/app/api/revalidate/route.ts
illiakovalenko: Looks like this template and file should be removed?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/eslint.config.mjs
illiakovalenko, line 1: We also need to add:
* AI related files: Claude.md, Skills, etc.
* .vscode related files
* prettier configuration
* Licence

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/next.config.ts
MenKNas: We can remove this, I added it for testing purposes related to the publishing webhook.

scripts/samples.json
MenKNas: Was initially added in case we wanted to revalidate on demand from CLI but this approach was abandoned. We can remove this it now.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/eslint.config.mjs
MenKNas: @illiakovalenko  My thoughts were to add these on a follow up PR to make it easier to read/review and not miss something out. Are we good with that?

illiakovalenko: @MenKNas My personal opinion is that we should include it here. Since we are introducing a template, it should provide a complete solution. Otherwise, splitting it out makes the template feel incomplete.
Especially because things like Prettier config or .vscode files can usually be easily copied from other templates anyway. AI-related files can also be adjusted with AI assistance if needed. It should take just a bit of time
I'm not strongly opposed here though - you can also check with Alex for his opinion.

MenKNas: @illiakovalenko No worries I can add them all as part of this PR

packages/nextjs/src/route-handler/revalidate-route-handler.ts
illiakovalenko: @MenKNas I'm not sure this one has been addressed yet. Could you please revisit it and take another look?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/postcss.config.mjs
illiakovalenko: @MenKNas I'm not sure this one has been addressed yet. Could you please revisit it and take another look?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/page.tsx
```diff
@@ -0,0 +1,87 @@
...
+    },
+  ];
+};
+<% } -%>
+// Metadata fields for the page.
+export const generateMetadata = async ({ params }: PageProps) => {
+  const { path, site, locale } = await params;
+
+  // The same call as for rendering the page. Should be cached by default react behavior
+  const page = await getSitecorePage({ site, locale, path: path ?? [] });
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/page.tsx]: When `draftMode().isEnabled`, the Page component renders via `client.getPreview(...)` (live preview data), but `generateMetadata` always uses getSitecorePage (cached data). In editing mode the page title could show stale metadata, can you check it, please?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/next.config.ts
```diff
@@ -0,0 +1,50 @@
...
+
+const nextConfig: NextConfig = {
+  // Enable Cache Components (`use cache`, `cacheTag`) in Next.js App Router.
+  cacheComponents: true,
+
+  // Enable Turbopack file system caching for faster dev startup (beta)
+  // See: https://nextjs.org/docs/app/api-reference/config/next-config-js/turbopack
+  experimental: {
+    turbopackFileSystemCacheForDev: true,
+  },
```
illiakovalenko: Starting from next.js v16.1.0 it's enabled by default in development, so we can remove this setting

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/postcss.config.mjs
MenKNas: @illiakovalenko We need this file for the tailwind config, it's the same as App Router

packages/nextjs/src/route-handler/revalidate-route-handler.ts
MenKNas: > @MenKNas I'm not sure this one has been addressed yet. Could you please revisit it and take another look?

This can be consolidated to a single handler: `createSitecoreRevalidateRouteHandler`. It accepts both manual `sc:-`prefixed tags and Sitecore webhook payloads on the same `POST /api/revalidate` route, dispatching by body shape. The other handlers can be dropped as their behavior is fully reachable through this one.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/next.config.ts
MenKNas: Correct, will remove.

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
```diff
@@ -0,0 +1,267 @@
...
+      });
+
+      for (const tag of tags) {
+        revalidateTag(tag, cacheProfile);
+      }
+
+      debug.revalidate('sitecore revalidate manual path end in %dms: %o', Date.now() - startTimestamp, {
+        tagsCount: tags.length,
+      });
+
```
illiakovalenko: Let's revisit, most likely we don't need to support the 'manual' case

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
```diff
@@ -0,0 +1,112 @@
...
+    buildSitecoreRouteCacheTag({
+      site: params.site,
+      locale: params.locale,
+      pathSegments,
+    }),
+    buildSitecorePersonalizedPageVariantCacheTag({
+      variantId: personalize.variantId,
+      componentVariantIds: personalize.componentVariantIds,
+    }),
+    buildSitecoreItemCacheTagFromRouteData(params.route, params.locale) ?? '',
```
illiakovalenko: Looks like variantId's are not supported by webhook
If it's not supported by the webhook -> then let's drop it

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
```diff
@@ -0,0 +1,267 @@
...
+    secret,
+    cacheProfile = 'max',
+  } = options;
+
+  const dictionaryTags =
+    sites !== undefined || defaultSite?.trim()
+      ? buildSitecoreDictionaryCacheTagsFromSites({
+          sites: sites ?? [],
+          baseLocale: defaultLocale,
+          extraDictionarySite: defaultSite,
```
illiakovalenko: Is it a "defaultSite"? Could you, please, double check? May be it could be renamed?

packages/nextjs/src/index.ts
illiakovalenko: As we discussed, please, revisit the API and evalute each of the statement

packages/nextjs/src/cache/sitecore-cache-tags.ts
```diff
@@ -0,0 +1,225 @@
...
+    .sort();
+  const suffix = extras.length > 0 ? `:${extras.join('+')}` : '';
+  return `${SITECORE_CONTENT_CACHE_TAG_PREFIX}:pvv:${variant}${suffix}`;
+}
+
+/**
+ * Minimal route data shape for building item cache tags from layout responses.
+ * @public
+ */
+export type SitecoreRouteDataLike = {
```
illiakovalenko: Let's remove this type and instead reuse the existing "RouteData" interface from "content/layout" package

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
MenKNas: This is actually creating confusion so I renamed to `extraDictionarySite` to match the underlying helper and reflect what it actually does (produces one extra `sc:dict:<site>:<defaultLocale>` tag for the canonical site).

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
MenKNas: We can drop this. After double checking, the webhook doesn't carry variant info and there's no invalidation path for `sc:pvv` .

packages/nextjs/src/cache/sitecore-cache-tags.ts
MenKNas: Valid point, removed `SitecoreRouteDataLike` and switched to `RouteData` from content/layout directly

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
MenKNas: Done, dropped the manual branch entirely.

packages/nextjs/src/cache/sitecore-cache-tags.ts
```diff
@@ -0,0 +1,219 @@
...
+   */
+  componentVariantIds?: string[];
+};
+
+/**
+ * Tag for a personalized page variant so caches do not bleed across variants.
+ * @param {BuildSitecorePersonalizedPageVariantCacheTagParams} params - Variant id and optional component variant ids.
+ * @internal
+ */
+export function buildSitecorePersonalizedPageVariantCacheTag(
```
illiakovalenko: We agreed to drop `sc:pvv` tags because the webhook does not carry variant info, let's remove it

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
```diff
@@ -0,0 +1,182 @@
...
+      : [];
+
+  const POST = async (req: NextRequest) => {
+    const startTimestamp = Date.now();
+    try {
+      const configuredSecret = secret ?? process.env[DEFAULT_SECRET_ENV_VAR];
+      if (!configuredSecret) {
+        debug.revalidate('sitecore revalidate: %s is not configured', DEFAULT_SECRET_ENV_VAR);
+        return NextResponse.json(
+          { error: `${DEFAULT_SECRET_ENV_VAR} is not configured.` },
```
illiakovalenko: Let's don't expose the internals - environment variable name. Instead we could return the generic message and the environment variable still name can be logged (line 97)
```ts
return NextResponse.json({ error: 'Server configuration error.' }, { status: 500 });
```

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
```diff
@@ -0,0 +1,182 @@
...
+
+      debug.revalidate('sitecore revalidate end in %dms: %o', Date.now() - startTimestamp, {
+        tagsCount: tags.length,
+        invocation_id: webhookBody.invocation_id ?? null,
+        continues: webhookBody.continues ?? false,
+      });
+
+      return NextResponse.json({
+        revalidated: true,
+        tags,
```
illiakovalenko: The full tag list exposes internal cache key structure (site names, locales, item GUIDs, path segments) to every (non)authenticated caller. Especially since now token becomes optional I think we need to expose only `tagsCount`

```ts
return NextResponse.json({
  revalidated: true,
  tagsCount: tags.length,  // safer default
  invocation_id: webhookBody.invocation_id ?? null,
  continues: webhookBody.continues ?? false,
});
```

packages/nextjs/src/cache/sitecore-page-cache-tags.ts
```diff
@@ -0,0 +1,113 @@
...
+} from './sitecore-cache-tags';
+
+/** @param {string} pathname - Raw pathname (may omit leading slash). */
+function normalizePathname(pathname: string): string {
+  const trimmed = pathname.trim() || '/';
+  return trimmed.startsWith('/') ? trimmed : `/${trimmed}`;
+}
+
+/** Trim leading and trailing `/` without regex (linear time; avoids ReDoS flags on path-derived input). */
+function trimSlashes(part: string): string {
```
illiakovalenko, line 19: Here we have a linting error, `trimSlashes` is missing tsdoc `@param "part"`

packages/nextjs/src/cache/sitecore-cache-tags.ts
```diff
@@ -0,0 +1,219 @@
...
+
+/**
+ * Parameters for {@link buildSitecoreItemCacheTag}.
+ * @internal
+ */
+export type BuildSitecoreItemCacheTagParams = {
+  itemId: string;
+  locale: string;
+  /**
+   * Published version number, or omit / `undefined` for "latest" (no version in the key).
```
illiakovalenko: I suppose "latest" is still present, so `version` is required to be present in the key
```suggestion
   * Published version number, or omit / `undefined` for "latest".
```

packages/nextjs/src/cache/sitecore-edge-webhook-revalidation.ts
```diff
@@ -0,0 +1,114 @@
+import {
+  SITECORE_CONTENT_CACHE_TAG_PREFIX,
+  buildSitecoreItemCacheTag,
+  dedupeSitecoreCacheTags,
+} from './sitecore-cache-tags';
+
+/**
+ * One content change entry as commonly seen in Experience Edge / Content Operations style payloads (POC shape).
```
illiakovalenko: Should we remove it? Does it stand for a different definition? Please, also see two other entries in the same file

```suggestion
 * One content change entry as commonly seen in Experience Edge / Content Operations style payloads.
```

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
```diff
@@ -0,0 +1,193 @@
...
+   * **`x-revalidate-secret`** header. When empty/omitted, revalidation proceeds without that header.
+   */
+  secret?: string;
+  /**
+   * Next.js `revalidateTag` cache profile (second argument). Default is `"max"` (recommended).
+   * Other string values may match profiles from `cacheLife` in `next.config`; objects may use `{ expire }` per Next.js docs.
+   */
+  cacheProfile?: RevalidateTagCacheProfile;
+  /** Locale for item tags when culture is missing, and for dictionary tags when a site has no language. */
+  defaultLocale?: string;
```
illiakovalenko, line 55: Let's mention our fallback value - 'en'

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
```diff
@@ -0,0 +1,193 @@
...
+  /** Sites list; merges one `sc:dict:<site>:<locale>` cache tag per site on every revalidation call. */
+  sites?: SiteInfo[];
+  /**
+   * Optional **dictionary-only** site name. When set, the handler appends one extra
+   * `sc:dict:<extraDictionarySite>:<defaultLocale>` tag on top of the per-site tags above.
+   * Use this as a defensive guarantee that the canonical site's dictionary tag is always
+   * invalidated even if the `sites` list is empty or carries a different language string.
+   * Note: this option does **not** influence routing or any other tag family.
+   */
+  extraDictionarySite?: string;
```
illiakovalenko: This one looks a bit like overkill to me. Let’s clarify it first and see whether we can simply remove this property instead.
We already provide a `sites` property list, and it should include the default site as well. Wouldn’t that be enough to handle our case?
The "sites" list is generated during the build step and it always contains at least one site (for example the one defined via env variables)

illiakovalenko (changes requested):
It looks good now! 👍
I found a some things still need to be addressed, but overall it looks solid

packages/nextjs/src/route-handler/sitecore-revalidate-route-handler.ts
MenKNas: We can actually remove `extraDictionarySite` since sites from `.sitecore/sites.json` already includes the default site from `generateSites`, so it was redundant.

illiakovalenko (approved):
Looks great! 👍 Thank you for all the effort!

PR #487, opened 2026-05-21
[Angular CSDK] Internalization and locale routing

art-alexeyenko:
List of changes:
- `ngx-translate` integration
  - uses `dictionary` from loader
  - implements TranslateLoader
- Adds locale-aware routing
 - locale-dependent error page
 - locale-dependent catch-all routing for content pages
 - locale-aware router links. /en will be added to links when current page has /en prefix. No prefix to links when current page does not have language prefix (defaultLanguage from config used as fallback)
 - Locales configured via sitecore.config
   - new section: `angular`
   - new property `angular.locales`
   - `redirect.locales` removed for Angular - its value will be the same as `angular.locales`
 - Reworked SitecoreContextService
   - exposes `effectiveLocale` and `urlLocale`
   - `urlLocale` helps ensure links are rendered with correct locale prefix
   - moved from explicit react-like context, to implicit reactive Angular-like context - reads page, dictionary from root

Some notes:
- scLink is made locale aware via context service
- scRouterLink and Angular's routerLink are made locale aware through custom `UrlSerializer`, injected into Angular


packages/angular/src/lib/providers.ts
```diff
@@ -43,24 +50,33 @@ export interface SitecoreAngularConfig {
...
 
-  return makeEnvironmentProviders(providers);
+  providers.push(SitecoreContextService);
+
+  const locales = init.sitecoreConfig?.angular?.locales ?? [
+    init.sitecoreConfig?.defaultLanguage ?? 'en',
+  ];
+  if (locales.length > 0) {
+    providers.push({ provide: UrlSerializer, useClass: LocaleUrlSerializer });
+  }
```
MenKNas: If  I read this correctly the ?? fallback above makes locales.length > 0 always true, so `LocaleUrlSerializer` replaces `DefaultUrlSerializer` for every consumer. Is this something we want? Doesn't it mean that all `/about` routes will be treated as `/en/about` even if someone hasn't opted in for i18n?

packages/angular/src/i18n/locale-utils.ts
```diff
@@ -0,0 +1,80 @@
...
+ * @param {string[]} locales - Configured locales.
+ * @returns {LocaleExtractionResult} Detected locale and the rest of the path.
+ * @public
+ */
+export function splitLocaleFromPath(pathname: string, locales: string[]): LocaleExtractionResult {
+  if (!pathname) {
+    return { locale: null, nonLocalePath: '/' };
+  }
+  const normalized = pathname.startsWith('/') ? pathname : `/${pathname}`;
+  const [, firstSegment = '', restPath = '', _queryFragmentSuffix = ''] =
```
MenKNas: By capturing `_queryFragmentSuffix` and never appending it back, does it drop the query/fragment from `nonLocalePath` ?

Does this happen? ->  `/en/about?queryParam=test` returns  simply `/about` ? 

If this is the case, I guess this is unwanted behaviour right?

packages/angular/src/lib/providers.ts
```diff
@@ -43,24 +50,33 @@ export interface SitecoreAngularConfig {
...
+    providers.push({ provide: NOT_FOUND_ROUTE_TOKEN, useValue: init.notFoundRoute });
   }
-  if (config.errorRoute) {
-    providers.push({ provide: ERROR_ROUTE_TOKEN, useValue: config.errorRoute });
+  if (init.errorRoute) {
+    providers.push({ provide: ERROR_ROUTE_TOKEN, useValue: init.errorRoute });
   }
 
-  return makeEnvironmentProviders(providers);
+  providers.push(SitecoreContextService);
```
MenKNas: `SitecoreContextService` is already `@Injectable({ providedIn: 'root' })`, so this `providers.push(SitecoreContextService)`  is probably redundant? Can you check please?

packages/create-content-sdk-app/src/templates/angular/src/app/app.routes.ts
```diff
@@ -1,23 +1,29 @@
...
+import scConfig from '../../sitecore.config';
 import { PageComponent } from './pages/page.component';
 import { NotFoundComponent } from './pages/not-found.component';
 import { ErrorComponent } from './pages/error.component';
 
+/**
+ * Error routes (`404` / `500`) live at the top level — both unprefixed and with a
+ * `:locale` parameter segment — so that `@angular/ssr` can match its `ServerRoute`
+ * entries by `path` and emit the correct HTTP status. The catchall page sits behind
+ * `createLocaleMatcher`, which still consumes a configured locale (if present) and
```
MenKNas: The TSDoc references `createLocaleMatcher`, but the exported helper is `scLocaleMatcher` . Can we rename in the comment to have everything aligned?

packages/angular/src/i18n/sitecore-translate-loader.ts
```diff
@@ -0,0 +1,15 @@
...
+@Injectable()
+export class SitecoreTranslateLoader implements TranslateLoader {
+  private readonly context = inject(SitecoreContextService);
+
+  getTranslation(): Observable<Record<string, string>> {
+    // dictionary in context will always be current locale dictionary
+    const dictionary = this.context.dictionary();
+    return of(dictionary ?? {});
+  }
+}
```
MenKNas, line 24: Just a suggestion, maybe we could remove comment in L:11 and add this at the TSDoc level for better DX? 

Something like:

```
/**
 * `ngx-translate` loader backed by {@link SitecoreContextService}.
 * Requires a `dictionaryLoader` resolver on the active route — without it, `dictionary()`
 * is `null` and translations resolve to `{}`.
 * @public
 */
```

packages/angular/.claude/settings.local.json
```diff
@@ -0,0 +1,9 @@
+{
+  "permissions": {
+    "allow": [
+      "Bash(yarn build *)",
+      "Bash(yarn lint *)",
+      "Bash(yarn workspace *)"
+    ]
+  }
+}
```
MenKNas: Is this something we need for Angular at this point?

MenKNas (changes requested):

MenKNas:
@art-alexeyenko Great work on expanding the Angular work for CSDK! Please see some suggestions for improvements.

packages/angular/src/i18n/locale-utils.ts
art-alexeyenko: Right now we mainly use this function to get locale from incoming request - we don't use query string anywhere (partially following the Angular router logic that doesn't consider it either).
But you're right it's helpful to have query available alongside other parts. I'm adding this as a part of return value

packages/angular/.claude/settings.local.json
art-alexeyenko: nope, this is there by mistake

packages/angular/src/lib/providers.ts
art-alexeyenko: good point.
We'll always initialize locales with at least the `defaultLanguage` present, and we'll want the serializer to be activated in most cases - but we also need to give users the option to disable or customize it, so I moved the provider into the angular app.

packages/create-content-sdk-app/src/templates/angular/src/app/app.routes.ts
```diff
@@ -9,7 +9,7 @@ import { ErrorComponent } from './pages/error.component';
  * Error routes (`404` / `500`) live at the top level — both unprefixed and with a
  * `:locale` parameter segment — so that `@angular/ssr` can match its `ServerRoute`
  * entries by `path` and emit the correct HTTP status. The catchall page sits behind
- * `createLocaleMatcher`, which still consumes a configured locale (if present) and
+ * `scLocaleMatcher `, which still consumes a configured locale (if present) and
```
MenKNas: Remove trailing space for `scLocaleMatcher`  (sorry, for some reason GH doesn't let me add this as a suggestion)

packages/angular/src/lib/providers.ts
```diff
@@ -1,23 +1,26 @@
-import type { SitecoreConfig } from '@sitecore-content-sdk/content/config';
 import type { SitecoreClient } from '@sitecore-content-sdk/content/client';
 import { EnvironmentProviders, makeEnvironmentProviders } from '@angular/core';
+import { UrlSerializer } from '@angular/router';
```
MenKNas: This can be removed now right? It should also be removed from the TSDoc

packages/angular/src/lib/providers.ts
```diff
@@ -1,23 +1,26 @@
...
 import {
   SITECORE_CONFIG_TOKEN,
   SITECORE_CLIENT_TOKEN,
   ERROR_ROUTE_TOKEN,
   NOT_FOUND_ROUTE_TOKEN,
 } from './tokens';
+import type { AngularSitecoreConfig } from '../config/define-config';
+import { LocaleUrlSerializer } from '../i18n/locale-url-serializer';
+import { SitecoreContextService } from './sitecore-context.service';
```
MenKNas: Same here, should be removed from both imports and TSDoc mentions

packages/angular/src/lib/providers.ts
```diff
@@ -43,24 +50,33 @@ export interface SitecoreAngularConfig {
  * @returns {EnvironmentProviders} Angular environment providers
  * @public
  */
-export function provideSitecoreAngular(config: SitecoreAngularConfig): EnvironmentProviders {
-  const providers = [];
+export function provideSitecoreAngular(init: AngularCSDKAppInit): EnvironmentProviders {
+  const providers: Array<unknown> = [];
```
MenKNas: Question, could this be typed to:

`const providers: Array<Provider | EnvironmentProviders> = []`;

so we can avoid the use of `unknown` and make it more robustly typed?

MenKNas (approved):

MenKNas:
@art-alexeyenko Thank you for addressing everything, I have approved but please have a look at my final round of comments for some extra improvements/cleanups.

PR #490, opened 2026-05-27
[nextjs] Add context to nextjs proxies that the developers can use to get information on what was executed inside each proxy

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
Based on the initial discussion https://github.com/Sitecore/content-sdk/discussions/420 and  POC https://github.com/Sitecore/content-sdk/pull/421 by @mjarosch, a more generic solution was developed. With the new approach we enable developers to have access to some information about the execution inside the nextjs proxies. This information can be used to perform other actions like side effects, analytics etc.

### Example:
```ts
// ... current proxy file contents

  const proxiesContext: ProxiesContext = new Map();

  const response = await defineProxy(preview, botTracking, multisite, redirects, personalize).exec(
    req,
    undefined,
    proxiesContext
  );

  const redirectsInfo = proxiesContext.get(redirects.name);

  if (isSuccessfulProxyExecution<SuccessfulRedirectsProxyExecution>(redirectsInfo))
    console.log('Redirects Proxy executed a redirect:', redirectsInfo);
  else console.log('Redirects Proxy failed:', redirectsInfo?.error);

  return response;

```

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/nextjs/src/proxy/multisite-proxy.ts
```diff
@@ -125,10 +151,27 @@ export class MultisiteProxy extends ProxyBase {
...
 
+      const successfulExecution: SuccessfulMultisiteProxyExecution = {
+        executedSuccessfully: true,
+        error: null,
+        rewritePath,
+        siteName,
+        isSitecorePreview,
+      };
+
+      context?.set(this._name, successfulExecution);
```
art-alexeyenko: can we add some tests to ensure proxy handlers complete fine when context is undefined?

packages/nextjs/src/proxy/multisite-proxy.ts
```diff
@@ -125,10 +151,27 @@ export class MultisiteProxy extends ProxyBase {
         cookies: response.cookies,
       });
 
+      const successfulExecution: SuccessfulMultisiteProxyExecution = {
```
art-alexeyenko, line 152: since we use objects for proxies, what do you think about defining proxyContextSuccess/proxyContextFail as methods on ProxyBase?

packages/nextjs/src/proxy/utils.ts
```diff
@@ -0,0 +1,17 @@
...
+
+/**
+ * Type guard to check if the proxy execution was successful
+ * @template SuccessfulProxyType - The type of the successful proxy execution information
+ * @template T - The type of the proxy execution information, which can be either successful or failed execution information
+ * @param {T} info - Information about executed proxy to be stored in the context
+ * @returns Type guard to check if the proxy execution was successful
+ * @public
+ */
+export function isSuccessfulProxyExecution<
```
art-alexeyenko, line 11: what do you think about having this as a method in ProxyBase? 
would simplify the check to `proxyObj.isSuccessful()` without the need to memorize and pass down types to the generic descriptor

packages/nextjs/src/proxy/proxy.ts
```diff
@@ -38,17 +39,35 @@ export type ProxyBaseConfig = {
...
+ * @public
+ */
+export type ProxiesContext = Map<string, ProxiesContextMapValue>;
+
 /**
  * Proxy handler class to be extended by all proxy implementations
  * @public
  */
 export abstract class ProxyHandler {
+  // TODO: Add the name property to the base class and make it abstract, so all proxies are forced to have a name. This will be used for better logging and for storing executed proxy information in the context. To avoid breaking change, we will not make it abstract for now.
```
art-alexeyenko: is this TODO still actual?

packages/nextjs/src/proxy/proxy.ts
```diff
@@ -276,9 +295,23 @@ export const defineProxy = (...proxies: ProxyHandler[]) => {
      * Execute all proxies
      * @param {NextRequest} req request
      * @param {NextResponse} [res] response
+     * @param {boolean} [generateContext] whether to generate context for storing additional information during proxy execution, e.g. for sharing between different proxies or for returning information about executed proxies.
      */
-    exec: async (req: NextRequest, res?: NextResponse) => {
+    exec: async <
+      GenerateContext extends boolean,
+      ExecReturnType = GenerateContext extends true
```
art-alexeyenko: this change may break things. Is there an alternative approach we can use?

art-alexeyenko (changes requested):
@sc-nikolaoslazaridis This is good - and I think we can refine it slightly more.
Please see comments - and please also change the proxy.ts file in sample to consume this change.

packages/nextjs/src/proxy/proxy.ts
art-alexeyenko: i.e. it will be easy to pass wrong result type into nextjs proxy logic by mistake

packages/nextjs/src/proxy/multisite-proxy.ts
sc-nikolaoslazaridis: All the current tests we currently have pass that. If no context provided nothing fails (we got full coverage on this new feature) .

packages/nextjs/src/proxy/proxy.ts
sc-nikolaoslazaridis: Yes, the thought process was that if devs already extend from this one they will face errors of not having a name. As mentioned in the call the are a couple of things that can be improved on v3 release.

packages/nextjs/src/proxy/proxy.ts
sc-nikolaoslazaridis: Hmm, during my manual testing i didn't face any issues... do you have an example of how could this break?

packages/nextjs/src/proxy/utils.ts
sc-nikolaoslazaridis: Going with that approach would mean we need to store the context per instance. Doing that will lock the ability to use a singleton (create an instance outside the proxy function in the proxy.ts to be shared across requests). I explored to have it as static method but didn't like it since you either have to have it in the base and again you need to pass generics and the result or if you have it in the child class you have a lot duplication. This utility will only be called by small percentage of the users so i decided to keep it separate and have the classes as clean as possible.

sc-nikolaoslazaridis:
> @sc-nikolaoslazaridis This is good - and I think we can refine it slightly more. Please see comments - and please also change the proxy.ts file in sample to consume this change.

I would not do it since it will be rarely used, unless you insist of adding it ill skip that comment if it is ok.

packages/nextjs/src/proxy/multisite-proxy.ts
sc-nikolaoslazaridis: We will have to provide the context and also need extra methods that will be single lines(will have to do the same thing basically).

packages/nextjs/src/proxy/proxy.ts
art-alexeyenko: I think we're best not to leave any technical debt when possible. In here, `_name` field and getName() getter can be defined on the Proxy base. `_name` can have a default value (overridden in descendants), to make this a non breaking change.

packages/nextjs/src/proxy/proxy.ts
art-alexeyenko: To me returning completely different, unconnected types based on function input seems like an antipattern. It is too easy to add a single bool to `exec()` call and have the whole `proxy()` function return a different result altogether (it's also not very obvious exec will do that), and break the app.

sc-nikolaoslazaridis: Adding parameters by mistake **should** break the app. If you run any kind of tests for your app, you should see that you changed something by mistake and kill the app.

packages/nextjs/src/proxy/proxy.ts
sc-nikolaoslazaridis: Added an abstract getter and a "default" to the ProxyBase class

packages/nextjs/src/proxy/proxy.ts
sc-nikolaoslazaridis: Moving the context creation to the proxy file so the developer should create it, use that instead of boolean param in the exec function

art-alexeyenko (approved):
Looks good, thanks again!

PR #491, opened 2026-05-27
[Angular CSDK] ISR-like loader cache implementation

art-alexeyenko:
Introduces a caching layer for loaders in Angular CSDK. It's designed to provide ISR-like behavior: cache entries for loaders are created and reused until becoming stale. When stale loader cache is accessed, it is revalidated (stale-while-revalidate, similar to next).
This is not a full fledged ISR implementation, however, and it's goal is to minimize Edge calls. Similarly, this does not introduce SSG support (it will be it's own can of worms later, but we have building blocks).

- adds ISR-like cache implementation
- cache contains loaders data (page, dictionary)
- support for OSR tag-based revalidation (following conventions from recent nextjs implementation)
- Temp Cache admin demo page for testing
- revalidation follows stale-while-revalidate principles
  - this includes changes to base loader resolver logic
  - two distinct loader flows: browser, server
  - cache only exists on server (we also ensure loaders themselves are only executed on server)
  - when cache entry ttl is reached or revalidate request received, cache entry is stale instead of removed (performance ++)
- tags support
  - there are two main cache implementations - default in-memory (similar to rjx-angular/isr) and unstorage
  - default one has a map of cache entries (with cache keys) and a collection of tag -> [cache keys] entries. This is how we find cache key by tag.
  - unstorage accepts a driver during implementation (memory, fs, redis etc) and uses it to invoke a storage. unstorage storage stores tags, cache keys and cache values in a similar way.
- cache is invoked in app, `src/server.ts` and cache object pointer is passed down into Angular SSR to ensure consistency.


packages/angular/src/server/server-loader-runner.ts
```diff
@@ -0,0 +1,159 @@
...
+      return { kind: 'error', status: 500, message: `No loader registered for id "${loaderId}"` };
+    }
+
+    const ctx: LoaderContext = { url, params, query, requestContext: angularRequestContext };
+
+    const cacheable = this.cache && (cacheOptions?.enabled ?? this.cache.enabled());
+
+    if (cacheable) {
+      const { key } = buildCacheKey(loaderId, ctx);
+      const read = await this.cache!.get(key);
```
sc-nikolaoslazaridis: Should we have a check for the cache before?

packages/create-content-sdk-app/src/templates/angular/src/server.ts
```diff
@@ -7,20 +7,72 @@ import {
 } from '@angular/ssr/node';
 import express from 'express';
 import { join } from 'node:path';
-import { createLoaderDataServiceMiddleware } from '@sitecore-content-sdk/angular';
+import fsDriver from 'unstorage/drivers/fs';
```
sc-nikolaoslazaridis, line 10: Maybe we can clean this up a bit and use memory cache and if the devs need something different they just switch.

packages/angular/src/server/models.ts
```diff
@@ -68,20 +69,25 @@ export type ExpressMiddleware = (
 ) => void | Promise<void>;
 
 /**
- * Loader registry type - maps loader IDs to loader functions
  * @public
+ * @deprecated Import {@link LoaderRegistry} from `@sitecore-content-sdk/angular` loader registry exports instead.
```
sc-nikolaoslazaridis: Is this something we need ("@deprecated")?

packages/angular/src/server/middleware/sitecore-revalidate-middleware.ts
```diff
@@ -0,0 +1,145 @@
...
+const DEFAULT_SECRET_ENV_VAR = 'SITECORE_REVALIDATE_SECRET';
+const DEFAULT_SECRET_HEADER = 'x-revalidate-secret';
+const DEFAULT_ENDPOINT = '/api/revalidate';
+
+/**
+ * Read a process environment variable
+ * @param {string} name - The name of the environment variable
+ * @returns {string | undefined} The value of the environment variable
+ */
+function readProcessEnv(name: string): string | undefined {
```
sc-nikolaoslazaridis: Can we have this as a utility function for the whole package?

packages/angular/src/server/middleware/sitecore-revalidate-middleware.ts
```diff
@@ -0,0 +1,145 @@
...
+
+    const startTimestamp = Date.now();
+
+    try {
+      const configuredSecret = resolveConfiguredRevalidateSecret(
+        secret,
+        readProcessEnv(DEFAULT_SECRET_ENV_VAR)
+      );
+
+      if (configuredSecret) {
```
sc-nikolaoslazaridis: Do we need to return if the secret is undefined?

packages/angular/src/server/middleware/sitecore-edge-webhook-revalidation.ts
```diff
@@ -0,0 +1,95 @@
...
+  options: CollectSitecoreTagsFromEdgeBodyOptions
+): string[] {
+  const { defaultLocale } = options;
+  const out: string[] = [];
+
+  for (const raw of body?.tags ?? []) {
+    if (typeof raw !== 'string') {
+      continue;
+    }
+    const s = raw.trim();
```
sc-nikolaoslazaridis: Maybe we need to rename this variable to "tag"?

packages/angular/src/server/cache/utils.ts
```diff
@@ -0,0 +1,176 @@
...
+}
+
+/**
+ * Deduplicates strings while preserving first-seen order.
+ * @param {string[]} values - Tag or key candidates.
+ * @returns {string[]} Deduplicated list.
+ * @internal
+ */
+export function dedupeCacheStrings(values: string[]): string[] {
+  const seen = new Set<string>();
```
sc-nikolaoslazaridis: The constructor accepts an array as argument and does the deduplication automatically https://developer.mozilla.org/en-US/docs/Web/JavaScript/Reference/Global_Objects/Set/Set#iterable

packages/angular/src/server/cache/utils.ts
```diff
@@ -0,0 +1,176 @@
...
+
+/**
+ * Strips `driver` from {@link GlobalLoaderCacheConfig} before passing config to backends.
+ * @param {GlobalLoaderCacheConfig} config - Global cache config from {@link createLoaderCache}.
+ * @returns {LoaderCacheConfig} Backend-safe config without the unstorage driver instance.
+ * @internal
+ */
+export function resolveConfig(config: GlobalLoaderCacheConfig): LoaderCacheConfig {
+  const { driver, ...rest } = config;
+  void driver;
```
sc-nikolaoslazaridis: What is this line doing?

packages/angular/src/server/cache/unstorage-loader-cache.ts
```diff
@@ -0,0 +1,196 @@
...
+  }
+
+  /** @inheritdoc */
+  async set(cacheKey: string, value: unknown, ttlSeconds: number, tags: string[]): Promise<void> {
+    const existing = await this.storage.getItem<LoaderCacheEntry>(this.cacheStorageKey(cacheKey));
+    if (existing) {
+      await this.unlinkTags(cacheKey, existing.tags);
+    }
+
+    const expiresAt = ttlSeconds > 0 ? Date.now() + ttlSeconds * 1000 : null;
```
sc-nikolaoslazaridis, line 60: Should we handle negative ttl as "invalidate/delete" operation?

packages/angular/src/server/cache/unstorage-loader-cache.ts
```diff
@@ -0,0 +1,196 @@
...
+        storedAt: entry.storedAt,
+        expiresAt: entry.expiresAt,
+        stale: entry.stale,
+      });
+    }
+    return out;
+  }
+
+  /** @inheritdoc */
+  resolveTtl(): number {
```
sc-nikolaoslazaridis: Can we use normal getters for the cases bellow? Something like:
```
get ttl(){
  return this.config.revalidate;
}
```

packages/angular/src/server/cache/loader-cache.ts
```diff
@@ -0,0 +1,33 @@
...
+ * });
+ * ```
+ * @public
+ */
+export function createLoaderCache(config: GlobalLoaderCacheConfig = {}): LoaderCache {
+  const resolved = resolveConfig(config);
+  if (config.driver) {
+    return new UnstorageLoaderCache(config.driver, resolved);
+  }
+  return new InMemoryLoaderCache(resolved);
```
sc-nikolaoslazaridis: Maybe we could delete this class and use the above one with the in memory driver  provided by us.

packages/angular/src/config/define-config.ts
```diff
@@ -32,6 +32,16 @@ export interface AngularSitecoreConfigInput extends Omit<SitecoreConfigInput, 'r
      * `defaultLanguage` is prepended automatically when absent.
      */
     locales?: string[];
+    /**
+     * Configuration for the ISR-like cache. Both fields default when omitted
+     * (`enabled: true`, `revalidate: 300`).
+     */
+    isrCache?: {
```
sc-nikolaoslazaridis: Lets think a different name, just to avoid any future confusion of the devs using the feature.

packages/angular/src/server/cache/unstorage-loader-cache.ts
art-alexeyenko: I don't think that's necessary. It is understood 0 ttl value means "never expire", and anything less than zero will mean the same.

packages/angular/src/server/cache/utils.ts
art-alexeyenko: true. simplified function

packages/angular/src/server/cache/utils.ts
art-alexeyenko: acting as decoy for lint :D . removed

packages/angular/src/server/models.ts
art-alexeyenko: changed to TODO

packages/angular/src/server/cache/unstorage-loader-cache.ts
sc-nikolaoslazaridis: ok, just got the idea from when you want to delete a cookie you pass a date in the past. Closing this one.

sc-nikolaoslazaridis (approved):
Good job!

PR #492, opened 2026-05-28
[content] Fix build fail error related to non string styles params

MenKNas:
- [ ] Changelog updated

## Description / Motivation

This PR provides a solution fix for a prerender build crash caused by calling .match() on non-string rendering params when `DetailedRenderingParams` returns objects.

This is done by introducing a small `asString` guard in themes traversal so regex logic runs only on real strings, while non-string values are safely ignored and existing fallback behaviour remains intact.

## Testing Details
- [x] Unit Test Added

## Types of changes


illiakovalenko (changes requested):
I think we’re missing a couple of scenarios here, which could introduce bugs. It would be good to test these cases locally as well.

1. There is a scenario where `styles` can be provided as an object, but its value (`Styles.value`) is currently ignored. If this is part of the defined contract, we should validate and properly handle the value; otherwise customers will not be able to apply styles as expected.
Image

2. I assigned a `styles` id to the component in Pages Editor and also added additional styles via the right-hand side panel (e.g. alignment, classname). In this case, we receive a merged JSON including the `styles` id, which is incorrect and should be addressed - likely by a different team - likely it should be done before our release, otherwise customers will report a new item, please create an item / test and track it. Currently we just set a huge string as a class.
Image

MenKNas:
> I think we’re missing a couple of scenarios here, which could introduce bugs. It would be good to test these cases locally as well.
> 
> 1. There is a scenario where `styles` can be provided as an object, but its value (`Styles.value`) is currently ignored. If this is part of the defined contract, we should validate and properly handle the value; otherwise customers will not be able to apply styles as expected.
> 
> Image
> 2. I assigned a `styles` id to the component in Pages Editor and also added additional styles via the right-hand side panel (e.g. alignment, classname). In this case, we receive a merged JSON including the `styles` id, which is incorrect and should be addressed - likely by a different team - likely it should be done before our release, otherwise customers will report a new item, please create an item / test and track it. Currently we just set a huge string as a class.
> 
> Image

@illiakovalenko 

First comment is valid, I have addressed it in order to cover the scenario of `styles` as an object.

On the second comment, this is inherently wrong. Styles should be a plain class string or a proper object, not a stringified JSON blob. This needs to be fixed on the Pages/Layout Service side.

packages/content/src/layout/utils.ts
```diff
@@ -1,6 +1,67 @@
...
+ * @returns {string | undefined} normalized string value, or undefined when not extractable
+ * @public
+ */
+export function getRenderingParamString(value: unknown): string | undefined {
+  if (value === null || value === undefined) {
+    return undefined;
+  }
+
+  if (typeof value === 'string') {
+    if (value.startsWith('{')) {
```
illiakovalenko: I'm not convinced we need to handle this unexpected scenario (json string) on our side. It appears to be a bug that should be addressed by the owning team first. I'd suggest raising the issue and aligning with Product on the expectations. Rather than introducing workarounds, we should resolve the root cause and ensure the data is generated correctly upstream. 
We don't need to make our things more complex.
What is your opinion?

packages/content/src/layout/utils.ts
```diff
@@ -1,6 +1,67 @@
...
+  return undefined;
+};
+
+/**
+ * Normalizes a rendering param value to a string.
+ * Layout Service may return DetailedRenderingParams as objects or JSON strings
+ * instead of plain strings (e.g. Styles, CSSStyles, GridParameters).
+ * @param {unknown} value rendering param value
+ * @returns {string | undefined} normalized string value, or undefined when not extractable
+ * @public
```
illiakovalenko: I think it should be considered as a "private" utility

illiakovalenko (changes requested):

packages/content/src/layout/utils.ts
MenKNas: I totally agree. Let's discuss this with Product. If the fix should be done upstream by some other team we can revert this on our side.

illiakovalenko (approved):
Looks good! 👍

PR #502, opened 2026-06-08
[nextjs-cache-components-template] Fix Sitecore 404 fetching during SSG and runtime

MenKNas:
- [ ] Changelog updated

## Description / Motivation
This PR fixes [JSS-9706](https://sitecore.atlassian.net/browse/JSS-9706) and [JSS-9708](https://sitecore.atlassian.net/browse/JSS-9708) in the `nextjs-app-router-cache-components` template.

It ensures cached page/error reads run before dynamic APIs (setRequestLocale, draftMode) so SSG `notFound()` and production 404s work with Cache Components. It also isolates draftMode in `<Suspense>` for Bootstrap, aligns error-page cache tags with real Sitecore item paths, and adds a `defaultSite` fallback in `sitecore.config.ts.example`.

## Testing Details
- [ ] Unit Test Added

## Types of changes

[JSS-9706]: https://sitecore.atlassian.net/browse/JSS-9706?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ
[JSS-9708]: https://sitecore.atlassian.net/browse/JSS-9708?atlOrigin=eyJpIjoiNWRkNTljNzYxNjVmNDY3MDlhMDU5Y2ZhYzA5YTRkZjUiLCJwIjoiZ2l0aHViLWNvbS1KU1cifQ


packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/page.tsx
```diff
@@ -42,6 +43,8 @@ export default async function Page({ params, searchParams }: PageProps) {
 
   // If the page is not found, return a 404
   if (!page) {
+    // Ensure site/locale are available to segment not-found.tsx during SSG and runtime.
+    setCachedPageParams({ site, locale });
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/page.tsx]: We already set the cached params in `layout.ts`, following the same approach as the base App Router template. Is there a reason this template needs to handle it differently and additionally set cache params in this branch?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -1,19 +1,22 @@
...
-import scConfig from 'sitecore.config';
+import { resolveSitecoreRouteContext } from 'lib/cache/resolve-sitecore-route-context';
 import Layout from 'src/Layout';
 import Providers from 'src/Providers';
 import { NextIntlClientProvider } from 'next-intl';
 
 export default async function NotFound() {
-  const { site, locale } = getCachedPageParams();
+  const cached = getCachedPageParams();
+  const context = resolveSitecoreRouteContext(cached);
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: I would like to better understand the reasoning behind these changes, as they seem to add extra complexity. What necessitates introducing a more complex context resolution mechanism? While in app-router basic template it just works

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/layout.tsx
```diff
@@ -9,11 +8,10 @@ export default async function SiteLayout({
   params: Promise<{ site: string }>;
 }) {
   const { site } = await params;
-  const { isEnabled } = await draftMode();
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/layout.tsx:12]: It's not the best practice to recommend consumers to access internal cookies directly in the application (The changes you did in Bootstrap.tsx). That approach should be limited to the SDK only.
You can access `draftMode` by wrapping the Bootstrap part in `Suspense`, from the first glance it doesn't break anything

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
```diff
@@ -26,11 +27,13 @@ type GetSitecoreErrorPageParams = {
   code: ErrorPage;
 };
 
-/**
- * Loads SXA-style Sitecore error pages with Next.js Cache Components and the same tag strategy as
- * {@link getSitecorePage}, so webhook / `revalidateTag` flows can invalidate updated error experiences.
- */
-export async function getSitecoreErrorPage(params: GetSitecoreErrorPageParams): Promise<Page | null> {
+type CachedSitecoreErrorPageParams = {
```
illiakovalenko: We currently have two identical interfaces. Please revisit and better to leave this one `GetSitecoreErrorPageParams`

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
illiakovalenko: I probably missed this earlier, but I have a question. We already have the item path and item ID, which can map to the regular page route. Why do we need to introduce `personalizedPathname` and a custom error page tag that isn't recognized by the revalidation route handler?

We already have the site and locale and can retrieve the page information, so we should be able to use the same cache tag as the regular page route. Please revisit this approach, I think `personalizedPathname` can be fully removed, unless I'm missing some requirement.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
```diff
@@ -50,3 +53,29 @@ export async function getSitecoreErrorPage(params: GetSitecoreErrorPageParams):
...
+
+  try {
+    return await getCachedSitecoreErrorPage(resolved);
+  } catch {
+    try {
+      return await client.getErrorPage(params.code, { site, locale });
+    } catch {
+      return null;
+    }
+  }
```
illiakovalenko: I'm not sure I fully understand this part. My understanding is that if we're unable to fetch the error page, we shouldn't fall back and attempt to fetch it again. In that case, the error should simply be propagated.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
```diff
@@ -26,11 +27,13 @@ type GetSitecoreErrorPageParams = {
...
- * {@link getSitecorePage}, so webhook / `revalidateTag` flows can invalidate updated error experiences.
- */
-export async function getSitecoreErrorPage(params: GetSitecoreErrorPageParams): Promise<Page | null> {
+type CachedSitecoreErrorPageParams = {
+  site: string;
+  locale: string;
+  code: ErrorPage;
+};
+
+async function getCachedSitecoreErrorPage(params: CachedSitecoreErrorPageParams): Promise<Page | null> {
```
illiakovalenko: Can we simplify this back to a single function? `getSitecoreErrorPage` is still using the cached function internally, so I'm not sure the extra abstraction provides any benefit

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
```diff
@@ -50,3 +53,29 @@ export async function getSitecoreErrorPage(params: GetSitecoreErrorPageParams):
...
 }
+
+/**
+ * Loads Sitecore error pages with Next.js Cache Components and the same tag strategy as
+ * {@link getSitecorePage}, so webhook / `revalidateTag` flows can invalidate updated error experiences.
+ * Returns `null` when site is missing or the error page fetch fails.
+ */
+export async function getSitecoreErrorPage(params: GetSitecoreErrorPageParams): Promise<Page | null> {
+  const site = params.site?.trim();
+  const locale = params.locale?.trim() || scConfig.defaultLanguage;
```
illiakovalenko: Can you please clarify why we need to trim these values? Both `site` and `locale` come from Sitecore AI, and I can't recall us doing this on any other pages.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/Bootstrap.tsx
illiakovalenko, line 1: Please see my related comment in `layout.tsx` for more details. I don't think we can apply these changes as proposed.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -15,6 +16,9 @@ export default async function NotFound() {
     code: ErrorPage.NotFound,
   });
 
+  // After the cached error-page fetch so next-intl's cached dictionary read does not hit DYNAMIC_SERVER_USAGE.
+  setRequestLocale(`${site || scConfig.defaultSite}_${locale || scConfig.defaultLanguage}`);
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: Please, define shared variables
```ts
const resolvedSite = site || scConfig.defaultSite;
const resolvedLocale = locale || scConfig.defaultLanguage;

....
```

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -15,6 +16,9 @@ export default async function NotFound() {
     code: ErrorPage.NotFound,
   });
 
+  // After the cached error-page fetch so next-intl's cached dictionary read does not hit DYNAMIC_SERVER_USAGE.
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: This doesn't clearly explain the purpose of this action. It mainly describes a problem rather than the intent. I would recommend explicitly clarifying the usage

```suggestion
  // Set after the cached error-page fetch so next-intl's cached dictionary read resolves the locale from here instead of falling back to a Dynamic API.
```

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
MenKNas: That's a leftover from a temporary fix I was testing, will remove.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
MenKNas: I think this is right so I removed synthetic personalizedPathname paths. Error-page route tags now use `context.itemPath` from the layout response same as `getSitecorePage`.

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
```diff
@@ -2,46 +2,36 @@ import { collectSitecorePageCacheTags, ErrorPage, type Page } from '@sitecore-co
...
 export async function getSitecoreErrorPage(params: GetSitecoreErrorPageParams): Promise<Page | null> {
   'use cache';
 
   const { site, locale, code } = params;
   const page = await client.getErrorPage(code, { site, locale });
-  const personalizedPathname = personalizedPathnameForErrorCode(code);
+
+  const sitecore = page?.layout?.sitecore;
+  const itemPath = sitecore?.context?.itemPath;
+  const personalizedPathname =
```
illiakovalenko: Can you please clarify why we still need to pass the personalized pathname? It's not required on a regular page, and we already have the path available. I expected that we can entirely remove `personalizedPathname` parameter, it's quite confusing

illiakovalenko (changes requested):

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/cache/get-sitecore-error-page.ts
MenKNas: It was kept there by mistake. Removed

illiakovalenko (approved):
Looks good 👍

PR #504, opened 2026-06-08
[cli][create-content-sdk-app] Exclude temporary .sitecore/* files from templates

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

Removed the generated default `.sitecore` map files from Next.js templates and updates map generation to recreate the `.sitecore` directory when it is missing.
This keeps generated apps cleaner while ensuring `generate-map` and import map generation still work after the folder is removed or ignored.

## Testing Details

- [x] Unit Test Added

Built and tested successfully. Added new unit tests for import map and generate map to verify the `.sitecore` directory is created when missing. All existing tests pass with the new coverage.

## Types of changes


packages/content/src/tools/codegen/import-map.ts
```diff
@@ -376,6 +376,15 @@ const prepImportMaps = async (paths: string[], separateMaps?: boolean): Promise<
   ];
 };
 
+// Ensures the .sitecore directory exists, creating it if necessary.
+const ensureImportMapPath = (filePath: string) => {
```
illiakovalenko: Can we check and create this folder if neccessary during the "build" command execution? at the very start, so we don't need to check it in every command

illiakovalenko (changes requested):

packages/content/src/tools/codegen/import-map.ts
sc-naveenhedallaarachchi: The "build" command itself requires the .sitecore folder to be there, and the generate-map can also run standalone or in watch/dev flows outside sitecore-tools. Do you have any suggestions?

illiakovalenko: > The "build" command itself requires the .sitecore folder to be there, and the generate-map can also run standalone or in watch/dev flows outside sitecore-tools. Do you have any suggestions?

All commands are executed as part of `sitecore-tools project build`. Could you please review the implementation details of this command? I believe we should be able to validate the `.sitecore` folder there.
If needed, this check could also be performed during component map generation, since it is a separate sitecore-tools CLI command.

sc-naveenhedallaarachchi: The requested changes have been done.

.changeset/sour-onions-drop.md
```diff
@@ -0,0 +1,6 @@
+---
+'create-content-sdk-app': minor
+'@sitecore-content-sdk/cli': minor
+---
+
+[cli][create-content-sdk-app] Update build and component map generation to recreate the .sitecore directory when it is missing after generated template map files are removed, and add test coverage for both flows.
```
illiakovalenko: Minor suggestion for improving the changelog entry: we typically don't mention details such as unit test coverage in changelog items.

```suggestion
[cli][create-content-sdk-app] Update build and component map generation to recreate the .sitecore directory when it is missing.
```

illiakovalenko (approved):
Looks good 👍

PR #505, opened 2026-06-08
[Angular CSDK] Angular Editing and component guards

art-alexeyenko:
Provides Angular editing and preview support.
Additionally, ports Angular guards support from JSS.

The editing flow is:
- Editing middleware accepts editing data, and makes an indicator of editing mode being active.
- Rewrite path to incoming editing path (no round-robin we have in next)
- page loader uses previewData to render the page
- client scripts/data loaded too
- field directives converted to structural directives to support extra editing markup


packages/create-content-sdk-app/src/templates/angular/src/content-sdk/loaders/page.loader.ts
```diff
@@ -12,9 +12,18 @@ import { getClient } from '../client/sitecore-client';
...
+    return getClient().getPreview(previewData);
+  }
+  const locale = (context.params['locale'] as string | undefined) || scConfig.defaultLanguage;
   const { nonLocalePath } = splitLocaleFromPath(context.url, scConfig.angular.locales);
-  const page = await resolveSitecorePage(nonLocalePath, scConfig, getClient(), { locale });
+  const site = scConfig.defaultSite;
+
+  const page = previewData
+    ? await getClient().getPreview(previewData)
+    : await getClient().getPage(nonLocalePath, { locale, site });
```
MenKNas: I am a little confused with this block. Why do we have an early return in the case of previewData? And then why do we check again if previewData exists and have the preview response assigned to the page? I guess we can consolidate in a single block and make this cleaner unless I am missing something.

packages/create-content-sdk-app/src/templates/angular/src/content-sdk/loaders/page.loader.ts
```diff
@@ -12,9 +12,18 @@ import { getClient } from '../client/sitecore-client';
...
+
   if (!page) {
     throw new NotFoundNavigationError();
   }
```
MenKNas, line 26: Also this block could be combined too. We could do something like:

```
const previewData = getEditingPreviewData(context.requestContext);
if (previewData) {
  const page = await getClient().getPreview(previewData);
  if (!page) throw new NotFoundNavigationError();
  return page;
}
const page = await getClient().getPage(nonLocalePath, { locale, site });
```

packages/angular/src/components/placeholder/sc-placeholder.component.ts
```diff
@@ -33,13 +42,53 @@ import { ScHiddenRenderingComponent } from '../sc-hidden-rendering.component';
...
+ * `fields`, `params`, and `rendering`).
+ *
+ * **Editing chrome (Metadata mode only).** When the page is in editing mode, the placeholder
+ * emits Sitecore Pages chrome markers using a declarative `<ng-template>` rendered into the
+ * same `ViewContainerRef` that hosts the dynamic child components. The structure matches
+ * Metadata-mode output:
+ *
+ * ```
+ * <code class="scpm" chrometype="placeholder" kind="open" id="…" />   ← once, outer
+ *   <code class="scpm" chrometype="rendering" kind="open" id="<uid>" data-csdk-component-runtime="server" />
```
MenKNas: Do we need `data-csdk-component-runtime="server"` to be mentioned in this case? This is not part of the `ng-template` which is correct and we don't need it right?

packages/angular/src/components/placeholder/sc-placeholder.component.ts
```diff
@@ -72,51 +121,114 @@ export class ScPlaceholderComponent {
...
+
+    // Empty placeholder: in Metadata edit mode, still wrap with an outer placeholder
+    // pair so Pages can discover and target the empty region. Otherwise emit nothing.
+    if (rawRenderings.length === 0) {
+      if (isEditing && editingChromeBlock) {
+        this.emitOuterChrome(container, editingChromeBlock, MetadataKind.Open, rendering, name);
+        this.emitOuterChrome(container, editingChromeBlock, MetadataKind.Close, rendering, name);
       }
+      return;
+    }
```
MenKNas, line 192: Shouldn't we call `resetEditorChromes()` here to trigger a chrome reset?

packages/create-content-sdk-app/src/templates/angular/src/server.ts
```diff
@@ -67,6 +70,25 @@ app.use(
...
+    metadataImport: () => import('.sitecore/metadata.json'),
+  })
+);
+
+/**
+ * Editing render endpoint (`/api/editing/render`). Rewrites `req.url` to the
+ * editor's requested route, stashes the preview payload on the request, then
+ * lets the Angular SSR engine render the page in-process.
+ */
+app.use(createEditingRenderMiddleware());
```
MenKNas, line 90: Ignore this if it's irreleveant, just a question...do we need to pass a `resolvePageUrl` like `/${previewData.language}${itemPath}` so editing hits `/en/about` instead of `/about` when using locale prefixed routes?

packages/angular/src/server/middleware/editing-render-middleware.ts
```diff
@@ -0,0 +1,298 @@
...
+/**
+ * Editing-aware request. The render middleware attaches `scEditing` to allow
+ * downstream code to detect the editing branch without reparsing headers.
+ * @public
+ */
+export interface ExpressEditingRequest extends ExpressRequest {
+  scEditing?: EditingPreviewData;
+}
+
+const REQUIRED_PARAMS = ['sc_site', 'sc_itemid', 'sc_lang', 'route', 'mode'] as const;
```
MenKNas, line 80: Again ignore if already discussed, we are intentionally deferring DL for Angular right?

packages/angular/src/components/field-directives/base-field.directive.ts
```diff
@@ -0,0 +1,141 @@
...
+  /** Consumer-supplied empty-field template; takes precedence over {@link defaultEmptyComponent}. */
+  abstract readonly emptyFieldEditingTemplate: Signal<TemplateRef<unknown> | undefined>;
+  /** Default component rendered when the field is empty in editing mode and no template is supplied. */
+  protected abstract readonly defaultEmptyComponent: Type<unknown>;
+
+  constructor() {
+    effect(() => {
+      this.field();
+      this.updateView();
+    });
```
MenKNas, line 72: Should `emptyFieldEditingTemplate` also be tracked in this `effect()` ?

packages/angular/src/components/field-directives/metadata-chrome.component.ts
```diff
@@ -0,0 +1,27 @@
...
+    '[attr.chrometype]': '"field"',
+    '[class]': '"scpm"',
+  },
+})
+export class FieldMetadataMarkerComponent {
+  @Input()
+  metadata?: Record<string, unknown>;
+
+  @HostBinding('attr.kind')
+  @Input()
```
MenKNas: Is there a reason why we are using `@Input` in some cases and `input()` in others? If not maybe we could aim for consistency by turning everything into signals?

MenKNas (changes requested):

MenKNas:
Really great job as always @art-alexeyenko thank you! Few minor comments/questions and we are ready to go

packages/angular/src/server/middleware/editing-render-middleware.ts
art-alexeyenko: yes, DL will come later - we need some time to figure out correct component generation for Angular.

packages/create-content-sdk-app/src/templates/angular/src/server.ts
art-alexeyenko: honestly we could go either way. When in editing flow, we will pass locale to request context/preview data and that's where it will be resolved from - instead of the route during normal editing.
but for consistency, yeah, this makes sense.

packages/angular/src/components/field-directives/metadata-chrome.component.ts
art-alexeyenko: an omission, fixed

packages/angular/src/components/field-directives/base-field.directive.ts
MenKNas: @art-alexeyenko should we also track `isEditing()` here?

packages/angular/src/server/middleware/editing-render-middleware.ts
```diff
@@ -266,7 +266,9 @@ export function createEditingRenderMiddleware(
       }
 
       const decodedRoute = flatQuery.route ?? '';
-      const targetRoute = options?.resolvePageUrl?.(decodedRoute, previewData) ?? decodedRoute;
+      const targetRoute =
+        options?.resolvePageUrl?.(decodedRoute, previewData) ??
+        `/${previewData.language}${decodedRoute.startsWith('/') ? '' : '/'}${decodedRoute}`;
```
MenKNas, line 271: Is always prefixing `/${language}` the right default for apps without locale-prefixed routes?

MenKNas (approved):

packages/angular/src/components/field-directives/base-field.directive.ts
art-alexeyenko: isEditing should not be changing during navigation - we either access site as editing or not - so we can skip it.

packages/angular/src/server/middleware/editing-render-middleware.ts
art-alexeyenko: discussed, leaving as it is now

PR #508, opened 2026-06-10
[content] Fix imageParams breaking preview context images

sc-shakyawijerathne:
## Description

- [x] Changelog updated

## Description / Motivation

Fixes a bug where images in preview context would fail to load when using `imageParams` with the `NextImage` component. 

**Root Cause:** Preview context URLs include authentication parameters (`ttc`, `tt`, `hash`) that are required by Sitecore's preview endpoint to validate requests. When `imageParams` were applied (e.g., `imageParams={{mw: 100, mh: 50}}`), these authentication parameters were being stripped from the URL, causing images to return 403/404 errors.

**Solution:** Added preview-specific authentication parameters (`ttc`, `tt`, `hash`) to the required params list in `getRequiredParams()`. These parameters are now preserved alongside the requested image optimization parameters, allowing images to load correctly in both preview and production contexts.

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other

## Types of changes


packages/content/src/media/media-api.ts
```diff
@@ -67,9 +67,9 @@ function searchParamsToQueryRecord(sp: URLSearchParams): { [key: string]: string
  * @public
  */
 export const getRequiredParams = (qs: { [key: string]: string | undefined }) => {
-  const { rev, db, la, vs, ts } = qs;
+  const { rev, db, la, vs, ts, ttc, tt, hash } = qs;
 
-  return { rev, db, la, vs, ts };
+  return { rev, db, la, vs, ts, ttc, tt, hash };
```
illiakovalenko, line 72: Could you confirm: when you did the manual test, did you verify the actual resulting image request (e.g. with mw/mh applied via `imageParams`, hash carried over) returns 200 from Edge? I want to rule out that this trades a "missing token" failure for a "signature mismatch" failure (403), which would still leave images broken just with a different error.

If Edge does validate hash against the full querystring, we may instead need to treat URLs containing ttc/tt/hash as "signed preview URLs" and pass them through unchanged (ignoring imageParams) rather than rebuilding the querystring.

illiakovalenko (changes requested):

packages/content/src/media/media-api.ts
sc-shakyawijerathne: Yes, I verified the actual image request returns HTTP 200 from Edge. The browser Network tab shows the image loads successfully with both imageParams (mw/mh) and preview auth params (ttc/tt/hash) present in the URL. The hash validation appears to allow additional optimization parameters.

image


image

illiakovalenko (approved):
👍

PR #509, opened 2026-06-10
Pass sc_variant header in Preview / Edit mode

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
In Edit Mode and Preview Mode, the Content SDK previously resolved personalization (A/B test variants and Page Level Personalization) client-side, by parsing `sc_variant` query parameter into a list of variant ids and
filtering each component's `experiences` map via `personalizeLayout`.

This PR moves that resolution to the server: the SDK now always sends a single `sc_variant` header on
the Preview GraphQL request used by `EditingService.fetchEditingData`, so the Preview GraphQL API returns
the already-personalized layout for the active variant.

The internal contract has been slightly updated: since Edit/Preview does not provide a list of variant IDs, A/B testing and Personalization cannot be configured together. We don't expect to receive a list of ids, a single id is provided per view

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/content/src/client/sitecore-client.ts
```diff
@@ -496,8 +494,6 @@ export class SitecoreClient implements BaseSitecoreClient {
     }
 
     let layout = data.layoutData;
-    const personalizeData = getGroomedVariantIds(variantIds);
-    personalizeLayout(layout, personalizeData.variantId, personalizeData.componentVariantIds);
```
art-alexeyenko, line 500: this will make A/B testing variants to never be displayed. Is this the intended outcome here?

illiakovalenko: @art-alexeyenko yes, it's expected
now we expect to receive a ready and personalized layout, we don't need to resolve experiences in layout for Preview / Edit (see PR description for more details)

art-alexeyenko (approved):

PR #511, opened 2026-06-11
[nextjs][react] Nested components inside the Link component should be rendered even when the LinkField value is empty.

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

This PR implements the `renderChildrenWhenEmpty` prop to the Link component. When set to `renderChildrenWhenEmpty={true}`, renders an empty anchor element with children instead of null when the link field value is empty.

## Testing Details

- [x] Unit Test Added
Comprehensive test coverage added including empty field scenarios, edit mode with metadata, prop filtering for Next.js-specific attributes, and backward compatibility verification.

## Types of changes


illiakovalenko (commented):
-

packages/react/src/components/Link.tsx
```diff
@@ -58,6 +66,13 @@ const LinkComponent: React.FC<LinkProps> = ({
   delete otherProps.editable; // prevent editable from being passed to the DOM
 
   if (isFieldValueEmpty(dynamicField)) {
+    if (renderChildrenWhenEmpty) {
```
illiakovalenko: This `renderChildrenWhenEmpty` block is duplicated verbatim in the `!link` branch below (lines 85-91). Both conditions represent "no usable link value" - could they collapse into a single early-exit (`isFieldValueEmpty(dynamicField) || !link`) with one shared render path, so the two copies don't drift apart later?

packages/react/src/components/Link.tsx
```diff
@@ -58,6 +66,13 @@ const LinkComponent: React.FC<LinkProps> = ({
...
+      return (
+        <a {...otherProps} key="link" ref={ref}>
+          {children}
```
illiakovalenko: If `renderChildrenWhenEmpty` is `true` but no `children` are passed, this still renders `<a></a>` - an empty anchor in the DOM. Minor, but could guard with `children &&` (or return `null`) to avoid the dead markup.

packages/nextjs/src/components/Link.tsx
```diff
@@ -53,10 +60,22 @@ export const Link = forwardRef<HTMLAnchorElement, LinkProps>(
       children,
       internalLinkMatcher = /^\//g,
       showLinkTextWithChildrenPresent,
+      renderChildrenWhenEmpty,
       ...rest
     } = props;
 
     if (!field || (!field.value && !(field as LinkFieldValue).href && !field.metadata)) {
+      if (renderChildrenWhenEmpty) {
```
illiakovalenko: If `renderChildrenWhenEmpty` is `true` but no `children` are passed, this still renders `<a></a>` - an empty anchor in the DOM. Minor, but could guard with `children &&` (or return `null`) to avoid the dead markup.

packages/nextjs/src/components/Link.tsx
```diff
@@ -32,6 +32,13 @@ export type LinkProps = ReactLinkProps & {
    * @default /^\//g
    */
   internalLinkMatcher?: RegExp;
+
+  /**
+   * Renders children even when the link field value is empty.
+   * When true, an empty anchor element containing the children is rendered instead of null.
+   * @default false
+   */
+  renderChildrenWhenEmpty?: boolean;
```
illiakovalenko: We probably don't need to document this property here, since we extend `ReactLinkProps` type, it includes this property

illiakovalenko (changes requested):

packages/nextjs/src/components/Link.tsx
sc-naveenhedallaarachchi: Done

packages/react/src/components/Link.tsx
sc-naveenhedallaarachchi: Done

packages/react/src/components/Link.tsx
sc-naveenhedallaarachchi: Done

packages/nextjs/src/components/Link.tsx
sc-naveenhedallaarachchi: Removed

illiakovalenko (approved):
Looks good 👍

PR #512, opened 2026-06-11
[Angular CSDK] Fix editing frames not applied correctly, dictionary not being used

art-alexeyenko:
Fixes two things, following the QA checkpoint:
- Editing frames not correctly displaying for dynamic placeholder in Pages, in Promo component.
- Dictionary not being utilized when using TranslatePipe from ngx-translate.

### DIFF
--- a/packages/create-content-sdk-app/src/templates/angular/src/app/app.ts
+++ b/packages/create-content-sdk-app/src/templates/angular/src/app/app.ts
@@ -1,6 +1,7 @@
-import { Component, signal } from '@angular/core';
+import { Component, effect, inject, signal } from '@angular/core';
 import { RouterOutlet } from '@angular/router';
-import { ScEditingScriptsComponent } from '@sitecore-content-sdk/angular';
+import { ScEditingScriptsComponent, SitecoreContextService } from '@sitecore-content-sdk/angular';
+import { TranslateService } from '@ngx-translate/core';
 
 @Component({
   selector: 'app-root',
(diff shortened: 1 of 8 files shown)


packages/create-content-sdk-app/src/templates/angular/src/app/app.ts
```diff
@@ -10,4 +11,14 @@ import { ScEditingScriptsComponent } from '@sitecore-content-sdk/angular';
 })
 export class App {
   protected readonly title = signal('angular-sample');
+  private context = signal(inject(SitecoreContextService));
```
MenKNas: Do we need the signal here? The effect already reacts to `page()` so I guess a plain `inject()` is enough?

packages/create-content-sdk-app/src/templates/angular/src/app/components/promo.component.ts
```diff
@@ -58,31 +58,29 @@ function promoProductJsonLd(
...
+          <div *scRichText="secondaryRichText"></div>
+        </div>
         }
+        <div class="field-promolink">
+          <a *scLink="promoLink()"></a>
+        </div>
       </div>
-    </article>
+      @if (jsonLd()) {
+      <app-structured-data [scriptId]="jsonLdScriptId()" [data]="jsonLd()" />
```
MenKNas: Was it intentional here that both Promo variants always show the link? It looks like we are mapping both variants to the same component, is that expected?

MenKNas:
We should also add a `.changeset` file

MenKNas (changes requested):

art-alexeyenko:
> We should also add a `.changeset` file

I don't think we need to changeset a bugfix - Angular feature is not formally released yet.

packages/create-content-sdk-app/src/templates/angular/src/app/components/promo.component.ts
art-alexeyenko: here I went for duplicating templates HTML, but formally introducing two variants. It's a tradeoff - we cannot easily reuse markup the way next.js does in its function components, if we try the template becomes more complex.  So, yes, HTML is duplicated, but the flow is easier to follow overall

MenKNas:
LGTM 🥳

MenKNas (approved):

PR #516, opened 2026-06-15
[Angular CSDK] Multisite, personalize and events 

art-alexeyenko:
- Personalization support
- Multisite support
  - Both implemented through Express middleware
  - Both support server and browser navigation
    - in case of browser navigation, variant and site data is populated for the `/_data` request, before resolving loader data for LoaderDataService
- Expanded cli.config logic to accept config object and call `generateSites()`
   - Angular config logic moved to separate submodule - in order to avoid importing Angular inject and other side effects when not needed.
- Refactors in loader logic for clearer flow and reducing type duplication
- Injectable Analytics wrappers for browser and server accessible within Angular context
- Add events to form component
- Add CDP page view component.


packages/angular/src/server/middleware/multisite-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+            (data.cookies?.[SITE_KEY] as string | undefined)) ||
+          siteResolver.getByHost(hostname).name;
+      }
+
+      if (!resolvedSite) {
+        resolvedSite = options.defaultSite ?? '';
+      }
+      req.scParams
+        ? (req.scParams.siteName = resolvedSite)
+        : (req.scParams = { siteName: resolvedSite, variantId: DEFAULT_VARIANT });
```
illiakovalenko: I'm not really sure that setting `variantId` in the multisite middleware is the right approach. The multisite middleware shouldn't need to be aware of other features.

packages/angular/src/server/middleware/multisite-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+      let resolvedSite: string;
+      const hostname = data.hostname || getHostname(req);
+
+      const isSitecorePreview = data.cookies?.[PREVIEW_KEY];
+
+      if (isSitecorePreview) {
+        // This cookie is required to be set in the Sitecore Preview mode to support navigation
+        // and preserve the site name
+        resolvedSite = data.cookies?.[SITE_KEY]!;
+      } else {
```
illiakovalenko: Based on my experience and the preview-related issues I've worked on, I'd recommend introducing a dedicated preview middleware. It would provide a proper place for preview-specific logic, including setting the site context, rather than adding that responsibility to the multisite middleware.

One of the reasons is that developers can easily remove the multisite middleware if they don't need multisite support. I have a plan to introduce a similar approach in Next.js, as we've run into the same issue there.

packages/angular/src/server/server-loader-runner.ts
```diff
@@ -30,23 +32,42 @@ export class ServerLoaderRunner {
...
       return { kind: 'error', status: 500, message: `No loader registered for id "${loaderId}"` };
     }
 
-    const ctx: LoaderContext = { url, params, query, requestContext: angularRequestContext };
+    const scParams = {
+      siteName: csdkRequestData?.scParams?.siteName ?? this.config.defaultSite,
+      variantId: csdkRequestData?.scParams?.variantId ?? DEFAULT_VARIANT,
+      componentVariantIds: csdkRequestData?.scParams?.componentVariantIds ?? [],
+    };
+
```
illiakovalenko, line 63: Can we just merge whatever we have in params? I suppose it should be resolved at this stage.
Currently it means we should keep these params aligned across multiple parts of angular package

packages/angular/src/server/middleware/personalize-middleware.ts
```diff
@@ -0,0 +1,338 @@
...
+      ) {
+        debug.personalize('skipped (%s is not a layout route)', path);
+        return next();
+      }
+      if (data.cookies?.[PREVIEW_KEY]) {
+        // No need to personalize for preview (layout data is already prepared for preview)
+        debug.personalize('skipped (preview)');
+        return next();
+      }
+      if ((options.skipForBot ?? true) && data.cookies?.[BOT_DETECTION_COOKIE]) {
```
illiakovalenko: Currently, there is nothing in Angular that triggers `botPageView`. No part of the pipeline sets this cookie, and there is no `createBotTrackingMiddleware` or equivalent `botPageView` call in the Angular package.
As a result, the cookie check will never be triggered.

packages/angular/src/server/middleware/multisite-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+ * @returns {ExpressMiddleware} The multisite middleware.
+ */
+export function createMultisiteMiddleware(options: MultsiteMiddlewareOptions): ExpressMiddleware {
+  const siteResolver = new SiteResolver(options.sites ?? []);
+  return (req: CsdkExpressRequest, res: ExpressResponse, next: ExpressNextFunction) => {
+    try {
+      // For browser loader navigations (/_data) routing data comes from the loader payload, not
+      // the request; getMiddlewareRequest normalizes both into path/query/data.
+      const { path, query, data } = getMiddlewareRequest(req);
+      if (path.startsWith('/api')) {
```
illiakovalenko: The multisite middleware skips only /api paths. It does not skip:
- File requests (.js, .css, .png, .woff2, etc.) - site resolver runs on every static asset
- /sitecore/* paths
Every static file request hits SiteResolver.getByHost() and sets a cookie - both wasteful and potentially incorrect.

packages/angular/src/server/middleware/personalize-middleware.ts
```diff
@@ -0,0 +1,338 @@
...
+        debug.personalize('skipped (no personalization configured)');
+        return next();
+      }
+
+      if (isPrefetch(req)) {
+        // Personalized, but this is a prefetch request.
+        // Don't execute a personalize request; otherwise, the metrics for component A/B experiments would be inaccurate.
+        // Disable caching to force revalidation on navigation (personalization WILL be influenced).
+        debug.personalize('skipped (prefetch)');
+        res.setHeader?.('Cache-Control', 'no-store, must-revalidate');
```
illiakovalenko, line 252: We only have here Cache-Control on prefetch requests, and never set x-proxy-cache. After a successful personalization that identifies a variant, Angular does not signal to upstream caches that this response is variant-specific. In environments with a CDN or reverse proxy in front of the Angular SSR server, the proxy might cache and serve one visitor's personalized response to another. Can you check it please? We have it in nextjs [proxy](https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/proxy/personalize-proxy.ts#L207)

packages/angular/src/server/middleware/multisite-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+      if (isSitecorePreview) {
+        // This cookie is required to be set in the Sitecore Preview mode to support navigation
+        // and preserve the site name
+        resolvedSite = data.cookies?.[SITE_KEY]!;
+      } else {
+        resolvedSite =
+          (query[SITE_KEY] as string | undefined) ||
+          (query.site as string | undefined) ||
+          (options.useCookieResolution &&
+            options.useCookieResolution(req as RequestInit) &&
```
illiakovalenko: `RequestInit` is the options object for the browser Fetch API (fields: method, headers, body, cache, credentials). It is not a request object. The Angular `CsdkExpressRequest` is an Express-style request. Casting one to the other is incorrect - at runtime, `useCookieResolution` receives the Express request, but the TypeScript signature says `RequestInit`. Any type-safe implementation of useCookieResolution that uses, say,
```
(req: RequestInit) => req.credentials === 'include'
```
would fail at runtime.

packages/angular/src/server/middleware/personalize-middleware.ts
```diff
@@ -0,0 +1,338 @@
...
+            },
+            adapter: personalizeServerAdapter(httpReq, httpRes),
+          }),
+        ],
+      });
+
+      const geo = options.extractGeoDataCb ? await options.extractGeoDataCb(req) : undefined;
+      const params = getExperienceParams(
+        query,
+        (data.headers?.referer as string) || '',
```
illiakovalenko: Can we also fallback to `req.referrer`? I suppose it's present like in next.js

packages/angular/src/server/middleware/multisite-middleware.spec.ts
```diff
@@ -0,0 +1,22 @@
...
+  });
+  it('should use provided sites collection', () => {});
+  it('should resolve site from request by cookie', () => {});
+  it('should resolve site from request by query parameter', () => {});
+  it('should resolve site from request by hostname', () => {});
+  it('should skip when in edting,preview mode', () => {});
+  it('should skip when in api route', () => {});
+  it('should skip when in static files', () => {});
+  it('should skip when in health check', () => {});
+  it('should skip when in static files', () => {});
```
illiakovalenko: Unit tests are missing

packages/angular/src/server/middleware/multisite-middleware.ts
```diff
@@ -0,0 +1,120 @@
...
+import { DEFAULT_VARIANT } from '@sitecore-content-sdk/content/personalize';
+import debug from '../../debug';
+import { PREVIEW_KEY } from '@sitecore-content-sdk/content/editing';
+import { getMiddlewareRequest } from '../utils';
+
+type MultsiteMiddlewareOptions = SitecoreConfig['multisite'] & {
+  sites?: SiteInfo[];
+  defaultSite?: string;
+  skip?: (req: CsdkExpressRequest) => boolean;
+};
```
illiakovalenko: Let's add tsdoc

packages/create-content-sdk-app/src/templates/angular/.gitignore
```diff
@@ -40,6 +40,10 @@ yarn-error.log
 # Loader cache (unstorage fs driver)
 .cache/
 
+# sitecore temp files (generated by sitecore-tools build)
+.sitecore/*
+!.sitecore/component-map.ts
```
illiakovalenko: Can we remove "component-map" entry? Since we decided to keep ".sitecore" folder empty

packages/create-content-sdk-app/src/templates/angular/src/server.ts
```diff
@@ -89,11 +86,44 @@ app.use(
...
+    defaultSite: config.defaultSite,
+  })
+);
+
 /**
  * Loader data endpoint (/_data). Must use the same loaders as the client registry
  * so client-side navigation can fetch route data via POST /_data.
  */
-app.use(createLoaderDataServiceMiddleware({ loaders: LOADERS, cache: loaderCache }));
+app.use(createLoaderDataServiceMiddleware(config, { loaders: LOADERS, cache: loaderCache }));
```
illiakovalenko, line 143: Do we need to follow a different convention here? Could we simply pass an additional `defaultSite` property in the first argument (object)?
As far as I can see, that’s the only missing piece at the moment (unless I’m overlooking something).

packages/create-content-sdk-app/src/templates/angular/src/server.ts
```diff
@@ -89,11 +86,44 @@ app.use(
...
+);
+
+/**
+ * Personalize middleware. Identifies page/component variants for the request via
+ * Sitecore CDP and writes them onto `req.scParams` so the page loader fetches the
+ * personalized layout and the loader cache keys per variant.
+ *
+ * NOTE: Personalize requires Edge configuration (contextId/clientContextId) and
+ * cannot work with local containers — the middleware disables itself when it is
+ * missing, which is expected in local development.
```
illiakovalenko: I think it's pretty obvious 🙂
```suggestion
 * cannot work with local containers.
```

packages/create-content-sdk-app/src/templates/angular/src/server.ts
```diff
@@ -89,11 +86,44 @@ app.use(
...
+ */
+app.use(
+  createPersonalizeMiddleware({
+    ...config.personalize,
+    ...config.api.edge,
+    locales: config.angular.locales,
```
illiakovalenko, line 132: I have a general comment, not specifically related to this PR. Could we consider moving these properties to the top level instead of nesting them under the `angular` property? Similar to how we do it in Next.js (e.g. generateStaticPaths)

Since the package is already imported from `/angular`, it implicitly indicates that the configuration is Angular-specific and can include Angular-related parts.

packages/content/src/site/utils.ts
```diff
@@ -18,6 +18,44 @@ export type SiteRewriteData = {
   siteName: string;
 };
 
+/**
```
illiakovalenko, line 21: It looks like we don't have unit test coverage for these functions

packages/angular/src/server/utils.ts
illiakovalenko, line 1: Could we please consider covering this with unit tests?

packages/angular/src/components/sc-form.component.ts
```diff
@@ -94,6 +98,32 @@ export class ScFormComponent {
...
+  /**
+   * Listens for the form's `form:engage` events (VIEWED / SUBMITTED) and dispatches them through
+   * the {@link SITECORE_ANALYTICS} façade, which lazily initializes the events SDK on first use
+   * (a no-op on the server and when analytics is disabled). This is the analytics seam for
+   * Angular
+   * @param {HTMLElement} formElement - Container holding the rendered form markup.
+   * @param {string} [componentId] - Rendering uid used as the CDP component instance id.
+   * @param {AbortSignal} [signal] - Removes the listener when the component is destroyed.
+   */
+  private subscribeToFormEvents(
```
illiakovalenko: Could we extend (e.g. add 3rd param) our existing core form implementation instead (`subscribeToFormEvents`)? We should avoid duplicating it

packages/angular/src/lib/analytics/sitecore-analytics.ts
illiakovalenko, line 1: Do you plan to add unit tests for the new analytics adapters as well?

illiakovalenko (changes requested):
Great job! 👍
There are a lot of changes here. Please see my suggestions below.

packages/angular/src/server/middleware/multisite-middleware.ts
art-alexeyenko: in that case our base model for useCookieResolution is wrong - we shouldn't use FetchAPI interface there.
I've added the change (should be non-breaking, IncomingMessage covers both Fetch API and express)

packages/create-content-sdk-app/src/templates/angular/src/server.ts
art-alexeyenko: Both the Angular and the Express layers call Loader Runner - and Loader Runner needs config to provide fallback values to site and language if none were provided (maybe more in the future?). We cannot inject config into runner unfortunately so we need to pass config through loader data service middleware.

packages/create-content-sdk-app/src/templates/angular/src/server.ts
art-alexeyenko: the logic for using the `angular` subsection is to adhere to multi-framework philosophy we taken with configs. We have the base configuration core that is reusable between framework (if necessary). `angular` subsection is not - and having it defined explicitly informs about this.

packages/angular/src/lib/analytics/sitecore-analytics.ts
art-alexeyenko: yes, done

packages/angular/src/server/middleware/multisite-middleware.ts
art-alexeyenko: I've made multisite middleware skip instead. When preview navigation scenarios are more clear we can add the preview middleware.

packages/angular/src/server/middleware/multisite-middleware.ts
art-alexeyenko: never mind - it breaks next.js package. I need to think more how we can approach this. Please review the rest in the meantime

packages/angular/src/server/middleware/personalize-middleware.ts
illiakovalenko: @art-alexeyenko It looks like this hasn't been addressed yet. Could you please double-check?

packages/angular/src/server/middleware/multisite-middleware.ts
illiakovalenko: @art-alexeyenko Can we add some tsdoc for parameters, please?

packages/create-content-sdk-app/src/templates/angular/src/server.ts
illiakovalenko: Yeah, the purpose is clear. In Next.js, we didn't introduce a separate property, so I'm just wondering if it's really needed. That said, it doesn't hurt.

packages/angular/src/server/middleware/personalize-middleware.ts
illiakovalenko: @art-alexeyenko It looks like this hasn't been addressed yet. Could you please clarify?

packages/angular/src/server/server-loader-runner.ts
illiakovalenko: @art-alexeyenko Do we need to enumerate all fields explicitly here (now we have default values)? The runner is a general mechanism but now needs to know about every feature's params (e.g. variantId, componentVariantIds are personalization concerns).
If we change LoaderContext.scParams from Required<CsdkRequestParams> to plain CsdkRequestParams, the runner only needs to provide the siteName fallback (which it legitimately owns via this.config.defaultSite). Everything else flows through from csdkRequestData?.scParams automatically - and adding a new feature field in the future won't require touching this file. What do you think?

packages/angular/src/server/middleware/personalize-middleware.ts
art-alexeyenko: right. bot detection is a separate task, so I didn't focus on it here

packages/angular/src/config/define-config.ts
```diff
@@ -43,6 +38,13 @@ export interface AngularSitecoreConfigInput extends Omit<SitecoreConfigInput, 'r
       revalidate?: number;
     };
   };
+  multisite?: {
+    enabled?: boolean;
```
illiakovalenko, line 42: One small improvement, can we please extend existing `multisite` config instead of `Omit` first and then override? So we don't need to list and duplicate all the props here

packages/angular/src/loaders/models.ts
```diff
@@ -1,13 +1,25 @@
...
- * Request context containing information from the incoming HTTP request.
+ * Content SDK request params like site name, variant ids
+ * @public
+ */
+export interface CsdkRequestParams {
+  /** Site name. Resolved from the request hostname */
+  siteName?: string;
+  /** Variant id. Either resovled from route or set to default variant id name */
+  variantId?: string;
+  componentVariantIds?: string[];
```
illiakovalenko, line 16: Please, add tsdoc

middleware-refactor-review.md
illiakovalenko, line 1: Let's exclude this file

illiakovalenko (approved):
Looks good 👍
Please, see a couple of comments below before merge

packages/angular/src/config/define-config.ts
art-alexeyenko: Alas, we can't. Current approach is least invasive - we can covert and pass config objects into internal packages in a more straightforward way. If we re-type multisite section only, we need to explicitly pass the multisite section into `content` methods: in define-config and define-cli-config both.
Ideally we want to create a common interface between Express and Nextjs Request type in `content` - but it would be a breaking change right now.

PR #518, opened 2026-06-16
[nextjs] Sitemap index route passes empty string instead of undefined, breaking sitemap-index.xml loading

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

/sitemap-index.xml was resolving to id: '' (same as /sitemap.xml) because only numeric paths matched the sitemap ID regex. That caused getSiteMap to look up a default sitemap.xml before falling back to index generation.

This change maps /sitemap-index.xml to id: undefined, so the handler skips the file lookup and generates the sitemap index directly. Numeric paths (/sitemap-1.xml) and /sitemap.xml behavior are unchanged.

**Why we can't remove the `|| ''` fallback and let the id be undefined :** Removing it would make `sitemap.xml` and `sitemap-index.xml` pass `undefined` as the id. That would always skip the [if (sitemapPath)](https://github.com/Sitecore/content-sdk/blob/dev/packages/content/src/client/sitecore-client.ts#L647) branch in `sitecore-client.ts` even if there is sitemap data, so existing default sitemap XML data would not be fetched or returned.

## Testing Details

- [x] Unit Test Added

Added and updated unit tests for sitemap route validation and default sitemap id handling.

## Types of changes


.changeset/gentle-rocks-clean.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sitecore-content-sdk/nextjs': minor
+---
+
+[nextjs] Fix sitemap route validation to retun undefined for sitemap-index path.
```
illiakovalenko:
```suggestion
[nextjs] Fix sitemap route validation to return undefined for sitemap-index path.
```

illiakovalenko (approved):
👍

PR #520, opened 2026-06-17
[nextjs][create-content-sdk-app] Fix cross-origin editing cookies

glorynguyen:
## Description / Motivation

When the Content SDK Next.js App Router template is deployed to Vercel (or other serverless platforms), XM Cloud editing mode breaks completely:

1. Editing cookies (`__prerender_bypass`, `sc_mode`, etc.) use default `SameSite=Lax`, which browsers block inside the cross-origin XM Cloud editor iframe (`pages.sitecorecloud.io` → rendering host)
2. `draftMode().isEnabled` returns `false` in `page.tsx` because draft cookies don't propagate between Vercel serverless functions

This PR fixes both issues at two levels:
- **SDK level:** `createEditingRenderRouteHandlers()` now patches all editing cookies with `SameSite=None; Secure`
- **Template level:** Route handlers add a defense-in-depth cookie wrapper; page templates detect editing mode via URL `searchParams` as fallback when `draftMode()` fails

Editing on localhost remains unaffected.

## Testing Details

- [x] Unit Test Added

All 10 packages pass `yarn test-packages` (799 nextjs tests pass). `yarn lint-packages` passes with 0 errors. `yarn api-extractor:verify` passes with no public API surface changes. Manual verification: deploy to Vercel, open page in XM Cloud editor, confirm editing cookies show `SameSite=None; Secure` in DevTools, confirm page renders in editing mode.

## Types of changes

### DIFF
--- a/.changeset/fix-editing-cookies-samesite-vercel.md
+++ b/.changeset/fix-editing-cookies-samesite-vercel.md
@@ -1,0 +1,6 @@
+---
+'@sitecore-content-sdk/nextjs': patch
+'create-content-sdk-app': patch
+---
+
+Fix cross-origin editing cookies and draft mode detection on Vercel. Editing cookies now include SameSite=None; Secure for cross-origin iframe compatibility, and page templates fall back to URL searchParams for editing detection when draftMode() returns false on serverless platforms.
--- a/packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/page.tsx
+++ b/packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/page.tsx
@@ -17,7 +17,7 @@ import { NextIntlClientProvider } from 'next-intl';
 import { setRequestLocale } from 'next-intl/server';
 
 type PageProps = {
-  params: Promise<{ site: string; locale: string; path?: string[]; [key: string]: string | string[] | undefined }>;
+  params: Promise<{ site: string; locale: string; path?: string[];[key: string]: string | string[] | undefined }>;
   searchParams: Promise<{ [key: string]: string | string[] | undefined }>;
 };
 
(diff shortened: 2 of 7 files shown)


sc-nikolaoslazaridis:
Hello @glorynguyen , thank you for your contribution. As we did an internal investigation the issue can be solved by setting the correct environment variables. It was identified that the `resolveServerUrl` function in https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/editing/utils.ts#L324 is not picking up the correct values and uses `http` instead of `https` causing issues with the cookies. 

Please verify the environment variables like `VERCEL` or `SITECORE_INTERNAL_EDITING_HOST_URL` are set during your tests and let us know the results.

glorynguyen:
> Hello @glorynguyen , thank you for your contribution. As we did an internal investigation the issue can be solved by setting the correct environment variables. It was identified that the `resolveServerUrl` function in https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/editing/utils.ts#L324 is not picking up the correct values and uses `http` instead of `https` causing issues with the cookies.
> 
> Please verify the environment variables like `VERCEL` or `SITECORE_INTERNAL_EDITING_HOST_URL` are set during your tests and let us know the results.

Thank you for the investigation @sc-nikolaoslazaridis! I verified that `VERCEL` is set in my deployment (Vercel injects it automatically), and `resolveServerUrl` correctly resolves to `https://`. The protocol of the internal request isn't the root cause here.

The issue my PR addresses is specifically the **response `Set-Cookie` headers sent back to the browser**. In the GET handler at [line ~228](https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/route-handler/editing-render-route-handler.ts#L228):

```ts
const convertedCookies = cookieStore.getAll().map((c: NextCookie) => `${c.name}=${c.value}`);
```

This strips all cookie attributes (`SameSite`, `Secure`, `Path`, `HttpOnly`). When these bare `name=value` strings are used in the response `Set-Cookie` header ([line ~263](https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/route-handler/editing-render-route-handler.ts#L263)):

```ts
responseHeaders['Set-Cookie'] = filteredCookies?.join('; ') || '';
```

the browser applies the default `SameSite=Lax`, which blocks the cookies inside the cross-origin XM Cloud editor iframe (`pages.sitecorecloud.io` → Vercel rendering host).

The `sameSite: 'none'` in `cookieStore.set()` (lines 214–226) only affects the internal cookie store used for the server-side fetch — those attributes don't carry through to the manually-constructed response header.

There is also a secondary correctness issue: multiple cookies are joined with `'; '`, which makes them look like attributes of a single cookie rather than separate cookies. My PR changes the separator to `', '` per the HTTP spec.

**To reproduce:** inspect the `Set-Cookie` response header from `/api/editing/render` in DevTools when loading a page in the XM Cloud editor — you'll see bare `name=value` cookies without `SameSite` or `Secure` attributes.

sc-nikolaoslazaridis:
@glorynguyen Do you mind giving  a bit more information about your setup? I did try both the latest v2 and v1 with Vercel and couldn't reproduce the issue. What you are saying about the cookies makes total sense but i need to be able to reproduce the bug in order to validate that the fix does what it needs to do.

sc-nikolaoslazaridis:
@glorynguyen Thank you again for your contribution. Unfortunately i have to close this PR as we could not reproduce the issue. 
I would suggest you create an Issue if the problem persists with all the needed information like steps to reproduce, environment, app setup, ContentSDK setup (version, configurations etc) and any other important information you might have. 
I understand your POV on the cookies and i would like you to check what is happening with the actual request we perform in order to get the html. We have a custom implementation of the draft mode that doesn't require any cookies to be set in the browser because we do an internal call to the page route instead of returning a Redirect as the Next.js documentation describes. This is the main technical reason we are not merging directly your PR.

PR #524, opened 2026-06-18
[nextjs][content] Support time-based preview in editing render endpoint

sc-shakyawijerathne:
- [x] Changelog updated

## Description / Motivation

This PR adds support for calendar-based preview in the JSS editing render endpoint. The editing middleware now accepts an optional `sc_previewTime` query parameter and forwards it as a header to Edge Preview GraphQL.

This enables XM Cloud Pages users to preview content at specific future dates, allowing validation of scheduled content changes before they go live.

## Testing Details

- [x] Unit Test Added

**Unit Tests:**
- Test that `sc_previewTime` query param is forwarded as header
- Test backward compatibility when `sc_previewTime` is absent

**Manual Testing:**
- Verified local editing render endpoint accepts `sc_previewTime` parameter
- Confirmed header is correctly forwarded to GraphQL requests

## Types of changes


packages/nextjs/src/editing/editing-render-middleware.ts
```diff
@@ -205,6 +205,11 @@ export class EditingRenderMiddleware extends RenderMiddlewareBase {
         ...getHeadersForPropagation(headers),
         [EDITING_PARAMS_HEADER]: JSON.stringify(previewData),
       };
+
+      if (query.sc_previewTime) {
```
illiakovalenko: We have [mapEditingParams](https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/editing/utils.ts#L68) method, you can map this parameter there

packages/content/src/editing/models.ts
```diff
@@ -17,6 +17,7 @@ export interface EditingRenderQueryParams {
   sc_layoutKind?: LayoutKind;
   sc_variant?: string;
   sc_version?: string;
+  sc_previewTime?: string;
```
illiakovalenko, line 20: The new parameter needs to be passed to [EditingService.fetchEditingData](https://github.com/Sitecore/content-sdk/blob/dev/packages/content/src/editing/editing-service.ts#L84), so that it is included in the layout request.

packages/nextjs/src/editing/editing-render-middleware.ts
illiakovalenko: Also, keep in mind that we have [editing-render-route-handler](https://github.com/Sitecore/content-sdk/blob/dev/packages/nextjs/src/route-handler/editing-render-route-handler.ts) - App Router. You, probably, shouldn't need to make any additional changes there, as updating the `mapEditingParams` method should apply this automatically. Probably only unit tests

illiakovalenko (changes requested):
Looks good 👍
Please, see a couple of comments below

packages/content/src/editing/models.ts
illiakovalenko: It should be set as a `sc_previewTime` header

packages/nextjs/src/editing/utils.ts
```diff
@@ -63,6 +63,7 @@ export const mapEditingParams = (query: {
         dataSourceId: query.dataSourceId,
         version: query.sc_version,
         generation: query.generation,
+        ...(query.sc_previewTime && { previewTime: query.sc_previewTime }),
```
illiakovalenko: In design studio we don't expect to pass it, so let's remove it. We need to support it only for Sitecore Pages

packages/content/src/client/sitecore-client.ts
```diff
@@ -476,6 +476,7 @@ export class SitecoreClient implements BaseSitecoreClient {
         mode,
         site,
         variantId,
+        ...(previewTime && { previewTime }),
```
illiakovalenko: Can we simplify it a bit?
```suggestion
        previewTime,
```

packages/content/src/editing/editing-service.ts
illiakovalenko, line 1: Let's add unit tests to cover new parameter

packages/nextjs/src/editing/utils.ts
illiakovalenko, line 1: Let's update unit tests

illiakovalenko (changes requested):

packages/nextjs/src/editing/utils.ts
```diff
@@ -72,6 +72,7 @@ export const mapEditingParams = (query: {
         version: query.sc_version,
         mode: query.mode,
         layoutKind: query.sc_layoutKind,
+        ...(query.sc_previewTime && { previewTime: query.sc_previewTime }),
```
illiakovalenko:
```suggestion
        previewTime: query.sc_previewTime,
```

illiakovalenko (approved):
Looks great! 👍

PR #532, opened 2026-06-24
[create-content-sdk-app] Build Failed on new project in Sitecore AI

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

Fixes App Router build failures in create-content-sdk-app when Sitecore Edge is unavailable at build time.

- Wrap Edge fetches in try/catch in page, not-found, generateStaticParams, and generateMetadata routes
- Return [] from generateStaticParams when generateStaticPaths is false instead of forcing a default prerender (`nextjs-app-router`)
- Return null from cached page helpers when Edge fetch fails
- Fall back to notFound() or default metadata instead of failing the build

## Testing Details

- [ ] Unit Test Added

## Types of changes


illiakovalenko (changes requested):
@MenKNas @sc-naveenhedallaarachchi @sc-shakyawijerathne Wrapping the calls in a try-catch block is a quick workaround, but it does not address the underlying issue or provide a proper long-term solution.
The root cause of this issue is the default site configuration defined [here](https://github.com/Sitecore/content-sdk/blob/dev/packages/create-content-sdk-app/src/templates/nextjs-app-router/src/app/%5Bsite%5D/%5Blocale%5D/%5B%5B...path%5D%5D/page.tsx#L67-L74). Adding a default site is not required, despite what the TSDoc currently states. We should avoid having any sites configured by default, since it doesn't exist in Sitecore AI.

packages/content/src/tools/generateSites.ts
```diff
@@ -65,14 +65,6 @@ export const generateSites = ({ destinationPath }: GenerateSitesConfig = {}): ((
...
     }
 
-    // Add default site to the list
-    const defaultSite: SiteInfo = {
-      name: scConfig.defaultSite,
-      hostName: '*',
-      language: scConfig.defaultLanguage,
-    };
-    sites.unshift(defaultSite);
-
```
illiakovalenko, line 75: This one should be present, since it allows you to run app in normal mode. It doesn't affect editing/preview/deployment to Sitecore AI, since in generateStaticParams you return empty array if `generateStaticPaths` flag disabled. (which is disabled when deploying to Sitecore AI)

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/sitecore.config.ts.example
```diff
@@ -21,7 +21,7 @@ export default defineConfig({
       apiHost: process.env.NEXT_PUBLIC_SITECORE_API_HOST || '',
     },
   },
-  defaultSite: process.env.NEXT_PUBLIC_DEFAULT_SITE_NAME || 'default',
+  defaultSite: process.env.NEXT_PUBLIC_DEFAULT_SITE_NAME,
```
illiakovalenko: I think here it's ok to keep it as it was.
Our main goal to don't provide any entries in `generateStaticParams` when application is used as editing host. If you run in normal mode it's totally fine to define a default site via env var or exact value since customers know about their sites

.changeset/fix-app-router-ssg-edge-build-failures.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sitecore-content-sdk/content': patch
+'create-content-sdk-app': patch
+---
+
+Stop synthesizing a default site in `generateSites` and return empty app-router static params when `generateStaticPaths` is false (editing hosts).
```
illiakovalenko: `generateStaticPaths` can be used in prod / editing host scenarios
```suggestion
Return empty app-router static params when `generateStaticPaths` is false.
```

illiakovalenko (changes requested):

illiakovalenko (approved):
👍

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx
```diff
@@ -2,16 +2,33 @@ import Link from 'next/link';
...
 export default async function NotFound() {
   const { site, locale } = getCachedPageParams();
-  const resolvedSite = site || scConfig.defaultSite;
+  const resolvedSite = (site || scConfig.defaultSite)?.trim();
   const resolvedLocale = locale || scConfig.defaultLanguage;
 
+  if (
+    !resolvedSite ||
+    isBuildValidationSite(resolvedSite) ||
+    isBuildValidationSite(resolvedLocale)
```
Comment (illiakovalenko) [inline packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: I'm not sure we it's correct to check for `isBuildValidationSite(resolvedLocale)`

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/lib/sitecore-build-validation.ts
```diff
@@ -0,0 +1,13 @@
+/**
+ * This is a placeholder `[site]` segment used only for Next.js Cache Components build-time validation
+ * when `generateStaticPaths` is false. Not a real Sitecore site — routes using this value
+ * skip Edge fetches and render the static not-found fallback so `next build` can succeed
+ * without a configured default site or content configured in Sitecore AI.
```
illiakovalenko, line 5: May be we can add a real example as a use case
```suggestion
 * without a configured default site or content configured in Sitecore AI.
 * One use case for this is running your application as an editing host when no sites are configured.
```

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/copilot-instructions.md
```diff
@@ -44,7 +44,8 @@ src/
 ## Library Usage
 
 ### @sitecore-content-sdk
-- Use `SitecoreClient` for content fetching (directly for preview/editing/static-params, indirectly via cache helpers for cached reads)
+- Use `SitecoreClient` for content fetching (directly for preview/editing and `getAppRouterStaticParams` when `scConfig.generateStaticPaths` is true; when false return `BUILD_VALIDATION_SITE` from `src/lib/sitecore-build-validation.ts` for Cache Components build validation; indirectly via cache helpers for cached reads)
+- **Build validation:** `_DEFAULT_` is build-only — skip Edge in page/metadata when `isBuildValidationSite(site)`; segment `not-found.tsx` skips Edge when site or locale is `_DEFAULT_`. Call `setCachedPageParams` before `notFound()` on real routes; read with `getCachedPageParams()` in segment not-found (do not use `headers()` there).
```
illiakovalenko: Can you please check if we have any other occurrences like this (it's mentioned at least in one another doc as well)? We're only checking the site using the token.

```suggestion
- **Build validation:** `_DEFAULT_` is build-only — skip Edge in page/metadata when `isBuildValidationSite(site)`; segment `not-found.tsx` skips Edge when site is `_DEFAULT_`. Call `setCachedPageParams` before `notFound()` on real routes; read with `getCachedPageParams()` in segment not-found (do not use `headers()` there).
```

illiakovalenko (approved):
Looks good 👍
Approved, left a couple of minor comments

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx
Comment (MenKNas) [reply packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/app/[site]/[locale]/[[...path]]/not-found.tsx]: That's a leftover from testing. Checking for site is enough, so this can be removed.

PR #537, opened 2026-06-25
[nextjs] Skip malformed redirect regex rules instead of failing entire redirect chain

sc-shakyawijerathne:
- [x] Changelog updated

## Description / Motivation

Fixes an issue where a single malformed regex in a Sitecore Redirect Map could prevent all redirects in that set from working.

When a redirect pattern was invalid (e.g. an unmatched parenthesis), regex compilation threw an error and aborted the entire redirect evaluation. Valid rules in the same Redirect Map were never reached.

This change safely compiles each redirect pattern, logs a warning for invalid rules, skips them, and continues evaluating the remaining rules.

## Testing Details

- [x] Unit Test Added

Added regression tests in `redirects-proxy.test.ts` covering:
- Malformed regex before a valid static redirect rule
- Malformed regex before a valid capture-group redirect rule

## Types of changes


packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -23,7 +23,20 @@ import { FailedProxyExecution, ProxiesContext, SuccessfulProxyExecution } from '
 const REGEXP_CONTEXT_SITE_LANG = new RegExp(/\$siteLang/, 'i');
 const REGEXP_ABSOLUTE_URL = new RegExp('^(?:[a-z]+:)?//', 'i');
 
-type RedirectResult = RedirectInfo & { matchedQueryString?: string; matchedPath?: string };
+/**
+ * Escape a string for safe use inside a RegExp source (e.g. locale segment).
+ * @param {string} string - The string to escape
+ * @returns {string} The escaped string
+ */
+function escapeRegExp(string: string): string {
```
illiakovalenko: We already have the same function implemeted in `packages\content\src\layout\rewrite-edge-host.ts`. Could you, please, have a look and share it instead?

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -437,6 +454,7 @@ export class RedirectsProxy extends ProxyBase {
           // Save the matched query string (if found) into the redirect object
           redirect.matchedQueryString = matchedQueryString || '';
           redirect.matchedPath = matchedPath || matchedPathWithQuery || '';
+          redirect.compiledPattern = regex;
```
illiakovalenko: Do we really need to set `compiledPattern` property? Can we instead just filter reg expressions within `matchFromRedirectMapRedirect`, so we don't need to additionally validate regexp in `handle` method?

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -267,9 +280,10 @@ export class RedirectsProxy extends ProxyBase {
       // Apply regex replacements to the target URL if the pattern is a regex
       const sourcePath = existsRedirect.matchedPath || reqUrl.pathname;
       const pathForCaptureMatch = sourcePath.replace(/\/*$/gi, '') || '/';
-      const matched = pathForCaptureMatch.match(
-        this.getRedirectPatternRegex(existsRedirect.pattern)
-      );
+      const redirectRegex =
+        existsRedirect.compiledPattern ??
+        this.safeCompileRedirectPattern(existsRedirect.pattern);
+      const matched = redirectRegex ? pathForCaptureMatch.match(redirectRegex) : null;
```
illiakovalenko: See my related comment https://github.com/Sitecore/content-sdk/pull/537/changes#r3489794981

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -616,6 +637,26 @@ export class RedirectsProxy extends ProxyBase {
...
+   */
+  private safeCompileRedirectPattern(pattern: string): RegExp | null {
+    try {
+      return this.getRedirectPatternRegex(pattern);
+    } catch (error) {
+      const message = error instanceof Error ? error.message : String(error);
+      console.warn(
+        `[RedirectsProxy] Invalid redirect regex; skipping rule. pattern=${pattern} (${message})`
+      );
+      debug.redirects('invalid redirect regex; skipping rule: %s (%s)', pattern, message);
```
illiakovalenko: I think we can drop `debug.redirects` log, since you've already added `console.warn` that is always present

.changeset/fix-redirects-malformed-regex.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sitecore-content-sdk/nextjs': patch
+---
+
+[nextjs] Skip malformed redirect regex rules instead of failing the entire redirect chain.
```
illiakovalenko: We are under `nextjs` scope already, so it's not required to add a prefix in changesets
```suggestion
Skip malformed redirect regex rules instead of failing the entire redirect chain.
```

illiakovalenko (changes requested):
Looks good! 👍
Please, see a couple of suggestions below

illiakovalenko:
@sc-shakyawijerathne I added one change - moved `escapeRegExp` under `core/tools`, this is more appropriate place for shared utilities

illiakovalenko (approved):

PR #546, opened 2026-07-06
[Angular CSDK] Adjust sample to reduce warnings during build

art-alexeyenko:
- Remove budget suggestions
- List allowed CJS dependencies we cannot replace yet


illiakovalenko (approved):
👍

PR #547, opened 2026-07-07
[content][nextjs][react] disableCodeGeneration: true causes build error - loadImportMap has no no-op fallback when import map generation is skipped

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

Address the issue when the apps with `disableCodeGeneration: true` failed to build because `.sitecore/import-map` was missing and `loadImportMap` was required on `SitecoreProvider`.
- **Codegen (`content`)**: When code generation is disabled, write empty import map stub files so TypeScript/bundlers can resolve `.sitecore/import-map` modules.
- **Runtime (`react`, `nextjs`)**: Make `loadImportMap` / `loadServerImportMap` optional and default to an empty import map (`Promise.resolve({ default: [] })`) when omitted.

## Testing Details

- [x] Unit Test Added

Verify that an app with `disableCodeGeneration: true` builds correctly on the first run without requiring `loadImportMap` to be passed manually or import-maps to be created manually.

## Types of changes


packages/react/src/components/SitecoreProvider.tsx
sc-nikolaoslazaridis: I think we should put a note here that this is optional and a description of what will happen in the case of the default fallback just like you did in the type above.

packages/content/src/tools/codegen/import-map.ts
```diff
@@ -391,7 +417,10 @@ export const writeImportMap = (args: WriteImportMapArgsInternal) => {
     }
 
     if (scConfig.disableCodeGeneration) {
-      debug.common('Skipping import map generation. Code generation functionality is disabled.');
+      debug.common(
+        'Code generation is disabled. Writing empty import map stubs for bundler resolution.'
```
sc-nikolaoslazaridis: Lets try to change the word `stub` here to avoid and future questions

packages/content/src/tools/codegen/import-map.ts
```diff
@@ -376,6 +376,32 @@ const prepImportMaps = async (paths: string[], separateMaps?: boolean): Promise<
   ];
 };
 
+const writeStubImportMaps = (
+  args: WriteImportMapArgsInternal,
+  defaultTemplate: (indexedImportMap: Map<string, ModuleExports>) => string,
+  clientTemplate: (indexedImportMap: Map<string, ModuleExports>) => string
+) => {
+  const sitecoreDir = path.join(process.cwd(), '.sitecore');
```
sc-nikolaoslazaridis, line 385: Do we have a util function to check/create the `.sitecore` folder?

sc-nikolaoslazaridis (commented):
Good job! I left some minor comments for you. A thing i would consider is to create a reusable util function for the noop case since we need it in multiple places

packages/content/src/tools/codegen/import-map.ts
sc-naveenhedallaarachchi: updated to check the directory through `ensurePathExists`

packages/content/src/tools/codegen/import-map.ts
sc-naveenhedallaarachchi: updated

packages/react/src/components/SitecoreProvider.tsx
sc-naveenhedallaarachchi: done

sc-nikolaoslazaridis (approved):

PR #555, opened 2026-07-13
[content] Restore Hidden Rendering in Pages edit mode

MenKNas:
- [ ] Changelog updated

## Description / Motivation
Fixes the hidden component issue in Pages edit mode, components hidden by personalization were removed from the layout instead of rendering as a Hidden Rendering placeholder with editing chrome.

Problem
A regression from server-side `sc_variant handling` in `getPreview()`: the editing API resolves personalization before the SDK sees the layout, so hidden components are dropped entirely. Pages then has nothing to attach actions to.

Solution
For edit mode only, we are now fetching preview layout without the `sc_variant` header (new `sendVariantHeader` option on `fetchEditingData`), then apply `personalizeLayout()` client-side so hidden components become Hidden Rendering. Preview mode is unchanged and still uses server-side variant resolution.
## Testing Details
- [x] Unit Test Added
- [x] Manual Test/Other => Verified fix manually in Pages (default + variant hide)

## Types of changes


illiakovalenko (changes requested):
It was explicitly requested that the `sc_variant` header be sent in both Preview and Edit modes. The `sc_variant` header helps optimize performance in Edit mode by reducing the number of experiences that need to be transferred.
We **should not** work around this issue by omitting the `sc_variant` header in Edit mode. Based on your explanation and my understanding, this appears to be a backend issue, as hidden variants are being ignored entirely. Could you please discuss this with the devex team and report it internally?

MenKNas:
@illiakovalenko After our discussion, it seems like this is not a workaround we should accept, hence the solution is moved on the backend side. I will close this PR.

PR #562, opened 2026-07-21
[Non-breaking] Refactor redirects proxy implementation and move reusable logic out

art-alexeyenko:

## Description / Motivation
Redirects proxy service went through non-breaking slimming down procedure. Whatever could be moved out and reused in other frameworks, was moved.
This PR also does a small half-fix, ensuring locales from URL are compared while ignoring case (only some paths were doing this previously).
This serves as prep to fixing some reported issues with redirects and implementing the SXA redirects for Angular.

*NOTE*: merge after https://github.com/Sitecore/content-sdk/pull/563

## Testing Details
- [x] Unit Test Added

## Types of changes
- [x] Chore?


.changeset/cold-shoes-do.md
```diff
@@ -0,0 +1,6 @@
+---
+'@sitecore-content-sdk/content': patch
+'@sitecore-content-sdk/nextjs': patch
+---
+
+Refactor redirects logic to be reusable across frameworks
```
illiakovalenko: Can you please mention also some other issues that you fixed in scope of this work, if there is any? At least I noticed that locale-prefix detection is now case-insensitive

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -186,109 +167,63 @@ export class RedirectsProxy extends ProxyBase {
...
-            incomingQS,
-            new URLSearchParams(targetQS || '')
-          );
-          targetUrl = `${targetMainUrl}?${mergedQueryString}`;
-        }
-
-        return this.dispatchRedirect(targetUrl, existsRedirect.redirectType, req, res, true);
+      const reqUrl = this.normalizeUrl(req.nextUrl.clone());
+      const incomingPathData: ProcessedPath = {
+        nonLocalePath: reqUrl.pathname,
```
illiakovalenko: As far as I can see `incomingPathData.nonLocalePath` is unused. Probably it was added just in order to satisfy type check, but actually `reqUrl.pathname` includes locale, so it's not right. Better to adjust the type here and remove `nonLocalePath`

packages/content/src/site/redirect-utils.ts
```diff
@@ -0,0 +1,275 @@
...
+};
+
+/**
+ * Splits a URL path into its (optional) leading locale segment, the remaining
+ * locale-less path, and the query string. Locales are compared case-insensitively.
+ * @param {string[]} configuredLocales configured site locales
+ * @param {string} urlPath path (optionally including a `?query`)
+ * @returns {ProcessedPath} broken-down path parts
+ */
+export const breakDownPath = (configuredLocales: string[], urlPath: string): ProcessedPath => {
```
illiakovalenko, line 41: `breakDownPath` has an edge case for the root path `/` after trailing-slash stripping, nonLocalePath probably resolves to '' rather than '/'. Not sure if it's expected by design. Also it's untested. A quick unit test for `breakDownPath(LOCALES, '/')` would pin this down.

packages/content/src/site/redirect-utils.ts
illiakovalenko, line 1: For new API can you please add the neccessary tags: `@public`, `@internal` etc.

illiakovalenko (changes requested):
Overall, this looks great! I like these changes 👍
Please take a look at my comments and suggestions below.

packages/content/src/site/redirect-utils.ts
art-alexeyenko: marked all as internal - they methods used before were basically the same accessibility

.changeset/cold-shoes-do.md
art-alexeyenko: Done. This is really a half-fix. Some paths were comparing locales case-insensitively, some not. We're now unifying the approach

packages/content/src/site/redirect-utils.ts
art-alexeyenko: Good point - and we also need to preserve trailing slashes if present. This has been added.

.changeset/cold-shoes-do.md
illiakovalenko: @art-alexeyenko Do you mention some changes / improvements in the changeset?

packages/content/src/site/redirect-utils.ts
```diff
@@ -0,0 +1,280 @@
...
+  const localePath = `/${requestLocale.toLowerCase()}${nonLocalePath}`;
+  const incomingQS = incomingPathData.queryString || '';
+  // regex rules are matched against the raw candidates (trailing slash preserved so slash-sensitive
+  // patterns still match); static rules ignore trailing slashes, so they use trimmed candidates
+  const pathCandidates = [nonLocalePath, localePath];
+  const staticCandidates = pathCandidates.map((path) => path.replace(/\/+$/g, '') || '/');
+
+  return redirectMapRedirects.find((redirect: RedirectResult) => {
+    // process static URL (non-regex) rules
+    if (isRegexOrUrl(redirect.pattern) === 'url') {
```
illiakovalenko, line 96: I checked deeper and prompted AI as well, can you have a look?
In `matchFromRedirectMapRedirect` (r78-131), static (isRegexOrUrl === 'url') rules match against `staticCandidates`, which strip trailing slashes (line 92). But regex rules (lines 112-124) only test the raw, slash-preserved pathCandidates - never the trimmed variants.

Previously the old implementation tested both a slash-stripped and non-stripped candidate for regex rules, so an anchored pattern like ^/old-page$ matched both /old-page and /old-page/. Now, a request to /old-page/ will fail to match that same regex rule, while an equivalent static rule (/old-page) still matches fine via `staticCandidates`. That's an inconsistency between the two rule types, and a behavior change - worth double-checking. No test currently exercises this (redirect-utils.test.tslash case).

packages/nextjs/src/proxy/redirects-proxy.ts
illiakovalenko, line 280: Is it possible to reuse `breakDownPath` here? since it's used for normalization as well

illiakovalenko (changes requested):

packages/nextjs/src/proxy/redirects-proxy.ts
art-alexeyenko: breakDownPath will return result with trailing slash, if it's present in input URL, so it doesn't fully normalize the path
This is needed to have a correct representation of the input - all the coditionals and special processing cases should fall to other logic

.changeset/cold-shoes-do.md
art-alexeyenko: Added a separate changeset

packages/content/src/site/redirect-utils.ts
art-alexeyenko: Thanks, that's a regression indeed.
Added a unit test and a fix

illiakovalenko (approved):
Well done! 👍

PR #563, opened 2026-07-21
[nextjs][create-content-sdk-app] RedirectsProxy doesn't work if LocaleProxy is removed

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

- Fix RedirectsProxy failing on App Router when LocaleProxy is removed: locale handling previously relied on the optional `x-sc-locale` header, which caused Pages Router `url.locale` assignment and 404s.
- Add Next.js-specific `redirects.localeInPath` (`true` | `false` | `null`) so redirect locale path prefixes are explicit:
  - `true` — prepend `/[locale]` (App Router with `[locale]` segment)
  - `false` — do not prepend (App Router without `[locale]`)
  - `null` (default) — keep `x-sc-locale` fallback (non-breaking)
- Set `localeInPath: true` in App Router templates (`nextjs-app-router`, `nextjs-app-router-cache-components`).

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -243,8 +243,11 @@ export class RedirectsProxy extends ProxyBase {
         }
 
         if (!isAppRouterRequest) {
```
illiakovalenko: We already have a check for the `isAppRouter` request, but we're currently entering this `if` block even when running in an App Router configuration.

My concern is that we shouldn't need to rely on workarounds here. Instead, let's address the root cause and find a reliable way to correctly detect the App Router configuration.

illiakovalenko (changes requested):

packages/nextjs/src/proxy/redirects-proxy.ts
sc-naveenhedallaarachchi: Updated the logic to detect App Router configurations

packages/nextjs/src/proxy/app-router-multisite-proxy.ts
```diff
@@ -40,4 +41,13 @@ export class AppRouterMultisiteProxy extends MultisiteProxy {
     const path = pathname.startsWith('/') ? pathname : '/' + pathname;
     return `/${siteName}${path}`;
   }
+
+  /**
+   * Marks the response so downstream proxies (e.g. RedirectsProxy) detect App Router
+   * without relying on LocaleProxy / x-sc-locale.
+   * @param {NextResponse} res response
+   */
+  protected withAppRouterSignal(res: NextResponse): void {
```
illiakovalenko: The Multisite feature is completely optional, and customers can remove the `MultisiteProxy`, so we can't rely on it being present. Is there another identifier we could use instead?

You may find this document helpful:
https://doc.sitecore.com/sai/en/developers/content-sdk/20/DEFAULT/disable-multisite-and-locale-proxies.html#use-a-single-sitesingle-locale-setup

One option would be to rely on `nextUrl.defaultLocale`. However, it's not guaranteed to be available, even in the Pages Router, if the customer hasn't configured it.

Also, even if we rely on `nextUrl.defaultLocale`, we still have the `else` branch where we prepend the locale to the path for the App Router. But what happens if the customer has removed the `LocaleProxy` and doesn't have a `[locale]` route segment? In that case, we can't simply prepend the locale to the path because it would likely result in a 404.

These edge cases need to be verified. At a minimum, we may need to introduce an explicit configuration flag (for example, `localeInPath`) for the redirects proxy so it knows whether locale-prefixed routes are actually being used.

illiakovalenko (changes requested):

packages/nextjs/src/proxy/app-router-multisite-proxy.ts
sc-naveenhedallaarachchi: Thanks Illia — agreed. 

Based on your suggestion, I added an explicit redirects.localeInPath flag (true | false | null) so redirect locale shaping doesn’t depend on optional proxies:

- true - App Router with a [locale] segment; prepend locale to the path
- false - App Router without [locale] (or no path locale); do not prepend
- null (default) - keep the existing x-sc-locale header fallback for compatibility

App Router templates set localeInPath: true by default. Setups that remove LocaleProxy and the [locale] segment should use localeInPath: false.

This is **non-breaking**: default remains null, so existing App Router apps with LocaleProxy and Pages Router apps with Next.js i18n keep current behavior without config changes.

Please review and let me know if any changes are needed.

packages/nextjs/src/proxy/app-router-multisite-proxy.ts
```diff
@@ -40,4 +40,4 @@ export class AppRouterMultisiteProxy extends MultisiteProxy {
     const path = pathname.startsWith('/') ? pathname : '/' + pathname;
     return `/${siteName}${path}`;
   }
-}
+}
```
illiakovalenko:
```suggestion
}

```

packages/nextjs/src/proxy/app-router-multisite-proxy.test.ts
```diff
@@ -43,4 +43,4 @@ describe('AppRouterMultisiteProxy', () => {
       expect(result).to.equal('/mysite/');
     });
   });
-});
+});
```
illiakovalenko:
```suggestion
});

```

packages/create-content-sdk-app/src/templates/nextjs-app-router/sitecore.config.ts
```diff
@@ -4,4 +4,9 @@ import { defineConfig } from '@sitecore-content-sdk/nextjs/config';
  * See the documentation for `defineConfig`:
  * https://doc.sitecore.com/xmc/en/developers/content-sdk/the-sitecore-configuration-file.html
  */
-export default defineConfig({});
+export default defineConfig({
+  redirects: {
+    // App Router uses a [locale] segment; keep locale in redirect paths even if LocaleProxy is removed
```
illiakovalenko: Suggesting a small improvement in description
```suggestion
    // Routes live under a [locale] segment (src/app/[site]/[locale]),
    // so redirect targets must include the locale prefix to resolve.
```

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/sitecore.config.ts
```diff
@@ -14,4 +14,8 @@ export default defineConfig({
       enabled: false,
     },
   },
+  redirects: {
+    // App Router uses a [locale] segment; keep locale in redirect paths even if LocaleProxy is removed
```
illiakovalenko: Suggesting a small improvement in description
```suggestion
    // Routes live under a [locale] segment (src/app/[site]/[locale]),
    // so redirect targets must include the locale prefix to resolve.
```

packages/content/src/config/define-config.ts
```diff
@@ -51,6 +51,8 @@ export const buildFallbackConfig = (env: { [key: string]: string | undefined }):
     redirects: {
       enabled: env.NODE_ENV !== 'development',
       locales: ['en'],
+      // null = unset: RedirectsProxy falls back to LocaleProxy header detection
+      localeInPath: null,
```
illiakovalenko: `redirects.localeInPath` looks Next.js-specific to me - should it live in packages/nextjs instead of the shared packages/content config schema?
Mainly we face this problem because of different dual nextjs architectures we need to support and some AppRouter specifics

.changeset/empty-years-spend.md
```diff
@@ -0,0 +1,7 @@
+---
+'@sitecore-content-sdk/content': patch
+'@sitecore-content-sdk/nextjs': patch
+'create-content-sdk-app': patch
+---
+
+RedirectsProxy no longer depends on LocaleProxy's `x-sc-locale` header for App Router locale shaping. Add `redirects.localeInPath` (`true` | `false` | `null`) and set `localeInPath: true` in App Router templates.
```
illiakovalenko: RedirectsProxy still depends on x-sc-locale header by default, it's not a breaking change.
Can we rephrase it a bit and explain the reason of this change? Also default (backwards compatible) value

illiakovalenko (changes requested):

packages/content/src/config/define-config.ts
illiakovalenko: If it will be neccessary (for example for Angular) we can make it shared move it back to `content`

packages/create-content-sdk-app/src/templates/nextjs-app-router/sitecore.config.ts
sc-naveenhedallaarachchi: Done

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/sitecore.config.ts
sc-naveenhedallaarachchi: Done

packages/content/src/config/define-config.ts
sc-naveenhedallaarachchi: moved redirects.localeInPath to Next.js config

.changeset/empty-years-spend.md
sc-naveenhedallaarachchi: Updated

illiakovalenko (approved):
Everything looks good from my side 👍

PR #564, opened 2026-07-22
[Atoms][Experimental] Compile runtime Document classNames into CSS

MenKNas:
- [ ] Changelog updated

## Description / Motivation
This PR implements runtime CSS compilation for Atoms Document classNames based on Nick’s POC ([#558](https://github.com/Sitecore/content-sdk/pull/558) / [5b48f34](https://github.com/Sitecore/content-sdk/commit/5b48f34392039e9c0ed624bf60c8b361439c4da9)).

Tailwind can’t see classes that only exist in MMS Document JSON at build time, so styles never ship. This compiles those classes at runtime and injects the CSS for Design Library editing and production RSC rendering.

Defaults to Tailwind, with a pluggable compiler registry for custom setups.

What was added extra beyond the POC:
- Fixed Tailwind `build()` candidate accumulation so only current Document classes are in the injected CSS (not stale ones from earlier updates)
- Wired both App Router starters (instrumentation, compileCssAction, serverExternalPackages)
- Added caching, clearer missing-compiler/CSS-file DX, and unit coverage for the accumulation fix

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/nextjs/tsconfig.json
```diff
@@ -28,6 +28,8 @@
     "search.d.ts",
     "client.d.ts",
     "atoms.d.ts",
+    "instrumentation.d.ts",
```
art-alexeyenko, line 31: can we put this logic under codegen submodule instead?

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/Providers.tsx
```diff
@@ -15,7 +16,12 @@ export default function Providers({ children, page }: { children: React.ReactNod
       componentMap={components}
       page={page}
       loadImportMap={() => import('.sitecore/import-map.client')}
-      atomsConfig={{ catalog, registry, navigate: router.push }}
+      atomsConfig={{
+        catalog,
+        registry,
+        navigate: router.push,
+        compileCssAction: compileCssForDocumentAction,
```
art-alexeyenko, line 23: do we *need* to pass the default implementation here? instead of setting in in nextjs package?

packages/create-content-sdk-app/src/templates/nextjs-app-router/instrumentation.ts
```diff
@@ -0,0 +1,12 @@
+/**
+ * Next.js instrumentation hook — runs once when the server process starts.
+ *
+ * Splits on runtime: Node.js-specific startup code (Tailwind compiler registration)
+ * lives in `instrumentation-node.ts` so that Turbopack does not attempt to trace
+ * Node.js built-ins (`path`, `fs`, `@tailwindcss/node`) for the Edge runtime.
+ */
+export async function register() {
```
art-alexeyenko, line 8: do we need this for pages router as well?

packages/react/src/components/DesignLibrary/StudioComponentServerWrapper.tsx
```diff
@@ -46,7 +53,33 @@ export const StudioComponentServerWrapper = async (props: StudioComponentServerW
...
+    if (compiler) {
+      try {
+        css = await compiler(classes);
+      } catch (err) {
+        if (process.env.NODE_ENV !== 'production') {
+          console.warn('StudioComponentServerWrapper: CSS compile failed', err);
+        }
+      }
+    } else if (process.env.NODE_ENV !== 'production') {
+      console.warn(
```
art-alexeyenko: let's replace this and the console.warn above with debug logs. They can be accessed on both prod and non-prod instances

packages/react/src/atoms/extract-document-classes.ts
```diff
@@ -0,0 +1,30 @@
...
+ *
+ * Walks the flat `doc.elements` map and collects every `className` string found
+ * in element props, splitting on whitespace and deduplicating. The resulting array
+ * can be passed to a CSS compiler (e.g. via `setAtomsCssCompiler`) to generate
+ * the utility rules needed for the Document's components.
+ * @param {Document} doc - The Document to extract classes from.
+ * @returns {string[]} Deduplicated array of class tokens.
+ * @public
+ */
+export function extractDocumentClasses(doc: Document): string[] {
```
art-alexeyenko, line 14: what do you think of making this framework-agnostic and putting in i.e. content for future use?

packages/nextjs/src/instrumentation/tailwind-compiler.ts
```diff
@@ -0,0 +1,99 @@
...
+  cssFilePath: string
+): (classes: string[]) => Promise<string> {
+  const handle = loadTailwindCssFile(cssFilePath);
+  const cache = new Map<string, string>();
+
+  return async (classes: string[]) => {
+    if (!classes.length) return '';
+
+    const key = atomsCssCacheKey(classes);
+    const cached = cache.get(key);
```
art-alexeyenko, line 89: what's your intention with the cache here?
I'm asking cause this cache will not do well in serverless deploys, it's mainly useful for onPrem or single-instance ones.
If atoms-powered app will only be used in those cases, this is fine, otherwise we might need to add some cache options for serverless (or accept the performance hit)

packages/nextjs/src/instrumentation/css-cache.test.ts
```diff
@@ -0,0 +1,30 @@
+/* eslint-disable no-unused-expressions */
+import { expect } from 'chai';
+import {
+  ATOMS_CSS_CACHE_MAX_ENTRIES,
+  atomsCssCacheKey,
+  setAtomsCssCacheEntry,
+} from './constants';
+
+describe('atoms CSS cache helpers', () => {
```
art-alexeyenko, line 9: if cache is here to stay, it's best to move its logic to smth like css-cache.ts

packages/nextjs/src/instrumentation/css-layers.ts
```diff
@@ -0,0 +1,35 @@
...
+ * Extracts the CSS layers needed for document-scoped injection.
+ * Includes `@layer theme` (CSS custom properties used by utility classes) and
+ * `@layer utilities` (the actual utility rules). Skips `@layer base` (CSS resets)
+ * since those are already present in the app's main bundle.
+ * @param {string} css - Full compiled Tailwind CSS string.
+ * @returns {string} Theme and utilities layers joined, or empty string if neither found.
+ * @internal
+ */
+export function extractDocumentCssLayers(css: string): string {
+  return [extractLayer(css, 'theme'), extractLayer(css, 'utilities')].filter(Boolean).join('\n');
```
art-alexeyenko, line 34: Is there some convention for DesignLibrary to use these two non-base layers only? Are more layer names possible?

art-alexeyenko (changes requested):
I do like how we just keep finding barely-public APIs for these kinds of things. Let's refine some more and bring it to a close.

packages/nextjs/tsconfig.json
MenKNas: This is not related to codegen. Codegen is about import maps/generated component tooling, this is related to registering the CSS compiler at Node startup so we can read styling classes at runtime (and it's atoms specific), meaning it should live under `instrumentation`

packages/react/src/atoms/extract-document-classes.ts
MenKNas: I totally agree and it's a future intention to move everything into `content` when we support other frameworks in Atoms, but for now it was already decided to keep everything under `react` package since we only support React components. Since this is atom specific, we are good to keep it like this for the beta version.

packages/nextjs/src/instrumentation/tailwind-compiler.ts
MenKNas: The cache is our best-shot for in-process memoization and we should take advantage of the caching mechanism as much as possible. For the cases where cache cannot be used, the performance cost is acceptable. Of course this is going to be related to the number of atomic components an application is using. For larger amount of components we can't really predict the cost in production environments although the impact should still be small. This feature is experimental in nature after all, so some things will need to be tested in action. I think it's preferable to have this as a starting point and see if we can refine or provide more case-specific solutions later (if we have to).

packages/nextjs/src/instrumentation/css-layers.ts
MenKNas: We only need utility rules that actually appear in our Document classNames (plus `theme` which includes the CSS variables those utilities depend upon).

packages/create-content-sdk-app/src/templates/nextjs-app-router/src/Providers.tsx
MenKNas: Yes, we need to pass it in the app (for now) since `compileCssForDocumentAction` is a Next.js Server Action and `SitecoreProvider` can’t default to that action itself. 

I am not against creating some Nextjs specific wrapper in the future to "hide" this little wiring, but I think it shouldn't be a blocker for now as the main target is to see if this works and addresses our styling issues. There will be a lot of room for refinements after we finalize the beta phase.

packages/react/src/components/DesignLibrary/StudioComponentServerWrapper.tsx
MenKNas: Valid point, I will update this to use debug logs.

packages/nextjs/src/instrumentation/css-cache.test.ts
MenKNas: Good point, I will move it into `css-cache.ts` (I like the naming)

packages/create-content-sdk-app/src/templates/nextjs-app-router/instrumentation.ts
MenKNas: Atoms styling POC was scoped to the App Router path where Tailwind + Atoms/Design Studio rendering already live (my educated guess it will always remain this way, but in any case for now it's all about App Router). Pages starter has no Tailwind stack and the production compile path (`StudioComponentServerWrapper`) is related only to RSCs/App Router.

art-alexeyenko (approved):
Approving - let's just add some CSS compilation guides and pointers in Pages Router sample before merge.

PR #568, opened 2026-07-27
[nextjs] Fix pending redirects issues

art-alexeyenko:
- [x] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)

## Description / Motivation
This is a composite PR to fix pending redirect proxy bugs, fresh off the heels of the refactor done recently.
- Fixes  locale always being attached on redirect when `isLanguagePreserved` is true, request URL does not have locale prefix and redirect does not change locale
- Fixes Redirect Proxy not matching encoded URLs with unicode characters in Redirect Map

## Testing Details
- [x] Unit Test Added
- [ ] Manual Test/Other (Please elaborate)

## Types of changes
- [x] Bug fix (non-breaking change which fixes an issue)
- [ ] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #577, opened 2026-08-04
[App Router] Configurable redirects locale prefix behavior

art-alexeyenko:
## Description / Motivation
- Reverts the changes around `isLanguagePreserved` from https://github.com/Sitecore/content-sdk/pull/568
- Expands on the configuration from https://github.com/Sitecore/content-sdk/pull/563 to provide a choice on behavior

`next-intl` locale prefix behavior conflicts with SXA redirects' `isLanguagePreserved` switch - the latter switch overrides outcome, regardless of the `languagePrefix` value set for `next-intl`. This PR attempts to make the redirect results more predictable and configurable by renaming and reusing the `localeInPath` setting.

| languagePrefix in next-intl | appLanguagePrefix | Result |
| -------- | ------- | ------- |
|  'as-needed' | undefined   |  Content SDK behavior - isLanguagePreserved overrides next-intl behavior and applies locale when true  |
|  'as-needed' |  'as-needed'   | next-intl behavior - locale prefix is never applied when locale is defaultLanguage |
|  'always'    |  'always'  | next-intl behavior - locale prefix always applied, regardless of isLanguagePreserved |
| 'never'  | 'never' | next-intl behavior - locale prefix is never applied, regardless of isLanguagePreserved |

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/nextjs/src/config/define-config.ts
```diff
@@ -87,11 +93,11 @@ export type SitecoreConfigInput = SitecoreConfigInputCore & {
    */
   redirects?: SitecoreConfigInputCore['redirects'] & {
     /**
-     * Whether redirect targets use a locale path prefix (`/[locale]/...`).
-     * `true` / `false` set App Router behavior; `null` (default) keeps `x-sc-locale` fallback.
-     * @default null
+     * Controls whether App Router redirect targets carry a locale path prefix (`/[locale]/...`).
+     * Mirrors next-intl's `localePrefix` strategy. Only relevant for App Router applications.
+     * @default undefined (treated as `as-needed` with `x-sc-locale` header fallback)
```
illiakovalenko: This doc frames the property as "mirroring next-intl's localePrefix", but there's no actual dependency on next-intl. Since this is the core package's public API doc, could we phrase it as library-agnostic, e.g. "Locale prefix strategy for App Router redirect targets (always/as-needed/never), independent of any specific i18n library. If you use next-intl, these values match its localePrefix strategy 1:1." The "mirrors next-intl" framing feels more at home in the template's sitecore.config.ts comment, where next-intl is actually in use.

packages/nextjs/src/config/define-config.ts
```diff
@@ -59,6 +54,17 @@ export const getNextFallbackConfig = (config?: SitecoreConfigInput): SitecoreCon
   };
 };
 
+/**
+ * Locale prefix strategy for App Router redirect targets, mirroring next-intl's `localePrefix`.
+ * - `always`: every locale is prefixed (`/[locale]/...`), including the default.
+ * - `as-needed`: only non-default locales are prefixed; the default locale stays bare.
+ * - `never`: no locale prefix is applied.
+ * When left `undefined`, RedirectsProxy behaves as `as-needed` and falls back to
+ * `x-sc-locale` header detection (Pages Router when the header is absent).
```
illiakovalenko: The same comment as below, please, double check

illiakovalenko (approved):
This solution looks even better to me - good proposal! 👍 
I have one suggestion below, thinking about the future and keeping our approach library-agnostic.
If you’ve already checked this scenario, great. If not, I’d recommend double-checking it by testing these settings without `LocaleProxy`, following our documentation: https://doc.sitecore.com/sai/en/developers/content-sdk/20/DEFAULT/disable-multisite-and-locale-proxies.html

packages/nextjs/src/proxy/redirects-proxy.ts
```diff
@@ -498,19 +484,15 @@ export class RedirectsProxy extends ProxyBase {
...
-      return 'app-with-locale';
-    }
-    if (this.config.appLocalePrefix === 'never') {
-      return 'app-without-locale';
+  private getRouterMode(req: NextRequest): 'pages' | 'not-pages' {
+    if (req.nextUrl.locale && req.nextUrl.defaultLocale) {
+      return 'pages';
     }
-    return this.isAppRouter(res) ? 'app-with-locale' : 'pages';
+    return 'not-pages';
```
illiakovalenko: Can we support just these two types: `pages` and `app-router`? I think that's how it was initially.

packages/nextjs/src/config/define-config.ts
```diff
@@ -87,11 +82,13 @@ export type SitecoreConfigInput = SitecoreConfigInputCore & {
    */
   redirects?: SitecoreConfigInputCore['redirects'] & {
     /**
-     * Whether redirect targets use a locale path prefix (`/[locale]/...`).
-     * `true` / `false` set App Router behavior; `null` (default) keeps `x-sc-locale` fallback.
-     * @default null
+     * Controls whether App Router redirect targets carry a locale path prefix (`/[locale]/...`).
+     * When explicitly set, mirrors next-intl's `localePrefix` strategy. When undefined, redirects proxy will use 'as-needed' strategy
```
illiakovalenko: Please double-check a comment I left in my previous review. I mentioned that we shouldn't explicitly mention `next-intl` in the SDK, since we don't have a direct dependency on it. It's fine to mention it in the template instead.

.changeset/empty-years-spend.md
```diff
@@ -3,4 +3,4 @@
 'create-content-sdk-app': patch
 ---
 
-Add `redirects.localeInPath` so App Router redirect locale handling works when LocaleProxy is removed (`x-sc-locale` is optional). `true`/`false` control path prefixes; default `null` keeps the `x-sc-locale` header fallback (non-breaking). App Router templates set `true`.
\ No newline at end of file
+Add root-level `appLocalePrefix` config so App Router redirect locale handling works when LocaleProxy is removed (`x-sc-locale` is optional). Mirrors next-intl's `localePrefix` strategy: `always`/`never` control path prefixes explicitly; unset (default) behaves as `as-needed` and keeps the `x-sc-locale` header fallback (non-breaking). App Router templates set `as-needed` so the default locale stays bare on redirect targets (matching their `routing.ts`), while non-default locales are still prefixed.
```
illiakovalenko: Please see my comment regarding mentioning `next-intl` in our SDK. We should avoid referencing it directly there.

.changeset/tasty-pianos-eat.md
```diff
@@ -2,4 +2,6 @@
 '@sitecore-content-sdk/nextjs': patch
 ---
 
-[App router] Locale always attached on redirect when `isLanguagePreserved` is true, request URL does not have locale prefix and redirect does not change locale
+[App router] Locale always attached on redirect when `isLanguagePreserved` is true, request URL does not have locale prefix and redirect does not change locale. Adds new behavior controlled by `appLocalePrefix` setting passed into redirect proxy:
+- When `appLocalePrefix` is undefined, proxy will act based on the `isLanguagePreserved` setting from a Redirect Map - locale prefix will be added on redirect when setting is `true`
+- When `appLocalePrefix` is explicitly set, [next-intl behavior](https://next-intl.dev/docs/routing/configuration#localeprefix) takes preference, regardless of the `isLanguagePreserved` setting.
```
illiakovalenko: Instead let's just list and document available settings here
```suggestion
- When `appLocalePrefix` is explicitly set, takes preference, regardless of the `isLanguagePreserved` setting.
```

illiakovalenko (changes requested):
Please revisit my very first [comment](https://github.com/Sitecore/content-sdk/pull/577#discussion_r3718594897) regarding the `next-intl` references in our docs. Also, please double-check the build failure.

packages/nextjs/src/proxy/redirects-proxy.ts
art-alexeyenko: Since we have just two outcomes here, I replaced the function with the boolean-resulting one. 
This function really doesn't guarantee pages router 100% - but it's used to determine if request should be processed in pages-like manner, so I reflected that in the name too.

.changeset/empty-years-spend.md
```diff
@@ -3,4 +3,8 @@
 'create-content-sdk-app': patch
 ---
 
-Add `redirects.localeInPath` so App Router redirect locale handling works when LocaleProxy is removed (`x-sc-locale` is optional). `true`/`false` control path prefixes; default `null` keeps the `x-sc-locale` header fallback (non-breaking). App Router templates set `true`.
\ No newline at end of file
+Add root-level `appLocalePrefix` config so App Router redirect locale handling works when LocaleProxy is removed (`x-sc-locale` is optional). It controls whether App Router redirect targets carry a locale path prefix (`/[locale]/...`) depending on the value:
```
illiakovalenko: In general, we’re kind of duplicating the changeset we added for the `nextjs` package here. Could we simplify it and just mention that we’ve started leveraging a new setting in our template(s)? create-content-sdk-app doesn't really bring this feature

illiakovalenko (approved):
Looks good to me!
Please see one minor comment below. Once that’s addressed, feel free to merge

PR #579, opened 2026-08-10
[next][angular] Work around fetch causing sitecore-tools:build to fail

art-alexeyenko:
## Description / Motivation
sitecore-tools:build will fail with an exception when the last command utilizes `fetch` (Windows only).  This happens because there are some hanging connections in keep-alive pool that undici leaves, that don't close before sitecore-tools:build calls process.exit().
More info here: https://github.com/nodejs/undici/issues/5680

This PR works around/fixes it for CSDK by introducing a client factory for CLI logic specifically.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/.changeset/public-games-talk.md
+++ b/.changeset/public-games-talk.md
@@ -1,0 +1,5 @@
+---
+'@sitecore-content-sdk/content': patch
+---
+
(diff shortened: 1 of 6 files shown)


illiakovalenko (approved):
👍

PR #580, opened 2026-08-11
[experimental] Add experimental features visibility API

MenKNas:
- [ ] Changelog updated

## Description / Motivation
This PR adds an experimental features visibility endpoint so Sitecore AI (Pages / Deploy) can discover which Content SDK experimental features a deployed editing host supports, and whether each is currently enabled via env vars. This is the first phase of the general plan of adopting the experimental features strategy moving forward.

- New editing-secret–protected API (same pattern as `/api/editing/config`) for Pages Router and App Router
- Feature catalog in `packages/nextjs/src/experimental.json`, returned as JSON with an enabled flag per feature
- Dummy catalog entries included for display/testing - Production default should be `[]`
- New routes are wired to all Next.js templates at `/api/editing/experimental`

## Testing Details
- [x] Unit Test Added

## Types of changes


art-alexeyenko (changes requested):
We need this logic to be reusable across frameworks, and so we'd need a middleware endpoint on Angular too.
The utils, types logic can be reused between next<->angular and some CORS/other logic might be reusable between app and pages route handlers.

illiakovalenko:
> We need this logic to be reusable across frameworks, and so we'd need a middleware endpoint on Angular too. The utils, types logic can be reused between next<->angular and some CORS/other logic might be reusable between app and pages route handlers.

Totally valid point. Let’s don't forget about Angular as well and reduce potential complexity for future work while keeping the solution extensible.
As much as we can without sacrificing clarity, we should aim to reuse functionality and keep shared logic centralized.

MenKNas:
@illiakovalenko  @art-alexeyenko Thanks for the feedback guys, I tried to address your concern(s) by doing the following:

- Moved experimental feature types, utils, and catalog into `@sitecore-content-sdk/content/experimental` so Next.js and Angular share the same core

- Added Angular `createExperimentalFeaturesMiddleware` and wired it in the Angular template (`/api/editing/experimental`) - @art-alexeyenko will need you to take an extra look here

- Extracted shared CORS/Auth for the Next.js Pages + App Router handlers to reduce duplication

packages/content/src/experimental-features/utils.ts
```diff
@@ -0,0 +1,62 @@
...
+export const defaultExperimentalFeatures =
+  experimentalFeaturesCatalog as ExperimentalFeatureData[];
+
+/**
+ * Resolves experimental feature metadata with current enabled status from env vars.
+ * @param {ExperimentalFeatureData[]} features - Experimental features catalog.
+ * @returns {ExperimentalFeatureStatus[]} Features with enabled status.
+ * @public
+ */
+export const resolveExperimentalFeatureStatuses = (
```
art-alexeyenko, line 29: AI hands have written this 😄 
We can probably fold it into `buildExperimentalFeaturesResponse` - or not, up to you

packages/content/src/experimental-features/utils.ts
```diff
@@ -0,0 +1,62 @@
+import experimentalFeaturesCatalog from '../experimental.json';
```
art-alexeyenko: let's call this `experimentalFeaturesCatalogStub` - to ensure this is treated like non-functional fallback. Experimental features should ideally be read per-framework, from framework packages (nextjs/angular) and passed into `buildExperimentalFeaturesResponse `

art-alexeyenko: or alternatively, just don't use this and only pass features JSON into `buildExperimentalFeaturesResponse` function from framework packages

packages/nextjs/src/editing/experimental-features-middleware.ts
```diff
@@ -0,0 +1,71 @@
...
+
+/**
+ * Configuration for {@link ExperimentalFeaturesMiddleware}.
+ * @public
+ */
+export type ExperimentalFeaturesMiddlewareConfig = {
+  /**
+   * Experimental features catalog. Defaults to the shared Content SDK catalog.
+   */
+  features?: ExperimentalFeatureData[];
```
art-alexeyenko: this should be read from an in-package JSON rather than controlled by an app (we don't want someone passing nonsense features, even if it's just for fun)

packages/nextjs/src/route-handler/experimental-features-route-handler.ts
```diff
@@ -0,0 +1,116 @@
...
+
+/**
+ * Options for {@link createExperimentalFeaturesRouteHandler}.
+ * @public
+ */
+export type ExperimentalFeaturesRouteHandlerOptions = {
+  /**
+   * Experimental features catalog. Defaults to the shared Content SDK catalog.
+   */
+  features?: ExperimentalFeatureData[];
```
art-alexeyenko: similarly, let's read this from common framework package file - unless this is a new approach different from the PR description

packages/angular/src/server/middleware/experimental-features-middleware.ts
```diff
@@ -0,0 +1,103 @@
...
+export interface CreateExperimentalFeaturesMiddlewareOptions {
+  /**
+   * Experimental features catalog. Defaults to the shared Content SDK catalog.
+   */
+  features?: ExperimentalFeatureData[];
+  /**
+   * Editing secret to validate. Defaults to the `SITECORE_EDITING_SECRET`
+   * environment variable.
+   */
+  editingSecret?: string;
```
art-alexeyenko: we probably don't need this override - either read process.env value directly and it will be the same secret used across the app, or it will be undefined if *somehow* the middleware executes on client

packages/angular/src/server/middleware/experimental-features-middleware.ts
```diff
@@ -0,0 +1,103 @@
...
+
+/**
+ * Options for {@link createExperimentalFeaturesMiddleware}.
+ * @public
+ */
+export interface CreateExperimentalFeaturesMiddlewareOptions {
+  /**
+   * Experimental features catalog. Defaults to the shared Content SDK catalog.
+   */
+  features?: ExperimentalFeatureData[];
```
art-alexeyenko: And same comment about features here

art-alexeyenko (approved):
Good stuff @MenKNas 
Approving now - I trust you will address the comments before merging.

PR #583, opened 2026-08-12
[Chore] Betas branch process prep

art-alexeyenko:
This PR intends to help with the beta release process. In order to do so:
- It makes canary prefix readable from env (`CANARY_PREFIX` variable)
  - prefix is set via GH actions
  - release snapshot scripts will read it from env
  - tag/prefix falls back to `canary` when not set in env
- canary prefix/tag is set based on branch conventions
- beta tag is derived from beta branch naming (i.e. `release/betas/dean` -> `beta.dean`)

Bonus: this PR also adds dynamic npm tag calculation for full releases (`latest` for publishes from `dev` branch, `ver1` for releases from `release/1.x.x and `legacy` for releases from the rest of applicable `release/` branches. The latter is a placeholder logic for now, to be refined later).

Final goal is to make the prep for beta branches easier than losing a $10 in Vegas (full beta process now in internal docs).

## Testing Details
[x] Release unit tests updated.
[x] Github actions YAML verified


illiakovalenko (approved):
Looks good to me 👍 
Moving forward, as planned, we won’t use `release/<version>` branches anymore. Instead, branches can be framework (product) specific. We will adjust this as part of the future work.

PR #585, opened 2026-08-12
[angular] Add loader prefetch on link hover/render (scRouterLink, scRichText)

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
Implements loader prefetch for `scRouterLink`/`scRichText` links in the Angular SDK, so route data is already loaded (or in flight) by the time the user actually navigates.

Three modes, selected globally via `sitecore.config`'s `angular.linkPrefetch.mode` or overridden per link via each directive's `prefetch` input:
- **`eager` (default)** - prefetch eagerly, as soon as the link renders.
- **`hover`** - defer until the pointer dwells on the link for `angular.linkPrefetch.delayMs` (default 100ms). Re-arms on every hover, rather than firing once and never again — a link whose earlier prefetch was already consumed by a real navigation correctly prefetches fresh data on a later hover.
- **`off`** - never prefetch.

Key pieces:
- `matchRouteChain` (`loaders/route-matcher.ts`) - resolves which loaders apply to a raw `href`, without triggering navigation, by walking `Router.config` with Angular's own public matching primitives (`defaultUrlMatcher`, custom `route.matcher`s). No private Router internals involved.
- `ClientPreLoaderDataService.prefetchForUrl()` - the entry point directives call; discovers `LOADER_ID`-tagged resolvers across the matched route chain (parent + child) and passes down to `ClientLoaderDataService.prefetch()`.
- `ClientLoaderDataService.prefetch()` gained a `force` option - hover callers pass `force: true` so a still-staged-but-unconsumed answer never suppresses a legitimate re-ask; eager callers omit it, so eager prefetching still dedupes permanently once staged.
- Both `ScRouterLinkDirective` and `ScRichTextDirective` wire this up, gated on platform (browser only), editing/preview mode, and external/`target="_blank"`/empty hrefs.

Also fixes a real route-matching bug found during testing: `matchRouteChain` would stop at a pass-through parent route (e.g. a locale matcher wrapping the app's catch-all page route) instead of descending into its children when the URL had already-zero remaining segments - meaning **the home page specifically never got its loaders discovered**, silently prefetching nothing. Covered by new regression tests.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/angular/src/loaders/pre-loader-data.service.ts
```diff
@@ -85,35 +81,80 @@ export class ClientPreLoaderDataService {
...
+   * @param {boolean} [options.force] - Forwarded to {@link ClientLoaderDataService.prefetch} — bypass the staged-response check. Hover callers pass `true` (every hover re-asks); eager callers omit it (dedupes permanently once staged).
+   */
+  prefetchForUrl(url: string, options?: { force?: boolean }): void {
+    if (!isPlatformBrowser(this.platformId)) {
+      return;
+    }
+
+    let chain: ReturnType<typeof matchRouteChain>;
+    let query: Record<string, string | string[]>;
+    try {
```
art-alexeyenko, line 135: The algorithm between this and `prefetchForRoute` is generally the same. To keep things DRY I'd replace `collectLoaders` with a function with common logic reused between `prefetchForRouter` and `prefetchForUrl`:
- gets loader Ids for each route segment + route params
- launches prefetch calls

packages/angular/src/loaders/pre-loader-data.service.ts
```diff
@@ -85,35 +81,80 @@ export class ClientPreLoaderDataService {
...
+      const tree = this.router.parseUrl(url);
+      const segments = tree.root.children[PRIMARY_OUTLET]?.segments ?? [];
+      chain = matchRouteChain(this.router.config, segments);
+      query = tree.queryParams as Record<string, string | string[]>;
+    } catch {
+      // Malformed href (e.g. invalid percent-encoding) — not this feature's job to surface.
```
art-alexeyenko, line 141: In addition to this, we need some guards for:
- non-angular route URLs (not matching the route tree)
- absolute URLs (would usually be external)

packages/angular/src/config/define-config.ts
```diff
@@ -1,6 +1,16 @@
...
+
+/**
+ * Link-prefetch strategy for `scRouterLink`/`scRichText` links:
+ * - `true` (default) — prefetch eagerly, as soon as the link renders.
+ * - `'hover'` — prefetch only once the pointer dwells on the link (see `delayMs`); starts
+ *   disabled and is enabled the moment the user shows intent.
+ * - `false` — never prefetch.
+ * @public
+ */
+export type LinkPrefetchMode = boolean | 'hover';
```
art-alexeyenko: for better clarity, let's just use the `eager`, 'hover' and 'off' values here - saves us the need for extra explanations in comments and makes the usecases obvious at a glance.

art-alexeyenko (changes requested):

packages/angular/src/loaders/pre-loader-data.service.ts
illiakovalenko: Good call. I extracted a `prefetchChain` that both methods now call: it walks a root -> leaf list of `{ resolve, params }` levels, collects the `LOADER_ID` resolvers per level, merges params per level, and launches the prefetch calls.

packages/angular/src/loaders/pre-loader-data.service.ts
illiakovalenko: Added guards
* Absolute / external URLs - prefetchForUrl now returns early via a new `isAbsoluteUrl` helper that flags anything carrying a scheme
* Non-matching route URLs - this was already covered, `matchRouteChain` returns null for a URL that doesn't match the route tree, and `prefetchForUrl` bails on it

packages/angular/src/config/define-config.ts
illiakovalenko: Agree, this shape gives a better clarity. Initially I had an idea to keep it more consistent to nextjs api, however it's not required

packages/angular/src/components/field-directives/sc-router-link.directive.ts
```diff
@@ -1,20 +1,44 @@
...
+ * - `'off'` — never prefetch.
+ *
+ * Override per link with `scRouterLinkPrefetch`.
+ *
  * Editing chrome + empty-field placeholder behavior is inherited from {@link ScLinkDirective}.
  *
  * Usage:
  * ```html
  * <a *scRouterLink="fields.Link">Optional child content</a>
+ * <!-- scRouterLinkPrefetch is another input of *this same* structural directive, so
```
art-alexeyenko: can this tsdoc be potentially shortened? the way it is, is a bit too much imo

packages/angular/src/loaders/pre-loader-data.service.ts
```diff
@@ -68,52 +65,108 @@ export class ClientPreLoaderDataService {
...
+        this.defaultLanguage
+      );
+      for (const loaderId of loaderIds) {
+        this.loaderData.prefetch({ loaderId, url, routeParams, query }, options);
       }
+    });
+  }
+
+  /**
+   * Resolves the loaders that would apply to `url` — without navigating — and prefetches each
```
art-alexeyenko: similar comment here, tsdoc is larger than the function

packages/angular/src/loaders/route-matcher.ts
```diff
@@ -0,0 +1,146 @@
...
+  return { consumed: result.consumed, posParams: params };
+}
+
+/**
+ * Recursively matches `routes` against `segments`, first-match-wins in declaration order —
+ * mirroring Angular Router's own matching semantics using only public Router primitives
+ * (`defaultUrlMatcher`, custom `route.matcher`s). Used to determine, from a raw href, which
+ * loader-bearing routes apply without triggering an actual navigation.
+ *
+ * Known v1 limitations (routes are skipped rather than matched incorrectly):
```
art-alexeyenko: I think this part is not neccessary?

art-alexeyenko (approved):
Thank you, this looks really good.
Ready to be shipped - but also please review tsdocs to not be so large

PR #587, opened 2026-08-13
[Angular] Import map generation

art-alexeyenko:

## Description / Motivation
Adds import map generation to Angular sample, in preparation for Design Studio support.
In addition to this, PR contains some patches for shared import map and getComponentList logic:
- more strict filtering of variant components (no false positives when component name is something like `myLogic.component`)
- make import-map generation include variants by default

Otherwise, implementation mirrors that of Next.JS apps (with adjustments for Angular types and quirks).

## Testing Details
- [x] Unit Test Added

## Types of changes


packages/content/src/tools/codegen/import-map.ts
```diff
@@ -80,6 +80,7 @@ type ImportMapEntry = {
 export type WriteImportMapArgs = {
   paths: string[];
   exclude?: string[];
+  includeVariants?: boolean;
```
sc-nikolaoslazaridis, line 83: Altho optional, isn't going to affect the next js apps? If this is only related to Angular i would not add that here to keep this one purely generic.

packages/angular/src/server/codegen/import-map-utils.ts
```diff
@@ -0,0 +1,44 @@
...
+/**
+ * Combines the default import entries with the generated import entries.
+ * Generated entries take precedence; default entries are merged in only when the module
+ * (or a specific named export within a shared module) is not already present.
+ * @param {ImportEntry[]} defaultImportEntries - The framework/built-in import entries.
+ * @param {ImportEntry[]} generatedImportEntries - The import entries produced from app components.
+ * @returns {ImportEntry[]} The combined import entries.
+ * @public
+ */
+export const combineImportEntries = (
```
sc-nikolaoslazaridis, line 10: Looks like can be a generic one unless we already have it...

packages/content/src/test-data/components/Hero.tsx
```diff
@@ -0,0 +1 @@
+﻿
```
sc-nikolaoslazaridis: Plase fix

packages/angular/src/server/tools/write-import-map.ts
```diff
@@ -0,0 +1,37 @@
...
+  args: WriteImportMapArgs
+): ((args: { scConfig: AngularSitecoreConfig }) => Promise<void>) => {
+  const command = writeImportMapContent({
+    ...args,
+    defaultTemplate: angularDefaultMapTemplate,
+  });
+  return async ({ scConfig }: { scConfig: AngularSitecoreConfig }) => {
+    // AngularSitecoreConfig is a structural superset of SitecoreConfig; only redirects.locales
+    // and the Express-typed multisite callback differ, neither of which the import map reads.
+    await command({ scConfig: scConfig as unknown as SitecoreConfig });
```
sc-nikolaoslazaridis: This casting seems dangerous. Can we properly adapt the types here?

packages/content/src/tools/templating/components.ts
```diff
@@ -156,9 +156,13 @@ function _getComponentList(
     );
   }, []);
 
-  return includeVariants
-    ? components
-    : components.filter((component) => !component.componentName.includes('.'));
+  if (includeVariants) return components;
+
+  const componentSet = new Set(components.map((c) => c.componentName));
```
sc-nikolaoslazaridis: An edge case i got in my mind and it is not blocking, just to  open the discussion if you feel its right. The component set is taking into account the component names. What if in another folder we have the same component name. 

Should we take into consideration the "current folder" to make less error-prone?

sc-nikolaoslazaridis (commented):
Do we need a changeset for the create-content-sdk-app package?

packages/content/src/tools/codegen/import-map.ts
art-alexeyenko: We technically fix a bug here. We want import-map to pull everything, including variants by default.

packages/content/src/tools/templating/components.ts
art-alexeyenko: this complicates the filtering logic somewhat, but yeah, good call and I think it's worth it

packages/nextjs/src/editing/codegen/import-map-utils.test.ts
```diff
@@ -1,146 +1,17 @@
...
-            { name: 'C', value: C },
-            { name: 'D', value: D },
-          ],
-        },
-        {
-          module: 'module-c',
-          exports: [{ name: 'E', value: E }],
-        },
-      ];
+describe('Module directives', () => {
```
sc-nikolaoslazaridis, line 4: I'm a bit confused on what is left in this file. What are we actually testing here?

art-alexeyenko: originally it tested import-utils + the shape of default map templates. The shape tests remain

sc-nikolaoslazaridis (approved):
Looks good! (let me know if i need to check it again after you fix the build job)

PR #589, opened 2026-08-14
feat[search]: add SearchService.suggest typeahead API

giglyonsense:
- [x] This PR follows the [Contribution Guide](https://github.com/Sitecore/content-sdk/blob/dev/CONTRIBUTING.md)
- [x] Changelog updated
- [ ] Upgrade guide entry added

## Testing Details

- [x] Unit Test Added
- [x] Manual Test/Other (Please elaborate)

## Types of changes

- [ ] Bug fix (non-breaking change which fixes an issue)
- [x] New feature (non-breaking change which adds functionality)
- [ ] Breaking change (fix or feature that would cause existing functionality to change)

PR #592, opened 2026-08-17
[search] - feat(suggest API onboarding) + chore(MLT support)

giglyonsense:
- [x] Changelog updated

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/search/src/search-service.ts
```diff
@@ -226,5 +359,73 @@ export class SearchService {
...
+    if (provided.length > 1) {
+      throw new TypeError(
+        `Query fields are mutually exclusive. Provide only one of: keyphrase, seedItemId, seedItemUrl. Received: ${provided.join(
+          ', '
+        )}`
+      );
+    }
+  }
+
+  private buildSearchQuery<T extends SearchDocument = SearchDocument>(
```
art-alexeyenko: small nitpick, let's call this `buildSearchUrlQuery` to better describe the intention

packages/search/src/search-service.ts
```diff
@@ -113,6 +139,59 @@ export interface SearchParameters<T extends SearchDocument = SearchDocument> {
...
+   * Document previews from preview results mode.
+   */
+  previewResults: T[];
+}
+
+/**
+ * Response from the Suggest Service.
+ * @public
+ */
+export interface SuggestResponse<T extends SearchDocument = SearchDocument> {
```
art-alexeyenko, line 169: This and the interface above are the same shape. Are both of them needed?

packages/search/src/search-service.ts
```diff
@@ -182,9 +266,7 @@ export class SearchService {
         },
         limit,
         offset,
-        query: {
-          keyphrase,
-        },
+        query: this.buildSearchQuery(params),
```
art-alexeyenko: seedItemId, seedItemUrl, keywords are mutually exclusive, but there's no validation for that in this current path.  If all 3 are passed, only one is used with no indication where the other 2 go.

packages/search/src/search-service.ts
```diff
@@ -226,5 +359,73 @@ export class SearchService {
...
+    }
+
+    if (seedItemUrl) {
+      return { seedItemUrl };
+    }
+
+    return { keyphrase: params.keyphrase ?? '' };
+  }
+
+  private validateSuggestParameters(params: SuggestParameters) {
```
art-alexeyenko, line 426: If there's absolutely no way to bring these 3 under one umbrella/one field - so that it's very obvious from the contract shape they are mutually exclusive - let's call this validation as soon into execution as possible, to do a fail fast situation.

packages/search/src/search-service.ts
```diff
@@ -202,6 +284,57 @@ export class SearchService {
...
+  /**
+   * Retrieve typeahead suggestions for a keyphrase.
+   * @param {SuggestParameters} params - The suggest parameters.
+   * @param {SearchServiceFetchOptions} [fetchOptions] - The fetch options.
+   * @returns {Promise<SuggestResponse<T>>} The suggest response.
+   * @throws {NativeDataFetcherError} if the request fails.
+   * @throws {TypeError} If search index ID is not provided.
+   * @throws {TypeError} If keyphrase is not provided or is empty.
+   */
+  async suggest<T extends SearchDocument = SearchDocument>(
```
art-alexeyenko, line 281: Unless I'm missing something - this is not being used anywhere right now in this PR, but `useSuggest` is present in feature branch. Can you double check the branch is healthy, up to date with `dev` and reports everything needed for the feature please?

art-alexeyenko (changes requested):
Thanks for the submission!
I see this PR overlaps a lot with https://github.com/Sitecore/content-sdk/pull/589 - should they be merged into one PR instead?

Also, I'd ask @sc-nikolaoslazaridis to take a look too in case I missed something in terms of search functionality.

packages/search/src/search-service.ts
giglyonsense: Removed the redundant one.

packages/search/src/models.ts
```diff
@@ -95,6 +95,46 @@ export type SearchDocument = {
...
 };
 
+/**
+ * Query payload for the `/v1/search` API.
+ * `keyphrase`, `seedItemId`, and `seedItemUrl` are mutually exclusive; provide at most one.
+ * `seedItemId` and `seedItemUrl` are used for More Like This (MLT) queries and are not
+ * supported by `/v1/search/suggest`.
+ * @public
+ */
+export interface SearchQuery {
```
sc-nikolaoslazaridis: We need to make the type reflect the business rules so the developers can get feedback during dev time instead on runtime. Can you please do that?

packages/search/src/search-service.ts
```diff
@@ -226,5 +344,94 @@ export class SearchService {
...
+    const fields = this.getTrimmedQueryFields(params);
+    const provided = this.getProvidedQueryFields(fields);
+
+    switch (provided[0]) {
+      case 'seedItemId':
+        return { seedItemId: fields.seedItemId };
+      case 'seedItemUrl':
+        return { seedItemUrl: fields.seedItemUrl };
+      case 'keyphrase':
+        return { keyphrase: params.keyphrase ?? '' };
```
sc-nikolaoslazaridis: Is there a reason we do not use the trimmed value like we do for the other 2 cases?

packages/search/tsconfig.json
```diff
@@ -14,5 +14,6 @@
     "types",
     "dist",
     "src/**/*.test.ts",
+    "scratch"
```
sc-nikolaoslazaridis: Is this needed here?

packages/search/src/search-service.ts
```diff
@@ -139,16 +206,18 @@ export class SearchService {
    * @throws {RangeError} If offset is not a positive number.
    * @throws {TypeError} If search index ID is not provided.
    * @throws {TypeError} If sort is not an array or an object.
+   * @throws {TypeError} If more than one of keyphrase, seedItemId, or seedItemUrl is provided.
+   * @throws {TypeError} If seedItemId or seedItemUrl is empty or whitespace only.
    */
   async search<T extends SearchDocument = SearchDocument>(
```
sc-nikolaoslazaridis, line 212: It looks like the query exclusivity validation runs multiple times. Can we run it once and any code that comes after it assumes that it is already validated?

packages/search/src/search-service.ts
```diff
@@ -226,5 +344,94 @@ export class SearchService {
...
+
+    return provided;
+  }
+
+  private validateQueryExclusivity<T extends SearchDocument = SearchDocument>(
+    params: SearchParameters<T>
+  ) {
+    const fields = this.getTrimmedQueryFields(params);
+    const provided = this.getProvidedQueryFields(fields);
+
```
sc-nikolaoslazaridis, line 388: I don't see a validation for the keyphrase here

packages/search/src/search-service.ts
```diff
@@ -226,5 +344,94 @@ export class SearchService {
...
+    if (params.seedItemId !== undefined && !fields.seedItemId) {
+      throw new TypeError('seedItemId must be a non-empty string');
+    }
+
+    if (params.seedItemUrl !== undefined && !fields.seedItemUrl) {
+      throw new TypeError('seedItemUrl must be a non-empty string');
+    }
+
+    if (provided.length > 1) {
```
sc-nikolaoslazaridis: I would move this as the first check to avoid running multiple checks. The case that all three values are set but then failing due to the bussines rules.

sc-nikolaoslazaridis (changes requested):
Please check the comments and come back questions if have any

packages/search/src/search-service.ts
giglyonsense: Yes, keyphrase can be empty for the search endpoint. But the suggest endpoint needs a non-empty keyphrase, hence the validation is added for the suggest endpoint which is added in the function `validateSuggestParameters`.

packages/search/src/models.ts
giglyonsense: image
Yes, they both are the same, except the interface name.

packages/search/src/search-service.ts
giglyonsense: It is missed, I will update it.

packages/search/tsconfig.json
giglyonsense: Removed.

packages/search/src/search-service.ts
giglyonsense: Removed the duplicate check. Thanks for noticing.

packages/search/src/search-service.ts
giglyonsense: Makes sense. I will update it.

packages/search/src/models.ts
sc-nikolaoslazaridis: I think i don't get it. We need the type to reflect inside the content sdk.

packages/search/src/search-service.ts
giglyonsense: Added a react hook for the same.

packages/search/src/models.ts
giglyonsense: @sc-nikolaoslazaridis 
Updated the model as discussed. Kindly take a look.
packages/search/src/models.ts

sc-nikolaoslazaridis (approved):
Looks good on my side, please don't merge it until Artem also approves it.

art-alexeyenko (approved):
Looks good on my end too. Thanks!

PR #593, opened 2026-08-18
[core] Content SDK - sitecore-tools project build fails when using pnpm because harded "npm query" in metadata.ts

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

- Metadata collection no longer always shells out to `npm query`.
- The active package manager is detected from `npm_config_user_agent` (fallback: `npm_execpath`), then the matching listing command is used.
- This covers **pnpm, yarn (classic and berry), bun, and npm**, so the same fix works across the popular package managers—not only pnpm.

## Why
Spawned package-manager commands should follow the manager that is already running the build, so metadata generation stays correct in non-npm projects (including cases where `devEngines.packageManager` restricts the project to another manager).

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/core/src/tools/metadata/metadata.ts
```diff
@@ -15,21 +52,45 @@ export interface Metadata {
 
 const trackedScopes = ['@sitecore', '@sitecore-feaas', '@sitecore-content-sdk'];
 
+// Longest name first, since 'pnpm' contains 'npm' and would be matched by it.
+const packageManagerNames: PackageManagerName[] = ['pnpm', 'yarn', 'bun', 'npm'];
```
art-alexeyenko: I'd suggest a rewrite of the manager name and query calculation. Package managers, their queries and the query parser are implicitly tied together through a type, an array, a function with different query outcomes. We can streamline that in one place and tie them explicitly instead, creating a tightly coupled contract, something like:
```
const packageManagement: Record<string, PackagesQuery> = {
    pnpm: {
      query:
        `pnpm list --depth Infinity --json ${
          allowWorkspaces ? '--recursive ' : ''
        }${TRACKED_SCOPE_GLOB}`,
      parse: parsePnpmList,
    },
    npm: {
      query: `npm query [name*=@sitecore] --workspaces ${allowWorkspaces}`,
      parse: parseNpmQuery,
    },
    yarn: {...},
    yarnClassic: {...},
    bun: {...}
  };
```
This way we can:
- Ensure package manager name and logic are visibly tied - better readability, maintainability
- The relations between managers and logic is visible at one glance
- Ensure there's no possible drift, and one identified manager will not be using queries/parser from another
- Still be able to filter the current manager by name (from `Object.keys(packageManagement)`)

packages/core/src/tools/metadata/metadata.ts
```diff
@@ -15,21 +52,45 @@ export interface Metadata {
...
+// pnpm and yarn match name patterns as globs, in which '*' does not cross the scope separator,
+// so both segments are wildcarded to cover every '@sitecore*' scope.
+const TRACKED_SCOPE_GLOB = '"@sitecore*/*"';
+
+const PNPM_DEPENDENCY_GROUPS = [
+  'dependencies',
+  'devDependencies',
+  'optionalDependencies',
+  'unsavedDependencies',
+] as const;
```
art-alexeyenko: probably an unneeded cast

art-alexeyenko (commented):
Great work - and thank you for seeing it through.
This is kind of good to go as it is now, but I've left some comments for your consideration on how we can improve the implementation and make it more resilient.

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: Thanks for the review! 

I've made the requested changes. Each package manager is now an entry in one `Record` (`pnpm`, `npm`, `yarn`, `yarnClassic`, `bun`) with its own `query` + `parse`. Detection uses `Object.keys(packageManagement)`, and the selected key is used to look up that same record, so a manager cannot pick up another manager’s command or parser.

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: Removed

packages/core/src/tools/metadata/metadata.ts
```diff
@@ -54,3 +101,249 @@ function getPackagesFromQueryResult(scPackages: Package[]): Record<string, strin
...
+}
+
+/**
+ * Parse an NDJSON stream, skipping lines that are not valid JSON - package managers mix
+ * diagnostics into their output
+ * @param {string} output command output
+ * @returns {any[]} parsed entries
+ * @internal
+ */
+// eslint-disable-next-line @typescript-eslint/no-explicit-any
```
art-alexeyenko: Let's get rid of using `any` as type. Consider using `unknown` or a specific type

packages/core/src/tools/metadata/metadata.ts
```diff
@@ -5,6 +5,31 @@ type Package = {
   version: string;
 };
 
+/**
+ * Package manager identifiers used for metadata collection.
+ * Yarn classic is a separate key because its listing command differs from yarn berry.
+ * @internal
+ */
+type PackageManagerName = 'npm' | 'pnpm' | 'yarn' | 'yarnClassic' | 'bun';
```
art-alexeyenko: probably don't need this class, can just use `string`. What do you think?

art-alexeyenko (changes requested):

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: Switched parseJsonLines to unknown[]. JSON.parse has no fixed shape, so TypeScript will not allow .value / .type / .name until the value is narrowed. isRecord and getRecordString do that narrowing for the Yarn parsers.

art-alexeyenko (approved):

PR #595, opened 2026-08-19
[content][react] Enhance withDatasourceCheck() to support isContentResolved Prop

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

- Add optional `isContentResolved` to `ComponentRendering`.
- Treat `isContentResolved: false` the same as a missing datasource in `withDatasourceCheck()` (hide in normal mode, editing error in editing mode).
- `isContentResolved: true` means the datasource content was resolved, so the wrapped component should render.
- Preserve existing behavior when the property is omitted so older Layout Service versions (for example SXP 10.5, which does not emit this field) keep using the original `dataSource` presence check.
- This flag is the opposite of `dataSourceResolveFailed`: `isContentResolved: false` is the failed-resolve case.

## Testing Details

- [x] Unit Test Added

Manual checks:
- Confirm `isContentResolved: true` still renders the wrapped component (normal and editing).
- Confirm `isContentResolved: false` hides the component in normal mode and shows the editing error in editing mode (even if `dataSource` is present).
- Confirm omitted `isContentResolved` (older Layout Service / SXP 10.5) still uses the original `dataSource` presence check.
- Confirm nested placeholder renderings are evaluated independently.

## Types of changes


packages/content/src/layout/models.ts
```diff
@@ -103,6 +103,12 @@ export interface ComponentParams {
 export interface ComponentRendering<T = ComponentFields> {
   componentName: string;
   dataSource?: string;
+  /**
+   * `true` when Layout Service failed to resolve this rendering's datasource item
+   * (for example because the item was deleted or archived).
+   * Omitted by older Layout Service versions; absence preserves existing behavior.
+   */
+  dataSourceResolveFailed?: boolean;
```
art-alexeyenko: The name returned by `rendered` field in layout query is `isContentResolved` - please double check we're using the correct naming, and ensure the field on `ComponentRendering` is correctly populated

art-alexeyenko (changes requested):

packages/content/src/layout/models.ts
sc-naveenhedallaarachchi: Thanks, confirmed.

Layout Service returns isContentResolved on each rendering in rendered. ComponentRendering now uses that field. withDatasourceCheck() treats isContentResolved === false as invalid (same fallback as a missing datasource); omitted/true keeps existing behavior.

Verified on a QA Edge env with the local app: valid DS - true and renders; unresolved DS (path still set) - false, editing error in Pages, hidden in normal mode.

art-alexeyenko (approved):
Good stuff, thank you Naveen.

art-alexeyenko (approved):

PR #602, opened 2026-08-25
[core] fix: include all Sitecore packages in metadata.json when using pnpm

sc-naveenhedallaarachchi:
- [ ] Changelog updated

### Issue
`pnpm sitecore-tools project build` omitted Sitecore packages from `metadata.json` compared to npm. Pages Router missed `@sitecore-feaas/clientside` and `@sitecore/components`; App Router also missed `@sitecore-content-sdk/personalize`.
Cause: `pnpm list` with a name filter silently dropped packages. Unfiltered JSON output was too large for `execSync` (`ENOBUFS`).
### Solution
Use `pnpm list --depth Infinity --parseable --long` and parse each `path:name@version` line. No `package.json` reads. npm, yarn, yarn classic, and bun are unchanged.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/core/src/tools/metadata/metadata.ts
```diff
@@ -105,9 +94,12 @@ function getPackagesFromQueryResult(scPackages: Package[]): Record<string, strin
 function getPackageManagement(allowWorkspaces: boolean): Record<string, PackagesQuery> {
   return {
     pnpm: {
-      query: `pnpm list --depth Infinity --json ${
-        allowWorkspaces ? '--recursive ' : ''
-      }${TRACKED_SCOPE_GLOB}`,
+      // `pnpm list` silently drops packages from its result when it is given a name pattern, and
+      // the tree it prints as JSON repeats every shared dependency, which overruns the output
+      // buffer of real apps. Its parseable output lists each installed package once instead.
+      query: `pnpm list --depth Infinity --parseable --long${
```
art-alexeyenko: are `--depth` and `--long` really needed here?

packages/core/src/tools/metadata/metadata.ts
```diff
@@ -33,20 +24,18 @@ export interface Metadata {
...
-  'unsavedDependencies',
-];
-
 // Matches the 'name@version' entries of the tree printed by `bun pm ls`, e.g '├── @scope/pkg@1.0.0'
 const BUN_PACKAGE_ENTRY = /^[\s│├└─]*((?:@[^\s@/]+\/)?[^\s@/]+)@([^\s]+)$/;
 
+// Matches the trailing 'name@version' of the 'path:name@version' entries printed by
+// `pnpm list --parseable --long`. The leading path is excluded from the match because it holds
+// directory separators, and a drive letter followed by a colon on Windows.
+const PNPM_PACKAGE_ENTRY = /:((?:@[^\s@:/\\]+\/)?[^\s@:/\\]+)@([^\s]+)$/;
```
art-alexeyenko: can this and BUN_PACKAGE_ENTRY be combined? If we don't need to match the `├──` prefix for Bun, same regex can be used for both managers.

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: --depth Infinity is needed because pnpm list defaults to depth 0 and skips nested packages. --long is needed so the output includes name@version; --parseable alone is only the install path.

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: Combined them. Bun still needs the ├── prefix; pnpm uses the path: delimiter instead, so that prefix is not applied to pnpm lines. A plain name@version regex would misread pnpm store paths.

packages/core/src/tools/metadata/metadata.ts
art-alexeyenko: Perhaps we can set depth to a more modest number like 2 or 3. @sitecore packages are never nested deep - and use the `@sitecore*/**` selector to have the sitecore packages and have a shorter processing loop.

packages/core/src/tools/metadata/metadata.ts
art-alexeyenko: But you already match the end of line here, it will match the precise place in output that has the package names and version. You can simply reuse the single regex between bun and pnpm, for example:
```
(?:@[^\s\/\\]+\/[^\s\/\\]+)@(\d+\.\d+\.\d+(?:-[a-z]+.[0-9]+)?)$
```

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: Depth 1 misses search; using --depth 3.

packages/core/src/tools/metadata/metadata.ts
sc-naveenhedallaarachchi: Done

art-alexeyenko (approved):

PR #604, opened 2026-08-26
AI Discoverability: Native Metadata Support for head application

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
**Add metadata and Open Graph tags to Next.js scaffolding templates**

Renders SEO metadata and Open Graph tags from Sitecore page fields in the generated Next.js head apps (`nextjs`, `nextjs-app-router`, `nextjs-app-router-cache-components`), so pages produced by the Content SDK carry proper `<title>`/`<meta>`/OG tags without any manual per-app wiring.

**SDK types** (`@sitecore-content-sdk/content`, re-exported via `react`/`nextjs`):
- `MetadataFields` (`baseMetadataTitle/Description/Author/Keywords`) and `OpenGraphFields` (`baseOgTitle`, `baseOgDescription`, `baseOgImage`, `baseOgType`), combined as `PageMetadataFields`. All 6 text fields are `Field<string>` (guaranteed valid strings from the backend — no `.toString()` needed at call sites).
- `OpenGraphImageField`/`OpenGraphImageFieldValue` — a dedicated type for `baseOgImage`, kept separate from the existing `ImageField`/`ImageFieldValue` used for rendered `<Image>`/`<NextImage>` fields.
- `RouteData.published`/`RouteData.updated` — new route-level (not `fields`-nested) ISO datetime properties.
- `OG_CREATION_TIME_TAG`/`OG_MODIFIED_TIME_TAG` — official Open Graph protocol tag-name tables (`article` → `article:published_time`/`article:modified_time`, `book`/`music.album`/`video.movie`/`video.episode` → their `*:release_date` tag), framework-agnostic since the same tag names apply regardless of rendering framework.
- Each template's `RouteFields` extends `PageMetadataFields` instead of redeclaring the fields locally — one reusable shape shared by every app.

**Field-to-tag mapping** (identical across all 3 templates, no cross-field fallback — a field with no value omits only its own tag):
- `<title>` always comes from the existing `Title` field only (falling back to `'Page'`) — `baseMetadataTitle` never feeds `<title>`, it renders its own `<meta name="title">` tag instead.
- `baseMetadataDescription/Author/Keywords` → `description`/`author`/`keywords` meta tags.
- `baseOgTitle/Description/Image` → `og:title`/`og:description`/`og:image(:width|:height|:alt)`.
- `baseOgType` → `og:type`; when set, also renders the correct creation-time tag from `route.published` (`article:published_time`, `book:release_date`, `music:release_date`, `video:release_date`) and, for `article` only, `article:modified_time` from `route.updated` (the only Open Graph type with a distinct update-time tag per the protocol).

**New reusable `@sitecore-content-sdk/nextjs` helpers** (replacing per-template inline derivation):
- `getPageMetadata(route, defaultTitle?)` — builds a ready-to-use Next.js `Metadata` object; both App Router templates' `generateMetadata` now call it directly instead of duplicating the field-derivation/`openGraph` block.
- `PageMetaTags` — a component rendering the equivalent tags via `next/head`, used standalone in the Pages Router template's `Layout.tsx` in place of inline JSX.
- Both share one internal field-resolution function, so title precedence/omission rules stay identical across all three templates.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
```diff
@@ -3,32 +3,69 @@
...
+
   return (
     <>
       <Scripts />
       <SitecoreStyles layoutData={layout} />
       <Head>
-        <title>{fields?.Title?.value?.toString() || 'Page'}</title>
+        <title>{title}</title>
         <link rel="icon" href="/favicon.ico" />
+        {metaTitle && <meta name="title" content={metaTitle} />}
```
art-alexeyenko: not that it's 100% necessary, but IMO it would be more manageable if we put all the meta tags in a separate component

packages/nextjs/src/metadata/page-metadata.ts
```diff
@@ -0,0 +1,80 @@
+import type { Metadata as NextMetadata } from 'next';
+import type { Field, PageMetadataFields, RouteData } from '@sitecore-content-sdk/content/layout';
+
+/** Open Graph types that define a creation-time field, and which `Metadata.openGraph` key it maps to. */
+const OG_CREATION_TIME_FIELD: Record<string, 'publishedTime' | 'releaseDate'> = {
```
art-alexeyenko: this one isn't react or next specific, right? Should it be moved to content for reuse?

packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
art-alexeyenko: and should/could we reuse `getPageMetadata` here?

art-alexeyenko (approved):
Looks good, thank you!
Left a couple comments for your consideration.

packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
sc-nikolaoslazaridis: Reusing `getPageMetadata` here would be possible only with extra translation logic so i wouldn't do it.

art-alexeyenko (approved):

PR #609, opened 2026-09-01
[bug][content] Revert to using raw component name in getComponentList

art-alexeyenko:
- [ ] Changelog updated

## Description / Motivation
- Fixes a regression in component/import map generations where components with underscore in names were not properly processed
- toPascalCase util no longer converts `snake_case`

Note: eventually we should enforce a convention for component names in Content SDK, which may be a breaking change.

## Testing Details
- [x] Unit Test Added

## Types of changes

### DIFF
--- a/.changeset/witty-papers-argue.md
+++ b/.changeset/witty-papers-argue.md
@@ -1,0 +1,5 @@
+---
+'@sitecore-content-sdk/content': patch
+---
+
+Revert to using the unmodified file name as componentName in getComponentList
--- a/packages/content/src/tools/templating/components.ts
+++ b/packages/content/src/tools/templating/components.ts
@@ -191,7 +191,7 @@ function _getComponentList(
             importPath: filePath.match(componentPathPattern)![1].replace(/\\/g, '/'), // use forward slashes for consistency
             // TODO: enabled using `toPascalCase()` for the two lines below with NextJS CSDK major release
             componentName: name,
-            moduleName: name.replace(/[^\w]+/g, ''),
+            moduleName: toPascalCase(name).replace(/[^\w]+/g, ''),
           };
         })
     );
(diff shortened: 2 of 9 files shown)


.changeset/witty-papers-argue.md
illiakovalenko, line 1: I suppose you need to add a changeset for Angular as well, since we're making changes to it.

packages/content/src/tools/templating/components.ts
```diff
@@ -188,8 +189,9 @@ function _getComponentList(
           return {
             filePath,
             importPath: filePath.match(componentPathPattern)![1].replace(/\\/g, '/'), // use forward slashes for consistency
-            componentName: toPascalCase(name),
-            moduleName: toPascalCase(name).replace(/[^\w]+/g, ''),
+            // TODO: enabled using `toPascalCase()` for the two lines below with NextJS CSDK major release
+            componentName: name,
+            moduleName: name.replace(/[^\w]+/g, ''),
```
illiakovalenko: `moduleName` loses word boundaries for kebab-case filenames (e.g. Angular's naming convention)

This reverts `moduleName` back to `name.replace(/[^\w]+/g, '')` alongside `componentName`, but only componentName was actually part of the breaking-change bug. moduleName is never matched against anything external - it's only used as an internal import alias in generated files (import * as ${moduleName} from ...).

Here is what we will have in a generated file:
```
moduleName (stripped here):       herolistcomponent

Angular toPascalCase(moduleName):    Herolistcomponent
```

This is safe for next.js too. It's more like a cosmetic issue if you will try to check a generated file

```suggestion
            moduleName: toPascalCase(name).replace(/[^\w]+/g, ''),
```

illiakovalenko (approved):
Approved in advance 👍
Overall, everything looks good. Please see a couple of comments below.

PR #611, opened 2026-09-02
[cli] Fix debug logs

MenKNas:
- [x] Changelog updated

## Description / Motivation
This PR fixes the issue of `sitecore-tools` not writing `content-sdk:*` debug logs when `DEBUG` was set in an `.env` file. The CLI was enabling the debug scopes on its own copy of the `debug` module instead of the one the SDK packages log through. It now enables them via `enableDebug` from `@sitecore-content-sdk/core`. 

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/cli/src/utils/enable-debug.ts
```diff
@@ -0,0 +1,26 @@
...
+ * The 'debug' module only reads `DEBUG` when it is first imported, which happens before the CLI
+ * loads `.env` files, so the scopes have to be enabled explicitly afterwards. This is done through
+ * `@sitecore-content-sdk/core` to update the same 'debug' instance the SDK packages log through -
+ * a bare `require('debug')` may resolve to a different copy of the module, in which case the
+ * enabled scopes are never applied to the SDK debuggers.
+ * @returns {string[]} the debug scopes which have been enabled
+ */
+export default function enableDebugLogging(): string[] {
+  const scopes = (process.env.DEBUG || '')
+    .split(',')
```
illiakovalenko: Is it possible to use whitespace-separated DEBUG values, can you double check? I think yes, however in our case currently we only split by comma

packages/cli/src/utils/process-env.ts
```diff
@@ -31,4 +32,11 @@ export default function processEnv(dir: string) {
   dotenvFiles.forEach(function(env) {
     dotenvExpand.expand(dotenv.config({ path: path.resolve(dir, env) }));
   });
+
+  // apply the DEBUG value now that the environment variables are loaded
+  const debugScopes = enableDebugLogging();
```
illiakovalenko: `enableDebugLogging()` side effect now runs twice per CLI invocation:

`processEnv()` is called from two places:
- sitecore-tools.ts:32 - processEnv(process.cwd()), unconditionally on every run
- load-config.ts:27 - processEnv(configFileDirectory), only when --config resolves to a directory different from process.cwd()

In case you provide `--config` flag `processEnv()` will be called twice and debug logs will be enabled twice and log message will be shown twice 
*Debug enabled for scopes: [content-sdk:*]*

Before this PR, enableDebugLogging() lived only in sitecore-tools.ts, outside processEnv()

Can we just call `enableDebugLogging()` separately and don't call it as a side effect within the `processEnv` function?

.changeset/jss-7330-cli-debug-logs.md
```diff
@@ -0,0 +1,5 @@
+---
+'@sitecore-content-sdk/cli': patch
+---
+
+Fix debug logs not being written by `sitecore-tools` when `DEBUG` is set in an `.env` file. The debug scopes are now enabled through `@sitecore-content-sdk/core`, so they are applied to the same 'debug' instance the SDK packages log through instead of a separate copy of the module the CLI could resolve on its own.
```
illiakovalenko: Can we mention that it solves the problem specifically for content-sdk monorepo setup? This issue is more like a chore

illiakovalenko (changes requested):

illiakovalenko (approved):
👍

PR #614, opened 2026-09-03
[react] Highlight area between Atoms elements when an Atom is being dragged over the canvas

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
**Atoms Editing Chrome — Phase 1 (`chrometype="atom"`)**

- Every rendered Atom now automatically gets wrapped in `<code type="text/sitecore"
  chrometype="atom">` chrome (via a new internal `withEditingChrome`/`withElementChromeKeys` pair
  wired into `createNCC`), so Atoms become selectable/editable/draggable in Design Studio
  with zero developer effort.
- Repeated atoms (rendered via `repeat`) get a distinct chrome id per row by suffixing the
  element's flat-spec key with the current repeat index from `useRepeatScope()`; the opening
  chrome block exposes this value as `data-element-name`, while the repeat's owning container keeps
  its plain, unsuffixed name.
- The opening atom chrome block exposes the registry element type as `data-atom-type`, allowing Design Studio to identify the atom implementation directly.
- Catalog-defined slot names are serialized as JSON in the opening block's `data-atom-slots` attribute,
  preserving the named slots declared by each atom definition for the editing experience.
- The existing top-level `PlaceholderMetadata` wrap in `DesignLibraryLowCodeComponent` now carries
  `data-component-type="atom"`, so Sitecore Pages' chrome selector can skip the whole Atoms subtree by
  ancestor instead of needing a per-kind naming convention.

## Testing Details

- [x] Unit Test Added

## Types of changes

### DIFF
--- a/.changeset/eager-tables-repeat.md
+++ b/.changeset/eager-tables-repeat.md
@@ -1,0 +1,5 @@
+---
+'@sitecore-content-sdk/react': minor
+---
+
+Highlight area between Atoms elements when an Atom is being dragged over the canvas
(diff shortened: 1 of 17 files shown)


illiakovalenko (approved):
Looks good and clear to me 👍 
I don’t have any questions

PR #616, opened 2026-09-07
[Angular][Breaking] Pre-release env name adjustments and alignment

art-alexeyenko:
- add `healthz` endpoint out of the box
- adjust public env variable for default site to align with nextjs (`CSDK_PUBLIC_DEFAULT_SITE_NAME`)
- add revalidate secret env to example env file for visibility


packages/content/src/config/define-config.ts
```diff
@@ -71,6 +71,7 @@ export const buildFallbackConfig = (env: { [key: string]: string | undefined }):
     defaultSite:
       env.SITECORE_DEFAULT_SITE ||
       env.CSDK_PUBLIC_SITECORE_DEFAULT_SITE ||
```
illiakovalenko, line 73: Can we remove CSDK_PUBLIC_SITECORE_DEFAULT_SITE?

packages/create-content-sdk-app/src/templates/angular/.env.example
```diff
@@ -12,7 +12,7 @@
 # CSDK_PUBLIC_SITECORE_API_HOST=
 
 # Site / language
-# CSDK_PUBLIC_SITECORE_DEFAULT_SITE=
+# CSDK_PUBLIC_DEFAULT_SITE_NAME=
 # CSDK_PUBLIC_SITECORE_DEFAULT_LANGUAGE=en
```
illiakovalenko: Since we want to keep consitency, let's rename it to "CSDK_PUBLIC_DEFAULT_LANGUAGE"

packages/create-content-sdk-app/src/templates/angular/src/server.ts
```diff
@@ -52,6 +52,9 @@ const loaderCache = createLoaderCache({
 
 app.use(express.json());
 
+/** Health check endpoint. */
+app.get('/healthz', (_req, res) => res.sendStatus(200));
```
illiakovalenko, line 56: Maybe it would be better to provide an OOTB middleware that customers can always reference, since this is a required part of the setup, like we did in nextjs?

illiakovalenko (approved):
Looks good 👍
Please see a couple of comments below.
Also, please add a **changeset**.

packages/create-content-sdk-app/src/templates/angular/src/server.ts
art-alexeyenko: I'm not sure we need a whole separate middleware for this tbh. The goal of the endpoint is to just report healthy, this fits within the server.ts logic

PR #617, opened 2026-09-08
[react] [nextjs] [create-content-sdk-app] Schema.org & JSON-LD Support

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
Sitecore's Layout Service can return `sitecore.context.schemas` — an array of
ready-made JSON-LD node objects (each with its own `@context`/`@type`). This adds support for
rendering them into a single `<script type="application/ld+json">` tag (serialized as one JSON
array, escaped to prevent `</script>` breakout) across all three Next.js templates.

**Changes:**
- `packages/content`: new `schemas?: Record<string, unknown>[]` on `LayoutServiceContext`; new
  framework-agnostic `getJsonLdSchema()` in `layout/json-ld.ts`; new `HTMLScript` type
  alongside `HTMLLink`.
- `packages/react`: new `JsonLdSchema` component wrapping `getJsonLdSchema()` in a
  `<script>` tag — framework-agnostic, no Next.js-specific dependencies.
- `packages/nextjs`: re-exports `JsonLdSchema`/`JsonLdSchemaProps` from `@sitecore-content-sdk/react`,
  plus `getJsonLdSchema`/`HTMLScript` from `@sitecore-content-sdk/content`.
- Templates: wired `<JsonLdSchema context={context} />` into the body of all three templates —
  `nextjs`, `nextjs-app-router`, and `nextjs-app-router-cache-components`.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/react/src/components/JsonLdSchema.tsx
```diff
@@ -0,0 +1,33 @@
...
+ */
+export const JsonLdSchema = ({ context }: JsonLdSchemaProps): JSX.Element | null => {
+  const schema = getJsonLdSchema(context);
+
+  if (!schema) {
+    return null;
+  }
+
+  return (
+    <script type={schema.type} dangerouslySetInnerHTML={{ __html: schema.innerHTML }} />
```
illiakovalenko: Could you, please, run prettier to format this file according to our rules?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/Layout.tsx
```diff
@@ -1,5 +1,5 @@
 import { JSX } from 'react';
-import { AppPlaceholder, DesignLibraryApp, Field, PageMetadataFields, Page } from '@sitecore-content-sdk/nextjs';
+import { AppPlaceholder, DesignLibraryApp, Field, PageMetadataFields, Page, JsonLdSchema } from '@sitecore-content-sdk/nextjs';
```
illiakovalenko: Could you, please, run prettier to format this file according to our rules?

packages/content/src/layout/json-ld.ts
```diff
@@ -0,0 +1,43 @@
...
+/**
+ * Builds the `<script type="application/ld+json">` data for the JSON-LD structured data
+ * schemas returned by the Sitecore Layout Service (`sitecore.context.schemas`), for
+ * consumption by any rendering layer (Next.js, React, Angular, etc). All schema objects
+ * are serialized into a single script as a JSON array.
+ * @param {LayoutServiceContext} [context] Context data from a Sitecore layout response
+ * @returns {HTMLScript | null} script tag data, or `null` if there are no schemas to render
+ * @public
+ */
+export const getJsonLdSchema = (context?: LayoutServiceContext | null): HTMLScript | null => {
```
illiakovalenko: Could this return `HTMLScript[]` instead of `HTMLScript | null`? It would match the sibling `getHeadLinks(): HTMLLink[]` (empty array as the
empty signal), remove the null branch from every caller, and leave room to emit multiple scripts later (per-schema blocks, or a `@graph` wrapper) without a
breaking change. This is `@public` API, so it's much cheaper to settle now than post-release.

illiakovalenko: The function is singular but emits all schemas as one script. `getJsonLdScripts` (or `getJsonLdSchemas`) reads truer,
and pairs naturally with `getHeadLinks`. What do you think?

packages/content/src/layout/json-ld.ts
```diff
@@ -0,0 +1,43 @@
...
+  const schemas = context?.schemas;
+
+  if (!schemas || !Array.isArray(schemas) || schemas.length === 0) return null;
```
illiakovalenko: This guards array and emptiness but not element shape. If Edge ever returns schemas: [null] (potentially it can be) or a string entry, we emit a
syntactically valid <script> containing an invalid JSON-LD node - the worst failure mode, because crawlers silently drop it and nothing errors. I would suggest filtering to plain objects and returning empty when nothing survives, plus a test case.

packages/content/src/layout/index.ts
```diff
@@ -33,6 +33,8 @@ export {
 
 export { getContentStylesheetLink } from './content-styles';
 
+export { getJsonLdSchema } from './json-ld';
```
illiakovalenko: The SDK already has an abstraction for "things we contribute to the document head": `SitecoreClient.getHeadLinks(layoutData)`.
Both the React template and the Angular template already wire it.
This adds a second, differently-shaped path - a free function taking context and returning one object rather than an array, consumed by a React-only component. Could we add for example
`getHeadScripts: HTMLScript[]` to SitecoreClient backed by this function? Both templates then wire it where they already wire head links, and
customers get a genuine override seam, since lib/sitecore-client.ts is a scaffolded, customer-owned file they can subclass. <JsonLdSchema /> still ships
for App Router (Next's Metadata API has no JSON-LD slot) - it just becomes a thin renderer over the client method.
What do you think about making it as a part of sitecore client potentially, is it possible?

packages/react/src/components/JsonLdSchema.tsx
```diff
@@ -0,0 +1,33 @@
+import React, { JSX } from 'react';
+import { getJsonLdSchema, type LayoutServiceContext } from '@sitecore-content-sdk/content/layout';
+
+/**
+ * Props for {@link JsonLdSchema}.
+ * @public
+ */
+export interface JsonLdSchemaProps {
```
illiakovalenko, line 9: Right now this is a pure pass-through of `sitecore.context.schemas` with no extension point. A customer wanting to add a (for example) BreadcrumbList
from the route or Organization from site config has to bypass the component entirely and re-implement the script escaping themselves. May be we could add `schemas?:
Record<string, unknown>[]` prop (and an `additionalSchemas` option on the function) merged with the context schemas. I suggest it because we want to support "customer-specific
extensions". Or is it considered to be handled purely through the Edge?

packages/create-content-sdk-app/src/templates/nextjs-app-router-cache-components/src/Layout.tsx
illiakovalenko, line 1: Make sure to add a subtask, we will need to write an upgrade guide and explain how to actually start utilizing JSON LD and Schemas

packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
```diff
@@ -24,6 +24,7 @@ const Layout = ({ page }: LayoutProps): JSX.Element => {
       <Head>
         <link rel="icon" href="/favicon.ico" />
       </Head>
+      <JsonLdSchema context={context} />
```
illiakovalenko: The template now has three SDK head components with three different input shapes: SitecoreStyles, PageMetaTags, JsonLdSchema. Longer term, could these potentially collapse into one `<SitecoreHead page={page} />`?

packages/content/src/layout/json-ld.ts
```diff
@@ -0,0 +1,43 @@
...
+const escapeForScriptTag = (json: string): string => {
+  const ESCAPE_LOOKUP: Record<string, string> = {
+    '&': '\\u0026',
+    '<': '\\u003c',
+    '>': '\\u003e',
+    '/': '\\u002f',
+    '\u2028': '\\u2028',
+    '\u2029': '\\u2029',
+  };
+  const ESCAPE_REGEXP = /[&<>/\u2028\u2029]/g;
```
illiakovalenko: ESCAPE_LOOKUP and ESCAPE_REGEXP are rebuilt on every call. Since they're constant, hoisting them to module scope is free. can we do that?

packages/react/src/components/JsonLdSchema.tsx
```diff
@@ -0,0 +1,33 @@
...
+ * Renders a single `<script type="application/ld+json">` tag containing all JSON-LD structured
+ * data schemas from a Sitecore route (`sitecore.context.schemas`), serialized as a JSON array.
+ * Renders nothing if there are no schemas. Framework-agnostic — the tag doesn't need to live
+ * inside `<head>` for structured data to be discovered by crawlers, so it can be rendered anywhere
+ * in the tree (e.g. nested inside `next/head`'s `<Head>` for Pages Router, or directly in a Next.js
+ * App Router Server Component).
+ * @param {JsonLdSchemaProps} props - Component props.
+ * @public
+ */
+export const JsonLdSchema = ({ context }: JsonLdSchemaProps): JSX.Element | null => {
```
illiakovalenko: Consider skipping render in editing / Design Library modes. Kind of harmless as-is, but adds extra noise in the canvas markup.

packages/content/src/layout/models.ts
```diff
@@ -43,6 +43,8 @@ export interface LayoutServiceContext {
   renderingType?: RenderingType;
   clientScripts?: string[];
   clientData?: Record<string, Record<string, unknown>>;
+  /** JSON-LD structured data node objects (each with its own `@context`/`@type`) for the current route. */
+  schemas?: Record<string, unknown>[];
```
illiakovalenko, line 47: This PR delivers the JSON-LD that Sitecore sends in `context.schemas`, which is fine. We are also generating schemas from existing page metadata ("including publication and modification dates"). We already read those fields in `resolvePageMetadataFields` (title, description, author, og:image, published/updated) for the <meta> tags, so it could build a basic Article/WebPage block from them.
If we add it, we'll need to handle duplicates: if Sitecore also sends an Article in context.schemas, the page would have two conflicting Article nodes. Skipping the generated node when context.schemas already has that `@type` would avoid it.
I hope my question is clear

.changeset/kind-memes-sink.md
```diff
@@ -0,0 +1,8 @@
+---
```
illiakovalenko, line 1: I clarified this with Product. We still need to introduce this feature for Angular as well, so we shouldn’t abandon this work, let's add it

Since these changes will be merged only after the full Content SDK release cycle anyway, they will be included in the Angular 1.1 release automatically

illiakovalenko (changes requested):
Good job! Overall, it looks ready. I’ve added a few clarification questions and suggestions

.changeset/kind-memes-sink.md
sc-nikolaoslazaridis: As discussed Angular implementation will be done in separate tickets.

packages/content/src/layout/models.ts
sc-nikolaoslazaridis: Both should be present and the schemas should be used as they come in the layout response

packages/react/src/components/JsonLdSchema.tsx
sc-nikolaoslazaridis: As discussed we leave it as is

packages/content/src/layout/index.ts
sc-nikolaoslazaridis: As discussed we don't need it at this moment.

packages/content/src/layout/json-ld.ts
sc-nikolaoslazaridis: Honestly i think having the complexity internally in this function makes the callers cleaner. In my opinion from the caller perspective is better to simply check for null instead of diving deeper and check for the length of the array. About the multiple scripts, this is something we can decide now on what to return, array of HTMLScripts adds extra functionality into the function because we would need to split the response and create multiples instead of only one. 

I would keep it as is unless you insist.

packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
sc-nikolaoslazaridis: Created a component to hold sitecore logic

packages/create-content-sdk-app/src/templates/nextjs/src/Layout.tsx
```diff
@@ -3,28 +3,25 @@
...
-  const { route, context } = layout.sitecore;
+  const { route } = layout.sitecore;
   const mainClassPageEditing = mode.isEditing ? 'editing-mode' : 'prod-mode';
 
   return (
     <>
-      <Scripts />
-      <SitecoreStyles layoutData={layout} />
+      <SitecoreHead page={page} />
       <PageMetaTags route={route} />
```
illiakovalenko: Can we put `PageMetaTags` under `SitecoreHead` as well?

illiakovalenko (approved):
Looks great now 👍

PR #619, opened 2026-09-08
[nextjs][angular] Remove version segment from sc:item cache tags

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
`sc:item` tags no longer carry a version segment (`sc:item:<id>:<locale>`), so cache writes and
webhook revalidation always agree.
Item cache tags could previously include a specific published version (`sc:item:<id>:<locale>:v<N>`)
when the layout response reported `itemVersion`, but webhook-driven revalidation always targets
`sc:item:<id>:<locale>:latest`. The mismatch meant those page cache entries were silently unreachable by `revalidateTag` and only went stale on cache TTL expiry.

A live/published render doesn't want version pinning anyway. The cache entry for a normal page render should mean "current content for this item," not "content as of version N" - that's exactly what :latest already expresses. There's no code path in this webhook-invalidation flow that legitimately wants a specific historical version cached and separately invalidated (editing/preview, where a specific version is meaningful, goes through its own dynamic rendering path and never touches this tag/cache-components flow at all).

## Testing Details

- [x] Unit Test Added

## Types of changes


MenKNas (approved):
LGTM! ✅

PR #620, opened 2026-09-09
[angular][nextjs] Normalize cache tag GUIDs for Edge webhook revalidation

sc-naveenhedallaarachchi:
- [x] Changelog updated

## Description / Motivation

- Cache invalidation failed when Experience Edge sent unhyphenated item IDs (`xxxxxxxx`) that did not match render-time tags (`xxxxxxxx-xxxx-xxxx-xxxx-xxxxxxxxxxxx`).
- Item IDs are now canonicalized to hyphenated lowercase GUIDs so webhook `revalidateTag` calls match tags registered during page render.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/angular/src/server/cache/utils.ts
```diff
@@ -137,13 +137,21 @@ export function sanitizeSitecoreCacheSegment(value: string): string {
...
  * @param {string} itemId - Raw Sitecore item id or GUID.
  * @returns {string} Normalized id segment.
  * @internal
  */
 export function normalizeSitecoreItemIdForCacheKey(itemId: string): string {
-  return itemId.trim().toLowerCase().replace(/[{}]/g, '');
+  const cleaned = itemId.trim().toLowerCase().replace(/[{}]/g, '');
+  const hex = cleaned.replace(/-/g, '');
+  if (/^[0-9a-f]{32}$/.test(hex)) {
+    return `${hex.slice(0, 8)}-${hex.slice(8, 12)}-${hex.slice(12, 16)}-${hex.slice(16, 20)}-${hex.slice(20)}`;
```
illiakovalenko: Rather than detecting a 32-char hex GUID and re-inserting hyphens at fixed offsets, can we canonicalize to the unhyphenated form instead, in both `normalizeSitecoreItemIdForCacheTag (nextjs)` and `normalizeSitecoreItemIdForCacheKey (angular)`?

Reasoning:
- Closes a correctness gap in the current fix. The regex gate (/^[0-9a-f]{32}$/) only hyphenates when the cleaned id is exactly 32 hex chars. Anything that doesn't match that exact shape falls through to return cleaned unchanged - which reintroduces the same tag-mismatch bug, just for a narrower set of ids. Stripping hyphens unconditionally has no such escape hatch - it's invariant to length/shape, so there's no format it can silently mis-normalize.
- Simpler, one code path. No regex validation, no branch, no slicing into 5 groups and reassembling with template literals - just strip and low in both packages, so the simplification oays off twice.

illiakovalenko (approved):
Looks good to me 👍

PR #623, opened 2026-09-11
[nextjs][angular] Fix dictionary revalidation triggering a full-site rebuild on every webhook call

illiakovalenko:
- [x] Changelog updated

## Description / Motivation
`createSitecoreRevalidateRouteHandler` (Next.js) and `createSitecoreRevalidateMiddleware` (Angular) both appended one `sc:dict:<site>:<locale>` tag per **configured** site on **every** webhook call, unconditionally, regardless of what the
`updates[]` payload said actually changed. Since dictionary data is fetched by every page
and shares that same tag, revalidating it on every call meant every page cache entry for that
site/locale was busted on every webhook - defeating the point of on-demand/targeted revalidation.

Experience Edge sends `"DictionaryEntry"` on dictionary-entry
webhook rows. We use that to scope dictionary revalidation precisely:
- An update is only treated as a dictionary change when `entity_definition` is `"DictionaryEntry"`
  (case-insensitive). Every othec:item:<id>:<locale>` exactly as
  before - dictionary tags are n updates.
- For a genuine Dictionary entry update, the site it belongs to is resolved from its
  `identifier` (format `<site>-<dictionary folder id>-<value>-<language>`) by matching it against
  the configured `sites` list
- If no configured site matches s skipped (debug-logged) rather than guessed - this can silently miss a dictionary update if `sites` doesn't include the relevant site, which is a deliberate fail-closed tradeoff over the previous fail-open (invalidate everything) behavior.

## Testing Details

- [x] Unit Test Added

## Types of changes


MenKNas (approved):
Looks perfect! Thank you @illiakovalenko

PR #625, opened 2026-09-15
[content] [react] [nextjs] Add styling config to the registry

sc-nikolaoslazaridis:
- [x] Changelog updated

## Description / Motivation
Adds a `stylingSolution` config to the atoms catalog so Design Studio and the
RH know which styling solution the app uses.

- `defineAtomsCatalog` accepts an optional top-level `stylingSolution`,
  defaulting to `'tailwind'` when not set.
- New closed union type `AtomsStylingSolution` (`'tailwind' | 'inline-css'`),
  defined in `@sitecore-content-sdk/content/atoms` and re-exported from
  `@sitecore-content-sdk/react` (root + `/atoms`) and
  `@sitecore-content-sdk/nextjs/atoms`.
- `stylingSolution` is included (always present, not optional) in the
  `atoms:catalog` Design Studio event payload (`SerializedCatalog`).
- `AtomsCatalog` is now generic (`AtomsCatalog<T>`): `stylingSolution` stays
  optional on the input but is guaranteed present once resolved.

## Testing Details

- [x] Unit Test Added

## Types of changes


packages/react/src/atoms/define-atoms-catalog.ts
```diff
@@ -40,6 +40,6 @@ import type { AtomsCatalogInput, Exact } from './types';
  */
 export function defineAtomsCatalog<T extends AtomsCatalogInput>(
   input: Exact<T, AtomsCatalogInput>
-) {
-  return defineCatalog(schema, input);
+): AtomsCatalog<T> {
+  return defineCatalog(schema, { stylingSolution: 'tailwind', ...input });
```
MenKNas: Shouldn't we keep the default solution when `StylingSolution` is undefined?

So maybe: `return defineCatalog(schema, { ...input , stylingSolution: 'tailwind',});`

packages/react/src/atoms/catalog-serializer.ts
```diff
@@ -46,6 +46,7 @@ export function serializeCatalog(catalog: AtomsCatalog): SerializedCatalog {
   );
 
   const serializedCatalog: SerializedCatalog = {
+    stylingSolution,
```
MenKNas, line 49: There won't be any defaults here?

MenKNas (changes requested):

MenKNas:
Also I have a more generic question, maybe I am missing something out. The default `tailwind` is right for App-Router but for Pages Router starter we don't use Tailwind. If I am not mistaken `tailwind` will be sent by default for the Pages Router template as well which looks wrong. Shouldn't we explicitly set defaults in both starters like:

- Pages Router: `stylingSolution: 'inline-css'`
- App Router: `stylingSolution: 'tailwind'`

packages/react/src/atoms/define-atoms-catalog.ts
sc-nikolaoslazaridis: That would overwrite whatever is in the input with tailwind i think, let me double check

packages/react/src/atoms/catalog-serializer.ts
sc-nikolaoslazaridis: by the time is here it is already defaulted to tailwind

sc-nikolaoslazaridis:
> Also I have a more generic question, maybe I am missing something out. The default `tailwind` is right for App-Router but for Pages Router starter we don't use Tailwind. If I am not mistaken `tailwind` will be sent by default for the Pages Router template as well which looks wrong. Shouldn't we explicitly set defaults in both starters like:
> 
> * Pages Router: `stylingSolution: 'inline-css'`
> * App Router: `stylingSolution: 'tailwind'`

This default took into consideration the examples from SCB team. I would avoid having 2 different defaults, i think we should set default to inline-css if we going to change it. Let me align with the architect.

MenKNas (approved):
LGTM ✅

PR #627, opened 2026-09-16
[csdk] Theming support mechanism - phase 1

MenKNas:
- [x] Changelog updated

## Description / Motivation
This is a proposal PR for enabling site-level design-token theming in the CSDK. It should not be treated as the final contract since there is still pending work on the backend side in order to finalize the contract and lock everything in. Main things that are included:

- Opt-in via `CSDK_FEATURE_THEMING` / `NEXT_PUBLIC_CSDK_FEATURE_THEMING` (none/0, site/1, page/2) mapped to `theming.mode`.
- When enabled, injects `<link rel="stylesheet" href="<edge-url>/theming/<site-id>">` through the existing `getHeadLinks()` pipeline, independent from Design Library enableThemes.
- Default remains off. Page theming is out of scope at this point which means that page currently behaves like site.
- Site ID is still TBD: a link is emitted only when `layout.sitecore.context.site.id` is present.

## Testing Details
Suggested steps for initial testing:

- Confirm unset / none / 0 adds no theming links
- Confirm site / 1 adds the site stylesheet when `context.site.id` exists
- Confirm no theming link when mode is site but site id is missing
- Confirm Design Library `enableThemes` behavior is unchanged
 - Confirm page / 2 emits only the site link (no page stylesheet) - This is enough for the current phase

 Review whether this config/env/site-id approach is the one we want to keep
- [x] Unit Test Added

## Types of changes


packages/content/src/layout/theming.ts
```diff
@@ -0,0 +1,86 @@
...
+
+export type { ThemingMode };
+
+/**
+ * Returns whether site-level design-token theming should emit a stylesheet link.
+ * `page` is treated as site-level until page theming is implemented.
+ * @param {ThemingMode} mode Theming mode
+ * @returns {boolean} Whether the site theme stylesheet should be included
+ */
+const isSiteThemingEnabled = (mode: ThemingMode): boolean => mode === 'site' || mode === 'page';
```
sc-nikolaoslazaridis: I would avoid keeping any references of the page mode for now

packages/content/src/layout/theming.ts
```diff
@@ -0,0 +1,86 @@
...
+export const THEMING_BODY_CLASS_NAME = 'sc-ds-theme';
+
+/**
+ * Returns the body class name for site-level design-token theming, or `undefined` when theming is off.
+ * `page` is treated as site-level until page theming is implemented.
+ * @param {ThemingMode} mode Theming mode from `sitecore.config`
+ * @returns {string | undefined} Body class name when site theming is enabled
+ * @public
+ */
+export const getThemingBodyClassName = (mode: ThemingMode): string | undefined =>
```
sc-nikolaoslazaridis: I think the name of the function is not representative on what it does. Can you either adapt the name or the functionality?

packages/content/src/layout/theming.ts
```diff
@@ -0,0 +1,86 @@
...
+  sitecoreEdgeUrl: string = constants.SITECORE_EDGE_PLATFORM_URL_DEFAULT
+): string => `${normalizeUrl(sitecoreEdgeUrl)}/theming/${encodeURIComponent(siteId)}`;
+
+/**
+ * Options for {@link getThemingStylesheetLinks}.
+ * @public
+ */
+export type ThemingStylesheetLinksOptions = {
+  /**
+   * Theming depth from `sitecore.config` `theming.mode`.
```
sc-nikolaoslazaridis: The word 'depth' might be confusing to external users... that goes for all the cases you mention it in the PR (i know i wrote it in the ticket...)

packages/nextjs/src/config/define-config.test.ts
```diff
@@ -328,6 +328,28 @@ describe('defineConfig', () => {
     });
   });
 
+  describe('config.theming', () => {
+    it('does not set a theming fallback from environment variables', () => {
+      process.env.NEXT_PUBLIC_CSDK_FEATURE_THEMING = 'site';
```
sc-nikolaoslazaridis: Can you remove everything related to the env variables since we wont support them?

sc-nikolaoslazaridis (changes requested):

packages/content/src/layout/theming.ts
MenKNas: It's more like notes to ourselves for phase 2 and since this is internal I wouldn't bother so much, unless we have decided to release this for phase 1.

